PR #11, opened 2022-03-24
Add compound v2 subgraph

dmelotik:
This subgraph follows the [schema-lending.graphql](https://github.com/messari/subgraphs/blob/master/schema-lending.graphql) schema for the Compound Finance protocol


subgraphs/compound-v2/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Protocol: https://compound.finance/
+
+Docs: https://compound.finance/docs
+
+Smart contracts: https://github.com/compound-finance/compound-protocol
+
+Deployed addresses: https://compound.finance/docs#networks
+
+Existing subgraphs: https://github.com/compound-finance/compound-v2-subgraph
```
this-username-is-taken, line 21: This is a more up-to-date subgraph: https://github.com/graphprotocol/compound-V2-subgraph

subgraphs/compound-v2/src/common/addresses.ts
```diff
@@ -0,0 +1,315 @@
...
+export const ADDRESS_ZERO = Address.fromString("0x0000000000000000000000000000000000000000")
+
+// "factory" address (Comptroller.sol)
+export const COMPTROLLER_ADDRESS = Address.fromString("0x3d9819210a31b4961b30ef54be2aed79b9c9cd3b")
+
+// PriceOracle Addresses
+export const PRICE_ORACLE1_ADDRESS = Address.fromString("0x02557a5E05DeFeFFD4cAe6D83eA3d173B272c904")
+
+// cToken addresses
+export const CAAVE_ADDRESS  =   Address.fromString("0xe65cdb6479bac1e22340e4e755fae7e509ecd06c")
```
this-username-is-taken, line 15: You shouldn't hardcode these. Otherwise we'd need to update the code and redeploy every time Compound adds a new market.

subgraphs/compound-v2/README.md
dmelotik: thanks- updated

subgraphs/compound-v2/src/common/addresses.ts
dmelotik: fixed with data source template

PR #15, opened 2022-03-26
Reference subgraph implementation

this-username-is-taken:
Creating a reference subgraph implementation (based on our Uniswap v2 subgraph) to bootstrap development.

- Added a common library of utility functions so you don't need to implement from scratch
- Introduced some loose guidelines around code organization to improve readability


mul53:
Look good

schema-common.graphql
```diff
@@ -535,7 +535,7 @@ type Swap implements Event @entity {
   tokenIn: Token!
 
   " Amount of token deposited into pool in native units "
-  amountIn: BigDecimal!
+  amountIn: BigInt!
```
0xbe1, line 538: may I ask why BigInt over BigDecimal? Is it because this is the amount in native units, for example, wei?

subgraphs/_reference_/README.md
```diff
@@ -0,0 +1,42 @@
...
+> - Make sure you include this README (but without this section) in the  
+>   subgraph, with links to protocol-related resource and build instruction.
+> - Make sure you run `npm run format` to format your code before committing.
+> - Try to follow the same folder structure inside of `src` for your subgraph.
+> - There is a set of useful functions in `src/common` that you can leverage.
+
+## Links
+
+- Protocol: https://uniswap.org/
+- Analytics: https://info.uniswap.org/
```
0xbe1, line 18: here's the v2 analytics https://v2.info.uniswap.org/#/

subgraphs/_reference_/README.md
```diff
@@ -0,0 +1,42 @@
...
+- Docs: https://docs.uniswap.org/protocol/V2/introduction
+- Smart contracts: https://github.com/Uniswap/v2-core
+- Deployed addresses: https://docs.uniswap.org/protocol/V2/reference/smart-contracts/factory
```
0xbe1, line 21: how about https://etherscan.io/address/0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f?

subgraphs/_reference_/README.md
```diff
@@ -0,0 +1,42 @@
...
+- Official subgraph: https://github.com/Uniswap/v2-subgraph
+
+## Build
+
+- Initialize subgraph (Subgraph Studio):
+  ```
+  graph init --product subgraph-studio
+  --from-contract <CONTRACT_ADDRESS> [--network <ETHEREUM_NETWORK>] [--abi <FILE>] <SUBGRAPH_SLUG> [<DIRECTORY>]
+  ```
+- Initialize subgraph (Hosted Service):
```
0xbe1, line 31: I guess we prefer hosted service? if so, I suggest moving this line above subgraph studio

subgraphs/_reference_/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "reference-subgraph",
+  "license": "UNLICENSED",
```
0xbe1, line 3: just curious, what will be the license of these subgraphs?

schema-common.graphql
this-username-is-taken: yeah, it's documented here: https://github.com/messari/subgraphs/pull/10

subgraphs/_reference_/schema.graphql
```diff
@@ -0,0 +1,557 @@
+enum Network {
```
0xbe1, line 1: I suggest adding a line here showing which exact schema template does it extend.

In this case, something like "Extend schema-dex-amm.graphql, see the diff at the bottom"

subgraphs/_reference_/README.md
this-username-is-taken: this you can find in the manifest though

this-username-is-taken: Thanks for helping out on the review btw!

subgraphs/_reference_/src/common/getters.ts
```diff
@@ -0,0 +1,115 @@
...
+    token = new Token(tokenAddress.toHexString());
+    token.symbol = fetchTokenSymbol(tokenAddress);
+    token.name = fetchTokenName(tokenAddress);
+    token.decimals = fetchTokenDecimals(tokenAddress);
+    token.save();
+  }
+  return token;
+}
+
+export function getOrCreateUsageMetricSnapshot(event: ethereum.Event): UsageMetricsDailySnapshot {
```
0xbe1, line 36: Any chance common functions like this, live in a shared location instead, to get rid of copy-and-paste?

subgraphs/_reference_/src/common/metrics.ts
```diff
@@ -0,0 +1,66 @@
+import { BigDecimal, Address, ethereum } from "@graphprotocol/graph-ts";
+import { _Account, _DailyActiveAccount, UsageMetricsDailySnapshot } from "../../generated/schema";
+import { FACTORY_ADDRESS, SECONDS_PER_DAY } from "./constants";
+import { getOrCreateDexAmm, getOrCreatePoolDailySnapshot, getOrCreateUsageMetricSnapshot } from "./getters";
+
+// Update FinancialsDailySnapshots entity
+export function updateFinancials(event: ethereum.Event): void {
```
0xbe1, line 7: why are they commented out?

subgraphs/_reference_/src/common/metrics.ts
```diff
@@ -0,0 +1,66 @@
...
+    dailyActiveAccount.save();
+    usageMetrics.activeUsers += 1;
+  }
+
+  usageMetrics.save();
+}
+
+// Update UsagePoolDailySnapshot entity
+export function updatePoolMetrics(event: ethereum.Event): void {
+  // get or create pool metrics
```
0xbe1, line 54: why are they commented out?

subgraphs/_reference_/package.json
this-username-is-taken: should be MIT

subgraphs/_reference_/README.md
this-username-is-taken: We actually prefer subgraph studio as it's recommended by The Graph. I will swap the lines below

subgraphs/_reference_/src/common/tokens.ts
```diff
@@ -0,0 +1,184 @@
...
+export function fetchTokenSymbol(tokenAddress: Address): string {
+  let contract = ERC20.bind(tokenAddress);
+  let contractSymbolBytes = ERC20SymbolBytes.bind(tokenAddress);
+
+  // try types string and bytes32 for symbol
+  let symbolValue = UNKNOWN_TOKEN_VALUE;
+  let symbolResult = contract.try_symbol();
+
+  // non-standard ERC20 implementation
+  if (symbolResult.reverted) {
```
0xbe1, line 19: how about

```
if (!symbolResult.reverted) {
  symbolValue = symbolResult.value;
  return;
}
...
```

To make it less nested?

subgraphs/_reference_/src/common/getters.ts
this-username-is-taken: Yeah, I've thought about that too. The problem is they depend on generated types from both the schema and the ABIs, which makes things a little bit more tricky

subgraphs/_reference_/src/common/metrics.ts
this-username-is-taken: These are meant more as boilerplates that'll be filled out depending on the subgraph, and will be different from subgraph to subgraph. They are `common` within a subgraph but not common across different subgraphs.

Added comments to explain

subgraphs/_reference_/src/common/tokens.ts
this-username-is-taken: good idea!

subgraphs/_reference_/src/common/getters.ts
0xbe1: Can we do this for utilities that depend only on schema?

0xbe1:
Do we intend to leave helpers impl as empty, so it is more of a reference rather than something that can actually be deployed? Thx

this-username-is-taken:
That's exactly right. I should add a comment to make that clear

0xbe1:
I would suggest a runnable reference implementation though, since that will reflect the reality :) just a suggestion

this-username-is-taken:
Yeah, that's tricky. The current Uniswap v2 subgraph syncs terribly slowly (will take months to fully sync). Ideally I'd want something that's both simple and syncs quickly (fully sync in a couple hours) for a full implementation. So far I haven't found one. I may consider a fork of uniswap on a sidechain that's low volume (e.g. Honeyswap or Ubeswap) but this isn't the highest priority right now.

subgraphs/_reference_/src/common/getters.ts
this-username-is-taken: I'm not sure what's the best way to do that but I do think that's better in the long run. I don't have the capacity to work on this right now. Would you like to take a stab at it? If not, I can probably get to this towards the end of next week.

I'm going to close out this PR for now but feel free to start a new one.

subgraphs/_reference_/src/common/metrics.ts
```diff
@@ -0,0 +1,71 @@
...
+  // let protocol = getOrCreateDexAmm();
+  // // Update the block number and timestamp to that of the last transaction of that day
+  // financialMetrics.blockNumber = event.block.number;
+  // financialMetrics.timestamp = event.block.timestamp;
+  // financialMetrics.totalValueLockedUSD = protocol.totalValueLockedUSD;
+  // ...
+  // financialMetrics.save();
+}
+
+export function updateUsageMetrics(event: ethereum.Event, from: Address): void {
```
harsh9200, line 23: Can we make it more generalized, so that it can handle both events for event Handlers and Calls for call Handlers

PR #25, opened 2022-04-01
Add Price Oracle

harsh9200:


subgraphs/stakedao/src/Oracle/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  let decimals = readValue<BigInt>(
+    token.try_decimals(),
+    constants.DEFAULT_DECIMALS
+  );
+
+  return decimals;
+}
+
+export function getUsdPricePerToken(tokenAddr: Address): [BigDecimal, BigInt] {
+  let netowrk = dataSource.network();
```
AP-Atul, line 29: `network` typo if not done on purpose.

subgraphs/stakedao/src/Oracle/common/constants.ts
```diff
@@ -0,0 +1,135 @@
+import { Address, BigDecimal, BigInt, TypedMap } from "@graphprotocol/graph-ts";
+import { UniswapPair__getReservesResult } from "../../../generated/Controller/UniswapPair";
+import { SushiSwapPair__getReservesResult } from "../../../generated/Controller/SushiSwapPair";
+
+///////////////////////////////////////////////////////////////////////////
+//////////////////////////////////? COMMON ?///////////////////////////////
```
this-username-is-taken, line 6: nit: can you change `?` to `/` for consistency?

subgraphs/stakedao/subgraph.yaml
```diff
@@ -27,10 +26,32 @@ dataSources:
           file: ./abis/Ethereum/ERC20.json
         - name: Strategy
           file: ./abis/Ethereum/strategy.json
-        - name: CalculationsCurve
-          file: ./abis/Price/CalculationsCurve.json
-        - name: CalculationsSushi
-          file: ./abis/Price/CalculationsSushiSwap.json
+        
+        #? Price Oracle
```
this-username-is-taken, line 30: nit: can you change `?` to `/` for consistency?

subgraphs/stakedao/src/Oracle/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  dataSource,
+  log,
+} from "@graphprotocol/graph-ts";
+
+import * as constants from "./common/constants";
+import { getCurvePriceUsdc } from "./routers/CurveRouter";
+import { getPriceUsdc as getPriceUsdcUniswap } from "./routers/UniswapRouter";
+import { getPriceUsdc as getPriceUsdcSushiswap } from "./routers/SushiSwapRouter";
+
+export function getUsdPricePerToken(tokenAddr: Address): BigDecimal[] {
```
this-username-is-taken, line 14: I think you can do this `[BigDecimal, BigDecimal]` as return value. Make a comment on what they are.

subgraphs/stakedao/src/Oracle/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  let network = dataSource.network();
+  let decimals = BigInt.fromI32(10)
```
this-username-is-taken, line 16: why not use a constant like `USDC_DENOMINATOR`?

subgraphs/stakedao/src/Oracle/routers/CurveRouter.ts
```diff
@@ -0,0 +1,141 @@
...
+  //? Use first coin from pool and if that is empty (due to error) fall back to second coin
+  let preferredCoinAddress = coins[0];
+  if (preferredCoinAddress.toHex() == constants.ZERO_ADDRESS_STRING) {
+    preferredCoinAddress = coins[1];
+  }
+
+  //? Look for preferred coins (basic coins)
+
+  let coinAddress: Address;
+  for (let coinIdx = 0; coinIdx < 8; coinIdx++) {
```
this-username-is-taken, line 87: make a comment about 8

subgraphs/stakedao/src/Oracle/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    log.warning(
+      "[getPriceUsdcSushiswap] tokenAddress: {}, sushiswapPrice: {}",
+      [tokenAddr.toHexString(), sushiswapPrice.div(decimals).toString()]
+    );
+
+    return [sushiswapPrice, decimals];
+  }
+
+  // TODO: WIP - ChainLink Router
+  // //? 4. ChainLink
```
this-username-is-taken, line 54: Should always go with chainlink first

subgraphs/stakedao/src/Oracle/routers/CurveRouter.ts
```diff
@@ -0,0 +1,141 @@
+import * as utils from "../common/utils";
+import { getPriceUsdc } from "./SushiSwapRouter";
+import * as constants from "../common/constants";
+import { Address, BigDecimal, BigInt, dataSource } from "@graphprotocol/graph-ts";
+import { CurveRegistry as CurveRegistryContract } from "../../../generated/Controller/CurveRegistry";
+
+export function getCurvePriceUsdc(
```
this-username-is-taken, line 7: Does this handle v2 pools (i.e. non-like tokens)? Not necessary for this iteration, but need to have it at some point

subgraphs/stakedao/src/Oracle/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  return [
+    constants.BIGDECIMAL_ZERO,
+    BigInt.fromI32(10)
+      .pow(constants.BIGINT_ZERO.toI32() as u8)
+      .toBigDecimal(),
+  ];
+}
+
+export function getUsdPrice(tokenAddr: Address, amount: BigInt): BigDecimal {
+  let usdPrice = getUsdPricePerToken(tokenAddr);
```
this-username-is-taken, line 71: You should not remove Yearn's oracle contracts completely. We should rely on Yearn's oracle when its available and fall back to your oracle when it's not in order to reduce the number of JSON-RPC calls

this-username-is-taken:
Great work!

A couple things:

1. Seems like the way you structured the code makes it very easy to use it as a drop-in library. That's very good! However, it will require a couple steps so you should include a `README.md` in the oracle's src folder to explain how to use it (i.e. include it in any subgraph). Maybe also a good idea to give a high level explanation on what the oracle is doing.
2. As commented in the code, we want to rely on existing onchain oracles as much as possible because these usually only involve a single contract call.
3. This is a good start but I can see this being slow to index given the amount of contract calls involved. There are a lot of things we can do to reduce that. For example, cache existing prices, store LP token addresses, hardcode a set of token addresses and route to the corresponding DEX directly. In any case, we don't need to work on optimization for now. Let's index with this code and see how things are first.

subgraphs/stakedao/src/Oracle/common/constants.ts
harsh9200: I am using this extension https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments which helps in writing better comments. If we can mainstream, then that would be great but I will remove this for now.

image

image

subgraphs/stakedao/src/Oracle/index.ts
harsh9200: Generally, a multiple return type throws an error in `thGraph`, [here](https://discord.com/channels/438038660412342282/438070183794573313/963050151390367824) is my question on theGraph Discord.

subgraphs/stakedao/src/Oracle/index.ts
harsh9200: I am still not sure if this denominator will work for every `tokenAddr` or not, let's wait for others to test the implementation on their own subgraphs.

subgraphs/stakedao/src/Oracle/routers/CurveRouter.ts
harsh9200: This is completely mimicking a oracle, so it should also work on v2 Pools.

subgraphs/stakedao/src/Oracle/index.ts
harsh9200: The ordering is completely dependent on the subgraphs requirement, my subgraph(Stake DAO) is completely dependent on Curve Router and only the Curve router works for me, that's why it's number one for me. 

Ideally, the subgraph developers should change the ordering depending on their subgraph, so that it could index a little faster and reduce unnecessary JSON RPC Calls.

subgraphs/stakedao/src/Oracle/index.ts
harsh9200: yup, in the next iteration, I am going to add the original oracle contracts, so that we can use them when they are available and reduce unnecessary JSON RPC Calls.

harsh9200:
@this-username-is-taken 

yeah, I will add a README for developers to easily implement this in their own subgraph. I totally agree that the indexing speed will be slow because of so many contract calls, but to handle this I can only include the original oracle contracts so that the prices could be fetched in one single contract call.

The other approaches that you mentioned (existing prices, storing LP token addresses, hardcode a set of token addresses, and routing to the corresponding DEX directly), are specifically dependent on subgraphs and for that, I also have to play with the schema, which is not ideal. I don't think this should be involved in the generic implementation.

subgraphs/stakedao/src/Oracle/README.md
```diff
@@ -0,0 +1,107 @@
...
+├── routers
+│   ├── CurveRouter.ts
+│   ├── SushiSwapRouter.ts
+│   └── UniswapRouter.ts
+└── index.ts
+```
+
+## Development Status
+
+### Mainet
```
this-username-is-taken, line 94: Mainnet

subgraphs/stakedao/src/Oracle/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+🔨 = In progress.  
+🛠 = Feature complete. Additional testing required.
+
+| Method                                                                               |      Type      | StartBlock | MultiCall | Status |
```
this-username-is-taken, line 99: Explain what Multicall is

this-username-is-taken:
Change `src/Oracles` to `src/oracles` for consistency

this-username-is-taken:
Can you add this to the reference subgraph to make it easier for people to use? Also add a comment in the reference subgraph's README to let people know this is available

PR #30, opened 2022-04-03
fix: LiquidityPool, Market and Vault should implements Pool

0xbe1:


0xbe1:
Not sure we should follow up with a commit that also adds `implements Pool` to existing subgraph schema such as uniswap v2 and v3

this-username-is-taken:
Unfortunately you can't do this because `MarketDailySnapshot` has a field `market` that is of type `Market`, which is different from `PoolDailySnapshot` that has the field `pool` instead. I'm not a GraphQL expert. Do you know ways that we can get around this in a nice way?

0xbe1:
> because MarketDailySnapshot has a field market that is of type Market, which is different from PoolDailySnapshot that has the field pool instead.

If that's the reason can we simply rename market to pool? 🤣 I don't fully get why this is the case

this-username-is-taken:
It's mostly terminology. For example, yield aggregators use vaults instead of pools, where pools may have a different meaning instead (means the underlying pool which the vault depends on). If it's causing confusion for you, I can probably just remove the interfaces and leave only the types.

this-username-is-taken:
Integrated changes into #47

PR #46, opened 2022-04-06
Aave calculation methodology v1.0.0

ken-m-1:


ken-m-1:
Used as example only. Will manage the changes here https://github.com/ken-m-1/subgraphs/tree/master/subgraphs

PR #47, opened 2022-04-07
Major schema update

this-username-is-taken:
#### Common Changes

- Changed `feesUSD` to `totalRevenueUSD`. For DEXes and Lending Protocols, everything stays the same since `fees` = `totalRevenue`. For yield aggregator, you now need to include all revenue (including supply side portion).
- Added `methodologyVersion` to the `Protocol` interface/entity so that we can track which methodology is being used.
- Removed `protocolTreasuryUSD` field from the `FinanceDailySnapshot` entity
- Removed the `Pool` and `PoolDailySnapshot` interface for clarity
- Added `Account` and `DailyActiveAccount` as a part of all schemas, since they are always used to compute usage metrics
- Added `totalVolumeUSD` to the `Protocol` interface to make aggregator easier
- Added a lot of documentation

#### Protocol-Type Specific Changes

- Added `totalDepositUSD` and `totalBorrowUSD` to the Lending schema. Note that `totalDepositUSD` should be the same as `totalValueLockedUSD` but `totalBorrowUSD` is different from `totalVolumeUSD` since the former is a point in time and latter is in aggregate.
- Made `outputToken`-related fields optional in the DEX schema since Bancor v2 doesn't have LP token (output token)
- Updated `LiquidityPoolFeeType` to account for LP fees
- Updated `to`/`from` fields for DEX deposit/withdraw


0xbe1:
I believe `Pool` is helpful if dex, lending, etc they all `implements Pool` so that we know by code that they share something

0xbe1:
I think you are not removing PoolDailySnapshot, do you? I find them still there

schema-yield.graphql
```diff
@@ -172,8 +175,8 @@ type FinancialsDailySnapshot @entity {
   " Gross revenue for the protocol (revenue claimed by protocol). Examples: AMM protocol fee (Sushi’s 0.05%). OpenSea 10% sell fee. "
   protocolSideRevenueUSD: BigDecimal!
 
-  " Fees paid by the users. e.g. 0.30% of swap fee in Sushiswap "
-  feesUSD: BigDecimal!
+  " Total revenue accrued by the protocol. e.g. 0.30% of swap fee in Sushiswap, all yield generated by Yearn. "
+  totalRevenueUSD: BigDecimal!
```
harsh9200, line 179: How is `totalRevenueUSD` different from `protocolSideRevenueUSD` ??

this-username-is-taken:
@0xbe1 yes, the interface is meant to serve that purpose, but due to some limitations in GraphQL, we can't do that in the schema

schema-yield.graphql
this-username-is-taken: `totalRevenueUSD` is literally all money coming in. In the case of yield aggregators, revenue are split between:
- Supply side (liquidity providers who deposit)
- Strategists
- Partners (in the case of Yearn's partnership program)
- Protocol themselves (all fees collected by the protocol after subtracting the strategist portion and the partner portion)

`protocolSideRevenueUSD` would correspond to the last item above

dmelotik:
a field that I have added in to my schema for compound that I think would be helpful for other subgraphs:

`lendingProtocol._totalVolumeUSD` - this just made it easier to update and throw into the financial snapshot / this might be a good metric to have anyway

And it may be useful to just add in `Account` and `DailyActiveAccount` in the actual schemas since it is probably the best way to calculate some of the usage metrics

schema-lending.graphql
```diff
@@ -122,23 +128,26 @@ type FinancialsDailySnapshot @entity {
...
-  protocolTreasuryUSD: BigDecimal
-
   " Only relevant for protocols with PCV. "
   protocolControlledValueUSD: BigDecimal
 
   " Total volume in USD "
   totalVolumeUSD: BigDecimal!
 
+  " Current balance of all deposited assets (not historical cumulative), in USD "
+  totalDepsoitUSD: BigDecimal!
```
MichaelC1999, line 138: Typo, same in the lending protocol entity

docs/Schema.md
```diff
@@ -50,12 +69,14 @@ There are broadly two kinds of yields in DeFi:
...
 
 It's also common for a single pool to have multiple reward tokens. For example, Sushiswap's MasterChef v2 allows for multiple `Rewarders`. Some Curve pools also have both CRV as a reward and also the pool token (e.g. FXS for FRAX, SNX for sUSD) as another reward.
 
+There are different ways to calculate `rewardTokenEmissionsAmount` and `rewardTokenEmissionsAmount`. In particular, you can calculate a theoretical emission amount based on the underlying emission equation, or you can calculate the realized amount based on harvests. It's recommended to use the theoretical amount as it's more accurate and consistent. When calculating these in a snapshot, you should calculate them as the per-block amount of the current block normalized to the during of that snapshot (e.g. normalize the per-block amount to the daily amount for a daily snapshot). The reasoning for that is when we show this data on the front-end, it'll be converted to an APR, which will drive user decision (e.g. decide where they want to invest their money in the future), so the data should be forward-looking.
```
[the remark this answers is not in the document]
this-username-is-taken: you guys have good eyes!

schema-lending.graphql
```diff
@@ -332,9 +256,18 @@ type Market @entity {
...
   totalVolumeUSD: BigDecimal!
 
+  " Current balance of all deposited assets (not historical cumulative), in USD "
+  totalDepositUSD: BigDecimal!
+
+  " Current balance of all borrowed assets (not historical cumulative), in USD. In the case of CDPs, this will be all minted assets "
+  totalBorrowUSD: BigDecimal!
+
   " Amount of input tokens in the market. The ordering should be the same as the market's `inputTokens` field. "
   inputTokenBalances: [BigInt!]!
```
dmelotik, line 266: I think it could be useful to specify more on what this exactly is. i had it wrong at first

schema-lending.graphql
```diff
@@ -332,9 +256,18 @@ type Market @entity {
...
 
+  " Price per share of input token in USD "
+  inputTokenPricesUSD: [BigDecimal!]!
+
   " Total supply of output token "
   outputTokenSupply: BigInt!
```
dmelotik, line 272: same here i guess. my understanding for compound is the number of cTokens outstanding. Once it is redeemed the cToken supply is decreased in the contract. And I misunderstood this at first

schema-lending.graphql
this-username-is-taken: hmm, can you give me an example of how I can clarify this?

PR #53, opened 2022-04-08
Add Liquity subgraph

spitko:
Some notes:

- There can multiple Event entities created in a single ethereum event, e.g deposit + borrow. This means the default id for events `{ Transaction hash }-{ Log index }` does not work, so i added the event type string to the end.

- FinancialsDailySnapshot.totalVolumeUSD is treated the same as Market.totalVolumeUSD - only borrow volume is counted.

I will update this once the new schema is finalized


this-username-is-taken:
> 

Would you be able to build entities with conflicting IDs like this https://github.com/messari/subgraphs/pull/71/commits/0f8a5def6f3331128ce832b2b5c858dd2f492e23

subgraphs/liquity/README.md
```diff
@@ -0,0 +1,9 @@
+# Liquity Subgraph
+
+## Links
+
+- Protocol website: https://www.liquity.org
+- Protocol documentation: https://docs.liquity.org
+- Smart contracts: https://github.com/liquity/dev/tree/main/packages/contracts/contracts
+- Deployed addresses: https://docs.liquity.org/documentation/resources#contract-addresses
+- Existing subgraphs: https://thegraph.com/explorer/subgraph?id=2TmD7Fq3K2BQGmUXNPzrNMXb6PDNbcBWfAHa5o3gRJC8
```
this-username-is-taken, line 9: Can you include the methodology in this PR: https://github.com/messari/subgraphs/pull/63

You can put it after `Links`

this-username-is-taken: Also can you update your smart contract interaction chart to reflect the subgraph implementation and add it here like: https://github.com/messari/subgraphs/blob/master/subgraphs/uniswap-v2/README.md#smart-contracts-interactions

this-username-is-taken:
Did you include `protocolSideRevenue`?

spitko:
> Did you include `protocolSideRevenue`?

It is included in the schema, but it is always zero

spitko:
Note that ETH price is not guaranteed to be up to date for Stability Pool events.
But they are only used for calculating current protocol TVL, so it's probably fine.

spitko:
Latest version query url, still indexing:
https://api.studio.thegraph.com/query/25077/liquity/v0.0.4

PR #58, opened 2022-04-10
Added file for calculating block rate moving average

steegecs:


subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (!block) {
+        block = new _Block(blockNumber.toString())
+        block.blockNumber = blockNumber
+        block.timestamp = timestamp
+    }
+    return block
+}
+
+export function getOrCreateBlockTracker(): _BlockTracker {
+    let blockTracker = _BlockTracker.load("BlockTracker")
```
this-username-is-taken, line 130: use named constant

subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+    blocks = blockTracker.blocks
+
+    // Create entity for the current block
+    let currentTimestamp = event.block.timestamp.toI32()
+    let currentBlockNumber = event.block.number.toI32()
+    let currentBlock = getOrCreateBlock(currentTimestamp, currentBlockNumber)
+
+    if (blocks.length == INT_ZERO) {
+        blocks.push(currentBlock.id)
+        blockTracker.blocks = blocks
```
this-username-is-taken, line 59: don't you need to update `blocksPerDay` here? Isn't it still `BIG_DECIMAL_ZERO`?

subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+export const WINDOW_SIZE_SECONDS = 86400
+export const WINDOW_SIZE_SECONDS_BD = BigDecimal.fromString(WINDOW_SIZE_SECONDS.toString())
+
+// The storage interval tells you to only store blocks that are space by at least this amount. 
+// Increasing this value will mean less blocks stored and less frequently computes blocksPerDay.
+export const BLOCK_STORAGE_INTERVAL = 5 as i32
+
+// Paramaters: [event.block.timestamp, event.block.number]
+// Call this function in event handlers frequently enough so that it updates on most blocks 
+export function updateMovingAverageBlockSpeedForecast(event: ethereum): void {
```
this-username-is-taken, line 46: I think it'll be convenient if you returned `blocksPerDay` here

subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+// Window of block time used to calculate the moving average.
+// Increasing means less deviation but also less sensitivity to changing block speeds.
+export const WINDOW_SIZE_SECONDS = 86400
+export const WINDOW_SIZE_SECONDS_BD = BigDecimal.fromString(WINDOW_SIZE_SECONDS.toString())
+
+// The storage interval tells you to only store blocks that are space by at least this amount. 
+// Increasing this value will mean less blocks stored and less frequently computes blocksPerDay.
+export const BLOCK_STORAGE_INTERVAL = 5 as i32
+
+// Paramaters: [event.block.timestamp, event.block.number]
```
this-username-is-taken, line 44: Comment like this: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns

subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+    let startTimestamp = currentTimestamp - WINDOW_SIZE_SECONDS
+
+    // Make sure you still have 2 blocks to calculate rate (This shouldn't really happen past the beginning).
+    while(blocks.length > INT_TWO) {
+        let oldestPushedBlock = _Block.load(blocks[INT_ZERO])
+
+        // Remove items from array and storage that were greater than 24 Hours ago.
+        if (oldestPushedBlock!.timestamp < startTimestamp) {
+            store.remove('_Block', oldestPushedBlock!.id)
+            blocks.shift()
```
this-username-is-taken, line 93: This is a little bit inefficient: https://stackoverflow.com/questions/22614237/javascript-runtime-complexity-of-array-functions

subgraphs/_Metrics/BlockSpeedForecast.ts
```diff
@@ -0,0 +1,142 @@
...
+        currentBlock.save()
+
+        // return because there is only 1 reference point.
+        return
+    }
+
+    // Add current timestamp and block numnber to array if new block is at least X blocks later than previously stored.
+    // Used to save memory and efficiency on array resizing.
+    let recentSavedBlock = _Block.load(blocks[blocks.length - INT_ONE])
+    if (currentBlockNumber - recentSavedBlock!.blockNumber > BLOCK_STORAGE_INTERVAL) {
```
this-username-is-taken, line 71: The typical pattern for handling exit conditions are:

```
if (x) {
  return
}
...
```

instead of:
```
if (x) {
  ...
} else {
  return
}

this-username-is-taken:
Great job!

I wonder if we should put this kind of helper functions in memory instead of putting the data in the store. 99% of the time the values should be the same but it'd be much more performant.

Can you move this inside of the reference subgraph instead?

Also, make sure you remove `.DS_Store`. Probably also a good idea to add it to `.gitignore`

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: Updated

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: The only time the length should be zero is for the first block. Which in this case you don't have an. interval to compute the rate.

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: Returns _blockSpeed_ now

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: Fixed comments

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: I understand that it is a little inefficient. There is not a native function for slicing off the first index while still getting O(1) pushing to the back. Would you suggest using a pointer to track the window and only slicing off the front of the array once a day or something?

subgraphs/_Metrics/BlockSpeedForecast.ts
steegecs: Fixed

subgraphs/_Metrics/BlockSpeedForecast.ts
this-username-is-taken: I think it's more accurate to use a default here (e.g. chain average) here instead of zero

steegecs:
Working on getting _updateMovingAverageBlockSpeedForecast()_ to run synchronously. The global variables are not updating as the program executes. Not found a solution that works yet.

subgraphs/_Metrics/BlockSpeedForecast.ts
this-username-is-taken: yeah, you can use a circular buffer which will give you constant time

this-username-is-taken:
What do you mean by getting it to run _synchronously_?

mehmetbeyb (commented):
subgraphs/_Metrics/BlockSpeedForecast.ts

mehmetbeyb (commented):
https://mainnet.events/agenda/speakers/1557126

steegecs:
https://stackoverflow.com/questions/23667086/why-is-my-variable-unaltered-after-i-modify-it-inside-of-a-function-asynchron

This Github issue more or less describes the issue I am having. Having trouble debugging using the methods described though.

PR #67, opened 2022-04-12
Updated Uniswap v2 to schema version 1.1.0

steegecs:


this-username-is-taken:
you need to resolve merge conflicts in the code

subgraphs/_reference_/src/common/rewards.ts
```diff
@@ -0,0 +1,198 @@
+//////////////////////////
+// AUTHOR: CHRIS STEEGE //
```
this-username-is-taken, line 2: Can you remove this? People can find the author with `git blame`. Also version is not needed for a helper lib

this-username-is-taken:
You should remove all `.DS_Store`s in the code. I've added it to .gitignore in latest commit in master so this will automatically be ignored going forward

subgraphs/_reference_/src/common/rewards.ts
steegecs: Yeah - I removed this file from this pull request. I'll change it in the other.

steegecs:
> you need to resolve merge conflicts in the code

All the merge conflict are resolved now.

steegecs:
> You should remove all `.DS_Store`s in the code. I've added it to .gitignore in latest commit in master so this will automatically be ignored going forward

Removed

PR #68, opened 2022-04-12
Feat: Balancer v2

cbrzn:


subgraphs/balancer-v2/src/common/metrics.ts
```diff
@@ -0,0 +1,200 @@
...
+    let pool = LiquidityPool.load(dex._poolIds[i])
+    if (pool) {
+      totalValueLocked = totalValueLocked.plus(pool.totalValueLockedUSD)
+      totalVolumeUsd = totalVolumeUsd.plus(pool.totalVolumeUSD)
+      totalFeesUsd = totalFeesUsd.plus(pool._totalSwapFee)
+    }
+  }
+
+  let supplySideRevenue = totalFeesUsd.times(BigDecimal.fromString("0.9"))
+  let protocolSideRevenue = totalFeesUsd.times(BigDecimal.fromString("0.1"))
```
this-username-is-taken, line 33: can this number be changed via governance or is it hardcoded?

subgraphs/balancer-v2/subgraph.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - kind: ethereum/contract
+    name: Vault
+    network: mainnet
+    source:
+      address: "0xBA12222222228d8Ba445958a75a0704d566BF2C8"
+      abi: Vault
+      startBlock: 12272146
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.5
```
this-username-is-taken, line 16: upgrade to 0.0.6

subgraphs/balancer-v2/tests/handlers.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  umaUsdcPoolAddress,
+  uma,
+  usdcWethPoolAddress,
+  bat,
+  batWethPoolAddress,
+  umaWethPoolId,
+  umaWethPoolAddress,
+} from "./state";
+
+const expectedWethPrice = "3010.039001622230703728310744540643";
```
this-username-is-taken, line 31: snake case for all constants `EXPECTED_WETH_PRICE`

subgraphs/balancer-v2/tests/handlers.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    amountOut,
+  );
+  handleSwap(swap);
+  let token = _TokenPrice.load(uma.id.toLowerCase());
+  if (token == null) throw Error("Token price is not defined");
+
+  assert.stringEquals(token.lastUsdPrice.toString(), expectedUmaPrice);
+});
+
+// test("Pool with weth and weight with low liquidity: Weth is the out token", () => {
```
this-username-is-taken, line 149: why is this commented out?

this-username-is-taken:
Can you use prettier to format?

subgraphs/balancer-v2/src/common/constants.ts
```diff
@@ -0,0 +1,172 @@
...
+export const SECONDS_PER_DAY = 60 * 60 * 24; // 86400
+export const MS_PER_DAY = new BigDecimal(BigInt.fromI32(24 * 60 * 60 * 1000));
+export const DAYS_PER_YEAR = new BigDecimal(BigInt.fromI32(365));
+export const MS_PER_YEAR = DAYS_PER_YEAR.times(new BigDecimal(BigInt.fromI32(24 * 60 * 60 * 1000)));
+
+/////////////////////////////
+///// Protocol Specific /////
+/////////////////////////////
+
+export class AddressByNetwork {
```
this-username-is-taken, line 97: Why not group by network instead? e.g.
```
class AddressByNetwork {
  vaultAddress: string;
  wethAddress: string;
  ...
}
```
and instantiate one per network?

This seems more scalable as we have more network-specific fields.

this-username-is-taken: Also, for network specific implementation, you should do something similar to Yearn: https://github.com/yearn/yearn-vaults-v2-subgraph

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (network === "polygon") {
+    network = Network.POLYGON;
+  } else {
+    network = Network.ETHEREUM;
+  }
+
+  if (protocol === null) {
+    protocol = new DexAmmProtocol(VAULT_ADDRESS.toHexString());
+    protocol.name = "Balancer V2";
+    protocol.schemaVersion = "0.0.2";
```
this-username-is-taken, line 39: update versions, include methodology version

subgraphs/balancer-v2/src/common/metrics.ts
cbrzn: nice catch - it is changed via governance

subgraphs/balancer-v2/tests/handlers.test.ts
cbrzn: This has been commented out because now it's interacting with the logic of the oracle written by Phoenix. I will also mock the contract of yearn lens to expect the correct behavior

subgraphs/balancer-v2/src/common/getters.ts
cbrzn: Will do when the latest schema update is merged :-D

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -0,0 +1,183 @@
...
+    token.symbol = fetchTokenSymbol(tokenAddress);
+    token.name = fetchTokenName(tokenAddress);
+    token.decimals = fetchTokenDecimals(tokenAddress);
+    token.save();
+  }
+  return token;
+}
+
+export function createPool(id: string, address: Address, blockInfo: ethereum.Block): void {
+  let pool = new LiquidityPool(id);
```
harsh9200, line 61: You are not initalizing `name` and `output` values in `LiquidityPool` entity.

subgraphs/balancer-v2/src/common/constants.ts
this-username-is-taken: yep, both are correct

subgraphs/balancer-v2/src/common/metrics.ts
```diff
@@ -0,0 +1,252 @@
...
+    // external source
+    if (currentBalance.times(tokenInfo.price).gt(BigDecimal.fromString("40000"))) {
+      token.lastPriceUSD = tokenInfo.price;
+      token.lastPriceBlockNumber = blockNumber;
+      token.save();
+      return;
+    }
+  }
+
+  if (getOrCreateDex().network == "MAINNET") return
```
cbrzn, line 214: Once the oracle module supports polygon and/or arbitrum, this should be modified

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -0,0 +1,285 @@
...
+} from "./constants";
+import { WeightedPool as WeightedPoolTemplate } from "../../generated/templates";
+import { WeightedPool } from "../../generated/Vault/WeightedPool";
+import { ConvergentCurvePool } from "../../generated/Vault/ConvergentCurvePool";
+
+export function getOrCreateDex(): DexAmmProtocol {
+  let protocol = DexAmmProtocol.load(VAULT_ADDRESS.toHexString());
+  if (protocol === null) {
+    let network = dataSource.network();
+    if (network === "matic") {
```
this-username-is-taken, line 32: these can be changed to `network = dataSource.network().toUpper()`

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -0,0 +1,285 @@
...
+      network = Network.POLYGON;
+    } else if (network === "arbitrum-one") {
+      network = Network.ARBITRUM;
+    } else {
+      network = Network.ETHEREUM;
+    }
+    protocol = new DexAmmProtocol(VAULT_ADDRESS.toHexString());
+    protocol.name = "Balancer V2";
```
this-username-is-taken, line 40: missing slug?

subgraphs/balancer-v2/src/common/metrics.ts
```diff
@@ -0,0 +1,252 @@
...
+  financialMetrics.dailyVolumeUSD = dailyVolumeUsd;
+  financialMetrics.cumulativeSupplySideRevenueUSD = totalRevenueGeneratedFee;
+  financialMetrics.cumulativeProtocolSideRevenueUSD = totalProtocolGeneratedFee;
+  financialMetrics.blockNumber = event.block.number;
+  financialMetrics.timestamp = event.block.timestamp;
+
+  financialMetrics.save();
+}
+
+export function updateUsageMetrics(event: ethereum.Event, from: Address): void {
```
this-username-is-taken, line 63: need to add usage hourly as well

subgraphs/balancer-v2/README.md
```diff
@@ -36,24 +39,37 @@ Sum across all Pools
...
+- Protocol: https://balancer.fi/
+- Analytics: https://dune.xyz/balancerlabs
+- Docs: https://docs.balancer.fi/
+- Smart contracts: https://github.com/balancer-labs/balancer-v2-monorepo
+- Deployed addresses: https://github.com/balancer-labs/balancer-v2-monorepo/tree/master/pkg/deployments#past-deployments
+- Official subgraph: https://github.com/balancer-labs/balancer-subgraph-v2
+
+## Build
+
+- Generate code from manifest and schema: `yarn codegen`
```
this-username-is-taken, line 69: Update this to account for mustache workflow

subgraphs/balancer-v2/subgraph.template.yaml
```diff
@@ -0,0 +1,96 @@
+specVersion: 0.0.2
+description: Balancer v2 subgraph for Messari
```
this-username-is-taken, line 2: remove the description line here

subgraphs/balancer-v2/subgraph.yaml
```diff
@@ -0,0 +1,96 @@
+specVersion: 0.0.2
+description: Balancer v2 subgraph for Messari
```
this-username-is-taken, line 2: same here

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -0,0 +1,305 @@
...
+    financialMetrics.cumulativeTotalRevenueUSD = BIGDECIMAL_ZERO;
+
+    financialMetrics.save();
+  }
+  return financialMetrics;
+}
+
+export function getOrCreateDailyUsageMetricSnapshot(event: ethereum.Event): UsageMetricsDailySnapshot {
+  // Number of days since Unix epoch
+  let id: i64 = event.block.timestamp.toI64() / SECONDS_PER_DAY;
```
this-username-is-taken, line 137: The ID should follow this format `{ # of days since Unix epoch time }-{ HH: hour of the day }`

cbrzn: Can you please share an example of another subgraph that is doing it this way? I took inspiration from Maker and Abracadabra

The reason I did not do it that way is that I couldn't find a way to get the hour based on only math. Probably I should just do something like:
```typescript
let dayId: i64 = event.block.timestamp.toI64() / SECONDS_PER_DAY;
let hourId: i64 = event.block.timestamp.toI64() / SECONDS_PER_HOUR;  

let hourlyMetricId = dayId.concat("-").concat(hourId)
```
Let me know if this makes sense to you :pray:

subgraphs/balancer-v2/src/common/getters.ts
cbrzn: Btw with arbitrum we can not use the default name, we need to use `arbitrum_one` instead of `arbitrum-one`

subgraphs/balancer-v2/src/common/getters.ts
this-username-is-taken: Understood. Let's go with `{ Smart contract address of the pool }-{ # of hours since Unix epoch time }` then. Seems like this is how Uniswap is implementing it. I guess this makes querying a lot easier because you can sort this way. I will update the schema

cbrzn: Okok cool, that's exactly how we are doing it :-D Here is the line of code: https://github.com/dOrgTech/subgraphs/blob/feat/balancer-v2/subgraphs/balancer-v2/src/common/getters.ts#L257

subgraphs/balancer-v2/src/common/getters.ts
this-username-is-taken: Can you update the logic here as well?

subgraphs/balancer-v2/src/common/metrics.ts
```diff
@@ -218,11 +219,10 @@ export function updateTokenPrice(
...
-    let token = Token.load(tokenInfo.address.toHexString());
-    if (!token) token = new Token(tokenInfo.address.toHexString());
+    const token = getOrCreateToken(tokenInfo.address);
     const index = tokenInfo.address == tokenB ? tokenBIndex : tokenAIndex;
     const currentBalance = scaleDown(pool.inputTokenBalances[index], Address.fromString(pool.inputTokens[index]));
-    // We check if current balance multiplied by the price is over 40k USD, if not,
+    // We check if current balance multiplied by the price is over 10k USD, if not,
     // it means that the pool does have too much liquidity, so we fetch the price from
     // external source
     if (currentBalance.times(tokenInfo.price).gt(BigDecimal.fromString("40000"))) {
```
this-username-is-taken, line 228: It's `10k` in the comments but `40k` in the code

subgraphs/balancer-v2/src/common/getters.ts
```diff
@@ -35,10 +35,7 @@ export function getOrCreateDex(): DexAmmProtocol {
     protocol.subgraphVersion = "1.2.0";
     protocol.methodologyVersion = "1.2.0";
```
this-username-is-taken, line 36: methodology version should align with the version in the README.

we can bump the schema version to 1.2.1 now and update the schema (just changes in comments)

subgraphs/balancer-v2/src/common/metrics.ts
cbrzn: nice catch :pray:

this-username-is-taken:
Great work!

PR #71, opened 2022-04-13
Schema update 1.2.0

this-username-is-taken:


subgraphs/_reference_/src/common/constants.ts
```diff
@@ -23,6 +26,27 @@ export namespace Network {
   export const XDAI = "XDAI";
 }
 
+// The network names corresponding to the ones in `dataSource.network()`
+// They should mainly be used for the ease of comparison.
+// See below for a complete list:
+// https://thegraph.com/docs/en/hosted-service/what-is-hosted-service/#supported-networks-on-the-hosted-service
+export namespace SubgraphNetwork {
```
steegecs, line 33: Missing "harmony" and "cronos"

this-username-is-taken: yeah, I don't think these two are supported by The Graph yet

this-username-is-taken: see https://thegraph.com/docs/en/hosted-service/what-is-hosted-service/#supported-networks-on-the-hosted-service

this-username-is-taken: I don't mind removing them from the list above

PR #77, opened 2022-04-15
compound - fix apy calcs

dmelotik:
Fix APY rate calcs and removed unnecessary totalDepositUSD/totalBorrowUSD calcs

See APY math here: https://compound.finance/docs#protocol-math


0xbe1:
there can be a function to abstract away the logic, so that we don't need to repeat the code for both borrow and supply @dmelotik

PR #81, opened 2022-04-15
compound fixes / updates

dmelotik:
- [x] Add power function
- [x] utilize rewards.ts
- [x] make gen types default and remove line in .gitignore
- [x] Change stableBorrowRate to null
- [x] fix liquidationPenalty
- [x] fix interest accruals
- [x] fix failed prod issue
- [x] get subgrpah running locally to fork published subgraph
- [x] fork prod subgraph and ensure error fixes
- [x] fix liquidation asset to cToken
- [x] snapshot revenues are cumulative
- [x] add schema 1.2.0
- [x] fix revenue issue

Deliverables:
- Add helper method for APY conversions
- Utilizes rewards.ts / updates method header
- added Big Decimal power function to _reference_
- `npm run codegen` will output to the default location / remove unnecessary line in .gitignore
- change stableBorrowRate to null
- fix liquidationPenalty so it saves
- fix liquidation asset to cToken
- snapshot revenues to cumulative
- remove long form Array creation (unnecessary)
- add schema 1.2.1 and resolve


subgraphs/_reference_/src/common/utils/numbers.ts
```diff
@@ -8,6 +8,14 @@ export function bigIntToBigDecimal(quantity: BigInt, decimals: i32 = 18): BigDec
   );
 }
 
+export function powerBigDecimal(base: BigDecimal, power: i32): BigDecimal {
```
0xbe1, line 11: I believe pow is optimized so that it runs faster than the for-loop-times below

Here're how I do it
```ts
// Formula: check out "Calculating the APY Using Rate Per Block" section https://compound.finance/docs
function convertRatePerBlockToAPY(ratePerBlock: BigInt): BigDecimal {
  let a = mantissaFactorBI
  let b = ratePerBlock.times(BLOCKS_PER_DAY).plus(mantissaFactorBI)
  return pow(a, b, 365).minus(BIGDECIMAL_ONE).times(BIGDECIMAL_HUNDRED)
}

// (a/b)^n
function pow(a: BigInt, b: BigInt, n: i32): BigDecimal {
  return a.pow(n).toBigDecimal().div(b.pow(n).toBigDecimal())
}
```

this-username-is-taken: yeah, powering can generally be done in logarithmic time: https://stackoverflow.com/questions/101439/the-most-efficient-way-to-implement-an-integer-based-power-function-powint-int

dmelotik: My struggle here is that blocksPerDay is a BigDecimal -- any thoughts on how to tackle this?

0xbe1: @dmelotik cast blocks_per_day into a bigdecimal then 🤔 i believe that's acceptable

dmelotik: @0xbe1 do you mean a BigInt?  - blocksPerDay is already a BigDecimal.

0xbe1: yeah i mean BigInt

0xbe1: turns out the approach I propose doesn't work because BigInt.pow(n) expects n between 0 and 255, while 365 is out of range 🤒

dmelotik: @0xbe1 see my latest solution and lmk your thoughts

0xbe1: we could keep the `pow` approach after graphprotocol fix their gas estimation on `pow`

subgraphs/compound-v2/src/mappings/helpers.ts
```diff
@@ -375,7 +375,9 @@ export function accrueInterestsOnBalances(market: Market, blockDifference: BigIn
     .times(market.depositRate.div(blocksPerYear))
     .times(blockDifference.toBigDecimal());
   let newInterestBigInt = BigInt.fromString(accruedSupplyInterest.truncate(0).toString());
-  market.inputTokenBalances[0] = market.inputTokenBalances[0].plus(newInterestBigInt);
+  let inputTokenBalance = new Array<BigInt>();
+  inputTokenBalance.push(market.inputTokenBalances[0].plus(newInterestBigInt));
+  market.inputTokenBalances = inputTokenBalance;
```
0xbe1, line 380: one-liner that applies to all entity array

```
market.inputTokenBalances = [market.inputTokenBalances[0].plus(newInterestBigInt)]
```

dmelotik: does this (https://thegraph.com/docs/en/developer/assemblyscript-api/#updating-existing-entities) not apply here?

I've also been doing the longhand way b/c of the comments in ambitious subgraphs pt 2: https://www.youtube.com/watch?v=1-8AW-lVfrA&t=3174s

this-username-is-taken:
Great work! Thanks for all the fixes!

PR #88, opened 2022-04-18
Major schema update (1.2.0)

this-username-is-taken:
### Common

- Added prefixes to diambiguate quantitative fields:
  - *cumulative*: sum of all historical data from day 1 up to this point. E.g. `cumulativeDepositUSD` means all deposits has ever been made to this protocol/pool.
  - *daily/hourly*: this only applies to snapshots and represents the sum of the snapshot interval (i.e. daily aggregate). E.g. `dailyActiveUsers` means all unique active users on a given day, up till now.
  - All other quantitative field indicates a spot balance. In other words, the value at this point in time. E.g. `totalValueLockedUSD` means the total TVL of the protocol/pool as of now.
- Updated chain enums to match `dataSource.network()`.
- Added `lastPriceUSD` and `lastPriceBlockNumber` as optional fields in the `Token` entity for tracking prices.
- Refactored `RewardToken` to include `Token` as a field.
- For snapshots, now saving both `cumulative` values and `daily` aggregates.
- Now saving hourly snapshots for usage metrics and pool data.

### DEX-AMM

- Added more detailed usage metrics: deposit/withdraw/swap count.
- Added `inputTokenWeights: [BigDecimal!]!` to track pool composition.
- Added per-token volume `dailyVolumeByTokenUSD: [BigDecimal!]!` to `LiquidityPoolDailySnapshot`.
- Named `PoolDailySnapshot` to `LiquidityPoolDailySnapshot` for consistency.
- Added `stakedOutputTokenAmount: BigInt` to `LiquidityPool` and `LiquidityPoolDailySnapshot`.

### Lending Protocol

- Added more detailed usage metrics: deposit/withdraw/repay/liquidate count.
- Now aggregating deposit/borrow/liquidation USD values.
- Added `InterestRate` entity and made `rates` into an array.
- Added `mintedTokens` and `mintedTokenSupplies` to the `LendingProtocol` entity.
- Added `exchangeRate: BigDecimal` to `Market` and `MarketDailySnapshot` entity.
- Assume only a single input token (i.e. changed `inputTokens` to `inputToken`).
- Updated comments to the `Liquidate` event for clarity.

### Yield Aggregator

- Added more detailed usage metrics: deposit/withdraw count.
- Assume only a single input token (i.e. changed `inputTokens` to `inputToken`).

### Generic

- Added `stakedOutputTokenAmount: BigInt`


docs/Schema.md
```diff
@@ -1,5 +1,44 @@
...
+  - hash/address: `String`
+  - block height: `BigInt`
+  - timestamp: unix timestamp in `BigInt`
+  - token amount: All token amounts should be BigInt to preserve precision (i.e. in wei)
+  - dollar amount: All USD amounts (including prices) should be BigDecimal
+- Certain prefixes may be used to indicate a particular type of value:
+  - *current*: indicates this is a spot balance. In other words, the value as of this point in time. E.g. `currentTvlUSD` means the total TVL of the protocol/pool as of now.
+  - *cumulative*: sum of all historical data from day 1 up to this point. E.g. `cumulativeDepositUSD` means all deposits has ever been made to this protocol/pool.
+  - *daily*: this only applies to snapshots and represents the sum of the snapshot interval (i.e. daily aggregate). E.g. `dailyActiveUsers` means all unique in the past day.
+- Use plurals when referring to multiple items, ususally stored as an array (e.g. tokens, balances).
```
cbrzn, line 33:
```suggestion
- Use plurals when referring to multiple items, usually stored as an array (e.g. tokens, balances).
```

schema-dex-amm.graphql
```diff
@@ -81,13 +82,14 @@ interface Protocol {
...
+  currentTvlUSD: BigDecimal!
 
-  totalValueLockedUSD: BigDecimal!
+  " # of cumulative unique users "
+  cumulativeUniqueUsers: Int!
 
-  " Total volume in USD "
-  totalVolumeUSD: BigDecimal!
+  " All historical volume in USD "
+  cumulativeVolumeUSD: BigDecimal!
```
cbrzn, line 92: I have a doubt regarding this one: This is all historical traded volume, right?

cbrzn: none, you just answered my doubt :) thanks (I just wanted to make sure I understood this)

dmelotik:
I believe @steegecs mentioned this too.

I like the descriptors prefixed on the field names. This is a small thing, but I do like `totalVolumeUSD` over `cumulativeVolumeUSD`
And if we want a field only to calculate for a given day then `dailyVolumeUSD` over `currentVolumeUSD`

So I would propose `daily` prefix for fields that we want just that given days metrics over `current` - `current` could be interpreted as total at that given time without the comment. And `total` speaks for itself - the total amount at that point in time.

steegecs:
>


> I believe @steegecs mentioned this too.
> 
> I like the descriptors prefixed on the field names. This is a small thing, but I do like `totalVolumeUSD` over `cumulativeVolumeUSD` And if we want a field only to calculate for a given day then `dailyVolumeUSD` over `currentVolumeUSD`
> 
> So I would propose `daily` prefix for fields that we want just that given days metrics over `current` - `current` could be interpreted as total at that given time without the comment. And `total` speaks for itself - the total amount at that point in time.

I still have this opinion as well about fields that use cumulative. Like the revenues, volumes, and unique users. But everything looks good to go if you want to leave that.

this-username-is-taken:
> I still have this opinion as well about fields that use cumulative. Like the revenues, volumes, and unique users. But everything looks good to go if you want to leave that. It seems more accurate this way. If you compare, for example, all historical volume and volume for a day, under the current paradigm you'd have cumulativeVolumeUSD and dailyVolumeUSD.
> 
> They are both cumulative however. One throughout the history of the protocol and the other over the period of a day. So it would be more accurate to say totalCumulativeVolumeUSD and dailyCumulativeVolumeUSD. But when you drop "cumulative' it still gets the point across clearly without the extra jargon.

The main thing here is I want to avoid `total` since we have `totalValueLocked`, which is a point-in-time value and not an aggregate. Having `totalVolume` and `totalValueLocked` can be slightly confusing.

schema-lending.graphql
```diff
@@ -31,29 +32,96 @@ type Token @entity {
...
+enum InterestRateSide {
+  " Interest rate accrued by lenders "
+  LENDER
+
+  " Interest rate paid by borrowers "
+  BORROWER
+}
+
+type InterestRate @entity {
+  " { Interest rate side }-{ Interest rate type }-{ Market ID }-{ Counter } "
```
0xbe1, line 106: what does Counter mean?

schema-lending.graphql
```diff
@@ -77,203 +145,273 @@ interface Protocol {
...
-  # and inputToken balances are used to calculate returns
+  " Token that need to be deposited to take a position in protocol "
+  inputToken: Token!
 
-  " Tokens that need to be deposited to take a position in protocol. e.g. WETH and USDC to deposit into the WETH-USDC pool "
-  inputTokens: [Token!]!
-
-  " Token that is minted to track ownership of position in protocol "
+  " Token that is minted to track ownership of position in protocol (e.g. aToken, cToken). Leave as null if doesn't exist "
   outputToken: Token
```
0xbe1, line 376: have we found a lending protocol that does not have outputToken? if not, maybe leave it required?

(if this makes sense to you, also change exchangeRate to required)

0xbe1 (commented):
Have gone through schema-lending (that I am most familiar with) and it looks good, just small comments.

Higher-level question: Do we decide to leave fields in `type Liquidate` generic? Such as `from` and `to`. I proposed using more specific names and we had discussions on Discord, would love to hear your thoughts here.

schema-lending.graphql
this-username-is-taken: You can have more than one rate. I will add doc

schema-lending.graphql
this-username-is-taken: yeah, none of the CDPs have outputToken

schema-yield.graphql
```diff
@@ -113,23 +169,44 @@ type YieldAggregator implements Protocol @entity {
...
+  " # of cumulative unique users "
+  cumulativeUniqueUsers: Int!
+
+  ##### Snapshots #####
+
+  " Daily usage metrics for this protocol "
+  dailyUsageMetrics: [UsageMetricsDailySnapshot!]! @derivedFrom(field: "protocol")
+
+  " Hourly usage metrics for this protocol "
+  hourlyUsageMetrics: [UsageMetricsDailySnapshot!]! @derivedFrom(field: "protocol")
```
harsh9200, line 204: This value should be `UsageMetricsHourlySnapshot`.

schema-generic.graphql
```diff
@@ -120,44 +221,71 @@ type FinancialsDailySnapshot @entity {
...
 type Pool @entity {
   " Smart contract address of the pool "
   id: ID!
 
   " The protocol this pool belongs to "
   protocol: Protocol!
 
-  " Tokens that need to be deposited to take a position in protocol. e.g. WETH and USDC to deposit into the WETH-USDC pool "
+  " Name of the pool (e.g. Curve.fi DAI/USDC/USDT) "
+  name: String
```
harsh9200, line 236: Add exclamation mark !!

schema-generic.graphql
```diff
@@ -120,44 +221,71 @@ type FinancialsDailySnapshot @entity {
...
+
+  " Symbol of liquidity pool (e.g. 3CRV) "
+  symbol: String
```
harsh9200, line 239: Add exclamation mark !!

schema-yield.graphql
```diff
@@ -221,68 +320,76 @@ type Vault @entity {
...
-  outputTokenSupply: BigInt!
+  outputTokenSupply: BigInt
 
   " Price per share of output token in USD "
-  outputTokenPriceUSD: BigDecimal!
+  outputTokenPriceUSD: BigDecimal
 
   " Amount of input token per full share of output token. Usually corresponds to the value of `pricePerShare` or `pricePerFullShare` in the vault contract. "
-  pricePerShare: BigInt!
+  pricePerShare: BigInt
```
harsh9200, line 368: This should be `BigDecimal`

schema-generic.graphql
this-username-is-taken: why?

this-username-is-taken: not sure if we can assume every pool has a name

schema-yield.graphql
this-username-is-taken: good point

schema-generic.graphql
this-username-is-taken: same as above, not sure if we can make this assumption. A pool doesn't need to be erc20

this-username-is-taken:
See fixes here: https://github.com/messari/subgraphs/pull/106

PR #103, opened 2022-04-22
Fix decimals bug in `uniswapRouter` and `sushiswapRouter`

harsh9200:


subgraphs/_reference_/src/Prices/common/constants.ts
```diff
@@ -12,6 +12,7 @@ export const BIGINT_TEN_THOUSAND = BigInt.fromI32(10000);
 
 export const BIGDECIMAL_ZERO = new BigDecimal(BIGINT_ZERO);
 
+export const DEFAULT_USD_DECIMALS = BIGINT_TEN.pow(6);
```
this-username-is-taken, line 15: Should be USDC Decimals right?

PR #114, opened 2022-04-26
Abracadabra

tannishmango:
update to schema 1.2.0
add new chains:
- arbitrum
- avalanche
- bnb
- fantom
- mainnet


subgraphs/abracadabra/config/bnb.json
```diff
@@ -0,0 +1,11 @@
+{
+  "startBlock": "12750809",
+  "network": "bnb",
```
this-username-is-taken, line 3: does this actually work? Shouldn't it be `bsc`?

subgraphs/abracadabra/src/common/constants.ts
```diff
@@ -13,7 +13,7 @@ export namespace Network {
   export const BSC = "BINANCE_SMART_CHAIN";
```
this-username-is-taken, line 13: You should update this file with the latest in the reference subgraph

subgraphs/abracadabra/config/bnb.json
tannishmango: just updated this

subgraphs/abracadabra/src/common/getters.ts
```diff
@@ -205,7 +205,7 @@ export function getOrCreateLendingProtocol(): LendingProtocol {
   LendingProtocolEntity.subgraphVersion = "0.0.6";
```
this-username-is-taken, line 205: subgraph version should be at least 1.0.0 since it's going to be deployed now

this-username-is-taken (dismissed):

subgraphs/abracadabra/src/common/getters.ts
tannishmango: fixed

PR #118, opened 2022-04-26
Add multi chain support to yearn-v2

harsh9200:
Chains Added: 
- Arbitrum One
- Fantom


subgraphs/yearn-v2/src/common/initializers.ts
```diff
@@ -241,3 +243,73 @@ export function getOrCreateVaultsHourlySnapshots(
...
+    vault.outputTokenSupply = constants.BIGINT_ZERO;
+
+    vault.outputTokenPriceUSD = constants.BIGDECIMAL_ZERO;
+    vault.pricePerShare = constants.BIGDECIMAL_ZERO;
+
+    vault.createdBlockNumber = block.number;
+    vault.createdTimestamp = block.timestamp;
+
+    vault.totalValueLockedUSD = constants.BIGDECIMAL_ZERO;
+    vault.totalValueLockedUSD = constants.BIGDECIMAL_ZERO;
```
this-username-is-taken, line 282: same?

harsh9200: yeah my bad, good catch

PR #123, opened 2022-04-28
Corrected Protocol Revenue calculation

KG-Messari:


subgraphs/inverse-finance/README.md
```diff
@@ -1,40 +1,44 @@
...
 
 ### Total Value Locked (TVL) USD
 
 Sum across all Pools:
 
-`Insert Calculation`
+`Sum of Deposits on Anchor`
 
-<Add notes to consider if any - delete if none>
+This includes INV staking
```
this-username-is-taken, line 17: Interesting. Why does this include INV staking? I thought we are not including staked tokens in TVL.

PR #124, opened 2022-04-28
add Methodology to README - ApeSwap, Bancor, Dodo

DustinJamesT:


subgraphs/apeswap/README.md
```diff
@@ -0,0 +1,73 @@
+# ApeSwap Subgraph
+## Calculation Methodology v0.0.1
```
this-username-is-taken, line 2: We should change the version to v1.0.0 if it's completed, in order to differentiate from a version that's still a work in progress

subgraphs/bancor-v2/README.md
```diff
@@ -1,9 +1,70 @@
...
+
+`Deposits`
+
+`Withdraws`
+
+###  Reward Token Emissions Amount
+
+To be added
+
+###  Protocol Controlled Value
```
this-username-is-taken, line 57: If it doesn't exist, we should probably just say doesn't exist. So that we won't need to revisit it in the future

this-username-is-taken:
Thanks Dustin! Just provided a couple minor feedback

DustinJamesT (commented):
fixed comments

DustinJamesT (commented):
comments fixed

PR #131, opened 2022-04-30
Update yearn-v2 revenue logic and Add missing mappings

harsh9200:


subgraphs/yearn-v2/src/modules/Vault.ts
```diff
@@ -42,7 +42,7 @@ export function _NewVault(
     vault.createdTimestamp = block.timestamp;
 
     vault.totalValueLockedUSD = constants.BIGDECIMAL_ZERO;
-    vault. = constants.BIGDECIMAL_ZERO;
+    vault.lastReport = constants.BIGINT_ZERO;
```
this-username-is-taken, line 45: lol, how did I miss this one

PR #132, opened 2022-04-30
Add Saddle Finance subgraph

spitko:
- External rewards are not tracked, e.g staking LP tokens in Alchemix/FRAX protocols.
- All pegged value asset prices are set to underlying asset price
- LP tokens in metapool are "unwrapped" to underlying tokens
- LP tokens deposited to a metapool are currently counted twice in protocol TVL
- Added two custom fields to schema: Token._pool and LiquidityPool._basePool


subgraphs/saddle-finance/src/entities/protocol.ts
```diff
@@ -0,0 +1,229 @@
...
+  SECONDS_PER_DAY,
+  SECONDS_PER_HOUR,
+  ZERO_ADDRESS,
+} from "../utils/constants";
+
+export function getOrCreateProtocol(): DexAmmProtocol {
+  let protocol = DexAmmProtocol.load(ZERO_ADDRESS);
+  if (!protocol) {
+    // Zero address because no registry
+    protocol = new DexAmmProtocol(ZERO_ADDRESS);
```
this-username-is-taken, line 31: We should use a more meaningful address. Maybe address of the first deployer. Main reason is that at one point we may have a "mega" subgraph that'll contain all DEX-AMMs, and there needs to be a way to differentiate their IDs

PR #136, opened 2022-05-03
yearn v2 fix multi chain 

harsh9200:


subgraphs/yearn-v2/subgraph.template.yaml
```diff
@@ -32,7 +32,7 @@ dataSources:
         ############## Price Oracle ###############
         ###########################################
         # ERC20
-        - name: ERC20
+        - name: _ERC20
```
this-username-is-taken, line 35: Can we change this to a more descriptive name? e.g. `PriceOracleERC20`

PR #153, opened 2022-05-07
Compound revenue fix

dmelotik:
Updated reveenue to calculate when `AccrueInterest` is emitted. This will also trigger `TotalBorrowUSD` updates.


0xbe1:
lgtm

PR #157, opened 2022-05-08
update yield ag schema

dmelotik:
Allow vaults to have multiple input tokens by making the vault id `{ Smart contract address of the vault }-{ Smart contract address of the inputToken }`


dmelotik:
unnecessary - see change reflected in raris schema

PR #158, opened 2022-05-08
use interestAccumulated

0xbe1:


dmelotik:
lgtm

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1264,7 +1231,7 @@ function updateMarket(
     blockTimestamp.toI32()
   );
   snapshot._dailyTotalRevenueUSD =
-    snapshot._dailyTotalRevenueUSD.plus(totalRevenueUSDDelta);
+    snapshot._dailyTotalRevenueUSD.plus(interestAccumulated);
```
dmelotik, line 1234: why do we need these extra fields for revenue in market snapshots and markets?

PR #162, opened 2022-05-10
rari fuse v1

dmelotik:
Create rari fuse v1 subgraph in compound-forks folder

TODO:

- [x] 1 LendingProtocol and make each market have a pool parameter (ie, one `FinancialSnapshot`)
- [x] add rewards for some pools
- [x] Add multi-chain config (Arbitrum in our case)
- [x] Test everything

Issues:

Mainnet:
- find block where sOHM was migrated to gOHM and change pools > fixed
- Rewards still not working > fixed - waiting for test subgraph to sync

Arbitrum: 
- 0xc0c997227922004da3a47185ac2be1d648db0062 the price of this asset is probably manipulated b/c small number of holders. However, this is the correct price on rari's website too. Probably a non-issue. just inflates the TVL. Potential fix: update the price and TVL recalc each time the protocol TVL is calculated.


0xbe1 (commented):
@dmelotik  great work!

fuse is indeed more complicated than existing compound forks, it is like a pool of compound protocols, if I understand correctly. great to see the skeleton code is somewhat flexible to handle the flexibility.

I have a question regarding the `price` part though. In existing compound forks, price is derived through an oracle built into the compound protocol. can you help me understand what makes fuse different as it leverages the price lib to do this?

thank you!

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1194,6 +1196,12 @@ export function updateMarket(
     underlyingToken.decimals
   );
 
+  log.warning("Updated price for {}, mantissa: {}, price: ${}", [
```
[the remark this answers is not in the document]
dmelotik: good catch

dmelotik:
@0xbe1 thanks for reviewing, and that is a good question. It probably needs more investigation and discussion, the issue I seemed to be having is that it seemed like fuse had widespread price oracle manipulation. 

I think i will try going back to their price oracle and sync it again to see the difference and if I can pinpoint what pools and see if we can make one off changes just for those pools.

The other question would be, since the price oracle is the "official price" is the manipulated price really what we should be displaying?

And another question I was thinking about, should we update price of every market each time we calculate the protocol TVL? The reason would be to normalize the price on pools that aren't touched very often, example being the $100m fuse arbitrum pool that only has one deposit.

subgraphs/compound-forks/schema.graphql
```diff
@@ -1010,3 +1012,20 @@ type _CircularBuffer @entity {
   " The current calculated number of blocks per day based on calculated block speed "
   blocksPerDay: BigDecimal!
 }
+
+type _FusePool @entity {
```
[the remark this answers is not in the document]
dmelotik: done

0xbe1:
@dmelotik 

1. what does "fuse had widespread price oracle manipulation" actually mean - can you plz share more details?
2. "should we update price of every market each time we calculate the protocol TVL" - is there a good way to cache price maybe that's desirable?

subgraphs/compound-forks/fuse-v1/src/constants.ts
```diff
@@ -61,6 +61,6 @@ export const GOHM_ADDRESS = "0x0ab87046fBb341D058F17CBC4c1133F25a20a52f";
 
 export const PROTOCOL_NAME = "Fuse";
```
this-username-is-taken, line 62: Let's call it `Rari Fuse`

subgraphs/compound-forks/fuse-v1/src/constants.ts
```diff
@@ -61,6 +61,6 @@ export const GOHM_ADDRESS = "0x0ab87046fBb341D058F17CBC4c1133F25a20a52f";
...
 export const PROTOCOL_SLUG = "fuse";
```
this-username-is-taken, line 63: `rari-fuse`

subgraphs/compound-forks/fuse-v1/src/constants.ts
```diff
@@ -61,6 +61,6 @@ export const GOHM_ADDRESS = "0x0ab87046fBb341D058F17CBC4c1133F25a20a52f";
...
-export const SUBGRAPH_VERSION = "1.0.13";
+export const SUBGRAPH_VERSION = "1.0.14";
 export const SCHEMA_VERSION = "1.2.1";
```
this-username-is-taken, line 65: 1.3.0

this-username-is-taken:
Everything else looks fine. As per discussion in Discord, let's go with the simple implementation (Rari's oracle).

subgraphs/compound-forks/fuse-v1/src/constants.ts
dmelotik: done

subgraphs/compound-forks/fuse-v1/src/constants.ts
dmelotik: done

subgraphs/compound-forks/fuse-v1/src/constants.ts
dmelotik: done

subgraphs/compound-forks/rari-fuse/src/mappings.ts
```diff
@@ -0,0 +1,942 @@
...
+    tryFuseFeeMantissa.reverted ? BIGINT_ZERO : tryFuseFeeMantissa.value,
+    RariFee.FUSE_FEE,
+    event.params.cToken.toHexString()
+  );
+
+  // get admin fee - pool owners (admin) collect this (ie, protocol revenue)
+  let tryAdminFeeMantissa = cTokenContract.try_adminFeeMantissa();
+  updateOrCreateRariFee(
+    tryAdminFeeMantissa.reverted ? BIGINT_ZERO : tryAdminFeeMantissa.value,
+    RariFee.ADMIN_FEE,
```
this-username-is-taken, line 263: Hmm, not sure if admin fee should be considered protocol revenue. Probably should check with @pietydespises

this-username-is-taken:
Great work 🔥🔥

Don't forget to deploy and update the deployment status in the README after you confirm the `adminFee`

PR #164, opened 2022-05-10
Refactor and features

MichaelC1999:


dashboard/src/common/deployment.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+export const Deployment = ({ networkName, deployment, status }: DeploymentProps) => {
+  const navigate = useNavigate();
+  const client = useMemo(() => NewClient(deployment.URL), [deployment.URL]);
+  const { data, error } = useQuery(ProtocolQuery, {
+    client,
+  });
+
+  const protocol = useMemo(() => data?.protocols[0], [data]);
+  // Probably still want to render if we couldn't fetch the status.
```
RickyRoller, line 23: It would be good to handle this situation so that there isn't a regression in functionality.
I left the logic you had in the deployments page commented out over there so it'll need to be brought into here.

PR #167, opened 2022-05-11
Update compound-fork base code

dmelotik:
Deliverable:

- fix small dependecy things
- update package.json


subgraphs/compound-forks/Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: codegen
 codegen:
-	graph codegen $(subgraph)/subgraph.yaml -o $(subgraph)/generated
+	graph codegen $(subgraph)/subgraph.yaml -o ./generated
```
0xbe1, line 3: i don't think we want to change this ... we want `make codegen` to generate code in specific protocols, not the root

we only need to run `graph codegen bastion-protocol/subgraph.yaml -o ./generated` once to generate code under root, while we run `make codegen subgraph=xxx` very often when developing a fork

subgraphs/compound-forks/.prettierrc.json
```diff
@@ -1 +1,8 @@
-{}
```
0xbe1, line 1: i think both are fine as long as they are consistent. the current code is also formatted, just under the default prettier schema

subgraphs/compound-forks/package.json
```diff
@@ -1,12 +1,18 @@
...
-    "prettier": "^2.6.2"
+  "dependencies": {
+    "@graphprotocol/graph-cli": "^0.29.2",
+    "@graphprotocol/graph-ts": "0.26.0",
+    "@typescript-eslint/eslint-plugin": "^5.16.0",
+    "@typescript-eslint/parser": "^5.16.0",
+    "eslint": "^8.11.0",
+    "eslint-config-prettier": "^8.5.0",
+    "prettier": "^2.6.0",
+    "typescript": "^4.6.2"
```
0xbe1, line 16: do we really need all of these?

0xbe1:
the graph-cli bump looks good, though not too sure if we want to commit both package-lock and yarn lock 🤔

dmelotik:
> the graph-cli bump looks good, though not too sure if we want to commit both package-lock and yarn lock 🤔

based on this convo, i don't think this is necessary. Gonna close this and just update the packages in my fuse-v1 PR since it is so small

PR #193, opened 2022-05-17
Dashboard v1.0.1

MichaelC1999:
DASHBOARD V1.0.1
-Debugging unaccounted for cases/random errors
-Added APR as a separate field from rewardTokenEmissionsUSD
-Reordering Charts
-Dashboard versioning
-capitalEfficiency timeseries on protocol level (financial daily snaps)
-baseYield on DEX liquidity pools
-Apply comparison test to protocols
-Refactored time series chart conditionals
-Rewrote code pointed out in "Dashboard code style improvements" issue raised by Ricky
-Refactored some elements into separate components


dashboard/src/interfaces/tabs/ProtocolTab.tsx
```diff
@@ -123,6 +124,45 @@ function ProtocolTab(
...
+        if (renderedField) {
+          list[entityName][entityField] = "Included";
+        } else {
+          const extrapolatedFields = Object.keys(dataFields).filter((df: string) => {
+            return df.includes(entityField);
+          });
+          if (extrapolatedFields?.length > 0) {
+            list[entityName][entityField] = "Array Included";
+          } else {
+            const req = '!' === entitiesData[entityName][entityField].split("")[entitiesData[entityName][entityField].split("").length - 1]
```
RickyRoller, line 143: is this just checking if the text has an `!`?

MichaelC1999: This checks whether or not that entity field is listed as required in the schema or not. I did it this way to avoid the redundancy of embedding another level of key/value pairs to have this property as a boolean. The text being checked is a field type followed by an exclamation point if required. Sometimes the field type specified on the schema has to be checked to determine some conditions about how that field should be handled.

RickyRoller: The reason I bring it up is that it looks like a good use case for some regex.
`!!/!/g.exec(entitiesData[entityName][entityField])`
Should return true/false if there is an `!` in the string

RickyRoller: Not critical but it was hard to read what it was doing so I wanted to bring it up

RickyRoller (approved):
Great improvements on breaking things out 👍  It's looking much easier to tell what's going on

PR #195, opened 2022-05-18
Vesper finance V3

jaideepdas:
Schema : 1.2.1
vesper-meta: 2.60.0

Note:
As vesper doesn't have a pool register. We are using their published npm package called `vesper-meta`. This package provides a list of pools and controllers. We are indexing all version 3 and above pools along with the orbit pools for mainnet.


jaideepdas:
formula updated so closing this PR

PR #213, opened 2022-05-21
add cream

0xbe1:


subgraphs/compound-forks/cream/src/mapping.ts
```diff
@@ -0,0 +1,186 @@
...
+    unitPerYear
+  );
+  _handleAccrueInterest(updateMarketData, comptrollerAddr, event);
+}
+
+function getOrCreateProtocol(): LendingProtocol {
+  let comptroller = Comptroller.bind(comptrollerAddr);
+  let protocolData = new ProtocolData(
+    comptrollerAddr,
+    "Cream Finance",
```
this-username-is-taken, line 177: Can you change name to "CREAM Finance" and slug to "cream-finance"? We should also change the folder name to align with slug

0xbe1: how about just "Cream" and "cream"?

PR #217, opened 2022-05-22
Fix Liquity cumulativeDepositUSD and LTV

spitko:


this-username-is-taken:
These rates should be in percentages (i.e. x100)

image

spitko:
> These rates should be in percentages (i.e. x100)
> 
> image

Yes, fixed those too 👍

PR #218, opened 2022-05-22
Add QiDao subgraph

spitko:
- Similar to Liquity, there is no interest rate, only a one-time closing fee, which is taken from collateral during repay. Market interest rate is set to 0.
- MAI price is fixed to $1.0.
- Vault ownership is tracked using NFTs, which can be transferred between addresses. This is not tracked in the subgraph.
- Vault reward token (Qi) emissions are calculated off-chain from snapshots and are not tracked.
- Masterchef contracts are not tracked.
- [Anchor](https://app.mai.finance/anchor) is not tracked.


this-username-is-taken:
Would you be able to work with @steegecs on adding CI/CD workflow into the QiDao subgraph?

this-username-is-taken:
Great job! 🚀🚀🚀

PR #223, opened 2022-05-22
Fix null input token entity

tannishmango:
Issue related to unexpected null input token in market entity


subgraphs/makerdao/schema.graphql
```diff
@@ -460,7 +460,7 @@ type Market @entity {
   liquidationPenalty: BigDecimal!
 
   " Token that need to be deposited to take a position in protocol "
-  inputToken: Token!
+  inputToken: Token
```
this-username-is-taken, line 463: This change breaks the schema. Why do you need this?

this-username-is-taken:
Can you update the subgraph version?

subgraphs/makerdao/schema.graphql
tannishmango: i changed it back

Issue #237, opened 2022-05-26
Compound & Forks not updating isActive / canUseAsCollateral / canBorrowFrom


dmelotik:
These fields are not being updated when they should. And if they are moved to false, there is no mechanism that is moving them out of false when the market is open again.

Sample query to see behavior:

`query MyQuery {
  markets {
    id
    name
    canBorrowFrom
    canUseAsCollateral
    isActive
    _reserveFactor
  	rates {
      id
      rate
    }
  }
}`

PR #251, opened 2022-05-31
Fix compound/comp-fork `market.canUseAsCollateral` update

dmelotik:
Fixing https://github.com/messari/subgraphs/issues/237


0xbe1 (commented):
lgtm

PR #262, opened 2022-06-01
getSnapshotRates small refactor

0xbe1:


subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1590,25 +1589,25 @@ export function getOrElse<T>(
...
+      ]);
+      continue;
     }
 
-    // update rate to current rate
-    _rate.rate = actualRate!.rate;
-    _rate.save();
-    snapshotRates.push(_rate.id);
+    // create new snapshot rate
+    let snapshotRateId = rates[i].concat(timeSuffix);
```
dmelotik, line 1604: this will work, but we probably want a hyphen between the rate id and `timeSuffix`

dmelotik (commented):
other than that lgtm

PR #264, opened 2022-06-02
Fixed bugs in market metrics; Updated README

tnkrxyz:
Commit f930dc474ad938888653e364db6409b7312694d3 fixed bugs in `MarketHourlySnapshot` and `MarketDailySnapshot` for item `cumulativeDepositUSD`, `cumulativeBorrowUSD`, and `cumulativeLiquidateUSD`. Values of the Market entity should be used but those of the protocol were used.

README.md was updated to bring information from the earlier specification at https://github.com/messari/subgraphs/blob/f96689a5db62de1238be0ef5a42b5a19d18eb315/subgraphs/inverse-finance/README.md#total-revenue-usd  & reflect the fact that fees from the Stablizer are included in protocol revenue.

### DIFF
--- a/subgraphs/inverse-finance/README.md
+++ b/subgraphs/inverse-finance/README.md
@@ -6,28 +6,44 @@
 
 Sum across all Markets:
 
-`CToken.get_cash() * priceOracle.getUnderlyingPrice(CToken)`
+`Sum of Deposits on Anchor`
+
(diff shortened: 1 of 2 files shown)


this-username-is-taken:
Can you only commit the updates (i.e. sync to latest and branch off there)?

tnkrxyz:
That's something I was hoping to fix/improve myself, but wasn't familiar with git rebase enough to do it. It's correctly re-based now.

PR #272, opened 2022-06-03
Curve - Updated directory structure and added deployment

steegecs:


subgraphs/curve/deployment/deployment.js
```diff
@@ -0,0 +1,104 @@
+import * as protocolNetworkData from './deploymentConfigurations.json' assert {type: "json"}
```
this-username-is-taken, line 1: why are you duplicating these deployment files per subgraph? can't they be shared?

Issue #283, opened 2022-06-04
Price library arbitrum error


dmelotik:
One of the curve mappings calls Fantom's config twice instead of Arbitrum's so an unexpected null wasm error is thrown:

`Handler skipped due to execution failure, error: Mapping aborted at src/prices/routers/CurveRouter.ts, line 56, column 7, with message: unexpected null wasm backtrace: 0: 0xaf44 - <unknown>!src/prices/routers/UniswapRouter/getLpTokenTotalLiquidityUsdc 1: 0xc0f8 - <unknown>!generated/schema/Market#get:outputTokenSupply 2: 0xc5aa - <unknown>!../src/mapping/updateProtocol 3: 0xd826 - <unknown>!src/mappings/handleNewFuseFee , handler: handleAccrueInterest`

this-username-is-taken:
which subgraph are you seeing this error in? You will need to update it in that subgraph as well. Probably a good idea to just update for all subgraphs containing this code

dmelotik:
did a search of the repository, doesn't seem to be an issue for any other protocols. I guess it only lived in the library

PR #293, opened 2022-06-07
Fix Arbitrum price lib curve config

dmelotik:
Fixing https://github.com/messari/subgraphs/issues/283


this-username-is-taken:
Nice catch!

PR #308, opened 2022-06-10
Beefy Finance

RiccardoGalbusera:
Subgraph for beefy finance:
- tracking of all vaults state
- metrics for usage and value
- setup deployment on all supported chains (arbitrum, aurora, avalanche, bsc, celo, fantom, fuse, matic, moonriver)
- added prices support for all needed chains


subgraphs/beefy-finance/src/mappings/protocol.ts
```diff
@@ -0,0 +1,144 @@
...
+  beefy.cumulativeSupplySideRevenueUSD = BIGDECIMAL_ZERO; //todo
+  beefy.cumulativeProtocolSideRevenueUSD = BIGDECIMAL_ZERO; //todo
+  beefy.cumulativeTotalRevenueUSD = BIGDECIMAL_ZERO; //todo
+  beefy.cumulativeUniqueUsers = BIGINT_ZERO;
+  beefy.vaults = [vaultId];
+  beefy.dailyUsageMetrics = [updateUsageMetricsDailySnapshot(block, beefy).id];
+  beefy.hourlyUsageMetrics = [
+    updateUsageMetricsHourlySnapshot(block, beefy).id,
+  ];
+  beefy.financialMetrics = [createFirstDailyFinancialSnapshot(block, beefy).id];
```
harsh9200, line 43: you don't need to update referenced entities. they are handled automatically by `theGraph`.

subgraphs/beefy-finance/src/mappings/deposit.ts
```diff
@@ -0,0 +1,71 @@
...
+  deposit.protocol = PROTOCOL_ID;
+
+  deposit.save();
+  return deposit;
+}
+
+export function getOrCreateFirstDeposit(vault: Vault): Deposit {
+  let deposit = Deposit.load("MockDeposit" + vault.id);
+  if (!deposit) {
+    deposit = new Deposit("MockDeposit" + vault.id);
```
harsh9200, line 53: change `MockDeposit` to `deposit` and add a `-` between them.

subgraphs/beefy-finance/src/mappings/withdraw.ts
```diff
@@ -0,0 +1,66 @@
...
+  withdraw.protocol = PROTOCOL_ID;
+
+  withdraw.save();
+  return withdraw;
+}
+
+export function getOrCreateFirstWithdraw(vault: Vault): Withdraw {
+  let withdraw = Withdraw.load("MockWithdraw" + vault.id);
+  if (!withdraw) {
+    withdraw = new Withdraw("MockWithdraw" + vault.id);
```
harsh9200, line 49: change `MockWithdraw` to `withdraw` and add a `-` between them.

subgraphs/beefy-finance/src/utils/snapshots.ts
```diff
@@ -0,0 +1,77 @@
...
+  VaultHourlySnapshot,
+} from "../../generated/schema";
+import { BIGDECIMAL_ZERO } from "../prices/common/constants";
+
+export function getVaultDailyId(block: ethereum.Block, vault: Vault): string {
+  const daysSinceEpoch = getDaysSinceEpoch(block.timestamp.toI32());
+  const id = vault.id
+    .concat("-")
+    .concat(daysSinceEpoch)
+    .concat("d");
```
harsh9200, line 15: remove this line

subgraphs/beefy-finance/src/utils/snapshots.ts
```diff
@@ -0,0 +1,77 @@
...
+  return id;
+}
+
+export function getVaultHourlyId(block: ethereum.Block, vault: Vault): string {
+  const hoursSinceEpoch = getHoursSinceEpoch(block.timestamp.toI32());
+  const id = vault.id
+    .concat("-")
+    .concat(hoursSinceEpoch)
+    .concat("h");
```
harsh9200, line 24: remove this line

subgraphs/beefy-finance/src/utils/metrics.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+function getTransactionCount(
+  protocol: YieldAggregator,
+  timeframe: BigInt[]
+): BigInt {
+  return getDepositCount(protocol, timeframe).plus(
+    getWithdrawCount(protocol, timeframe)
+  );
+}
```
harsh9200, line 216: This should be handled at the transaction level. Example: [Convex Subgraph](https://github.com/messari/subgraphs/blob/master/subgraphs/convex/src/modules/Metric.ts#L14-L64)

subgraphs/beefy-finance/src/mappings/protocol.ts
```diff
@@ -0,0 +1,144 @@
...
+  ]);
+  const dailySnapshot = updateUsageMetricsDailySnapshot(block, protocol);
+  if (
+    protocol.dailyUsageMetrics[protocol.dailyUsageMetrics.length - 1] !=
+    dailySnapshot.id
+  ) {
+    protocol.dailyUsageMetrics = protocol.dailyUsageMetrics.concat([
+      dailySnapshot.id,
+    ]);
+  }
```
harsh9200, line 71: You don't need to handle referenced fields, they are automatically referenced by theGraph.

subgraphs/beefy-finance/src/mappings/vault.ts
```diff
@@ -0,0 +1,335 @@
...
+  let fee: VaultFee | null;
+  let lastSnapshot = VaultDailySnapshot.load(
+    vault.dailySnapshots[vault.dailySnapshots.length - 1]
+  );
+  if (lastSnapshot) {
+    let currentTotalRevenue = lastSnapshot.dailyTotalRevenueUSD;
+    let currentRevenueSupplySide = currentTotalRevenue;
+    let currentRevenueProtocolSide = BIGDECIMAL_ZERO;
+    for (let k = 0; k < vault.fees.length; k++) {
+      fee = VaultFee.load(vault.fees[k]);
```
harsh9200, line 223: You are only updating the fees when the vault is created, this logic will fail if the fees are updated. You should either add handlers for this or do JSON RPC API calls whenever you need performance or withdrawal fees.

harsh9200 (changes requested):
Good Work, let me know if you have any doubts

subgraphs/beefy-finance/src/mappings/protocol.ts
RiccardoGalbusera: The update function is here also to create a new snapshot if it has not been created yet.
Do you want me to update the function's name to make it clear?

harsh9200: I meant you don't need to update the `many` side of `derivedFrom` fields. The 'many' side is automatically derived. So you don't need to update `dailyUsageMetrics`, `hourlyUsageMetrics`,  `financialMetrics` here. [Ref](https://thegraph.com/docs/en/developer/create-subgraph-hosted/#:~:text=Reverse%20Lookups%23,as%20is%20practical.)

RiccardoGalbusera: ok, now I fully understood, thanks for the clarification

RiccardoGalbusera:
I have updated the pr accordingly to your comments, let me know if you have any other doubts

RiccardoGalbusera:
Hello @harsh9200, regarding the chargedFees event, I originally did not include it because not all beefy vaults emit it (especially old vaults on BSC), so I decided to calculate those values using the other data at my disposal. If you want, I can change this part of the implementation, but it's possible that we will suffer some kind of data loss in doing so.
Thanks again for the feedback.

harsh9200:
Hello @RiccardoGalbusera, the current implementation for calculating `supplySideRevenue` and `protocolSideRevenue` is not correct. You are subtracting the percentage fees from the total harvested amount. whereas the fees are paid in `WMATIC` (in the case of polygon chain) or `BNB` (in the case of BSC) depending on the chain. You should utilize the `ChargedFees` event for more accurate fees and in case there is no such event in the strategy you can always reverse engineer the fees logic present in the contract source code.

subgraphs/beefy-finance/src/mappings/protocol.ts
```diff
@@ -0,0 +1,172 @@
...
+  vault.save();
+  let fee: VaultFee | null;
+  for (let i = 0; i < fees.length; i++) {
+    fee = VaultFee.load(fees[i]);
+    if (fee && fee.feeType == "WITHDRAWAL_FEE") {
+      const revenue = withdrawnAmount
+        .toBigDecimal()
+        .times(fee.feePercentage)
+        .div(BIGDECIMAL_HUNDRED)
+        .div(BIGINT_TEN.pow(token.decimals as u8).toBigDecimal());
```
harsh9200, line 138: This revenue should be in USD.

harsh9200 (approved):
Good Work!

this-username-is-taken:
Appreciate all the hard work guys!

PR #316, opened 2022-06-14
Added blacklist and minimum liquidity threshold configurations for pricing

steegecs:


subgraphs/uniswap-forks/protocols/quickswap/config/networks/matic/matic.ts
```diff
@@ -81,4 +81,10 @@ export class QuickswapMaticConfigurations implements Configurations {
   getUntrackedPairs(): string[] {
     return toLowerCaseList([]);
   }
+  getUntrackedTokens(): string[] {
+    return [];
+  }
+  getMinimumLiquidityThreshold(): BigDecimal {
+    return BigDecimal.fromString("100000");
```
this-username-is-taken, line 88: Might be better if we have this as a constant in a shared file?

steegecs: Yeah - I can put the value in the protocol specific constants folder. Then if more flexibility is wanted, I could add constants.ts to the particular network folder.

steegecs: Added constants for 100,000 and 400,000 into the shared constants file.

this-username-is-taken:
Great work! Appreciate all the debugging!

PR #263, opened 2022-06-02
dforce subgraph

tnkrxyz:
This is a work-in-progress PR for the dforce subgraph (mainnet):

- [x] debug subgraph indexing

- [x] verify information correctness

- [x]  incorporate any feedback

Relevant commit starting from f4fbb6fb65068f5dd90c1d16193ded4959fcd3fd (commits prior have been merged)

PR #317, opened 2022-06-14
Refactored to make compound-forks handlers more generic

tnkrxyz:
This PR refactors src/mapping.ts for compound-forks to make the handlers more generic (accepting necessary information as arguments to handler functions instead of from a particular Event object), following the discussion at https://github.com/messari/subgraphs/pull/263#issuecomment-1145430732 .

Two test runs are ongoing: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aurigami and https://thegraph.com/hosted-service/subgraph/tnkrxyz/venus-subgraph (no errors so far).

### DIFF
--- a/subgraphs/compound-forks/src/mapping.ts
+++ b/subgraphs/compound-forks/src/mapping.ts
@@ -169,14 +169,16 @@ export class UpdateMarketData {
 // event.params.cToken:
 // event.params.oldCollateralFactorMantissa:
 // event.params.newCollateralFactorMantissa:
-export function _handleNewCollateralFactor(event: NewCollateralFactor): void {
-  let marketID = event.params.cToken.toHexString();
+export function _handleNewCollateralFactor(
+  marketID: string,
+  newCollateralFactorMantissa: BigInt
+): void {
   let market = Market.load(marketID);
   if (market == null) {
     log.warning("[handleNewCollateralFactor] Market not found: {}", [marketID]);
     return;
   }
-  let collateralFactor = event.params.newCollateralFactorMantissa
+  let collateralFactor = newCollateralFactorMantissa
     .toBigDecimal()
     .div(mantissaFactorBD)
     .times(BIGDECIMAL_HUNDRED);
@@ -200,9 +202,9 @@ export function _handleNewCollateralFactor(event: NewCollateralFactor): void {
 // event.params.newLiquidationIncentiveMantissa
 export function _handleNewLiquidationIncentive(
   protocol: LendingProtocol,
-  event: NewLiquidationIncentive
+  newLiquidationIncentiveMantissa: BigInt
 ): void {
-  let liquidationIncentive = event.params.newLiquidationIncentiveMantissa
+  let liquidationIncentive = newLiquidationIncentiveMantissa
     .toBigDecimal()
     .div(mantissaFactorBD)
     .minus(BIGDECIMAL_ONE)
@@ -231,9 +233,9 @@ export function _handleNewLiquidationIncentive(
 // - newPriceOracle
 export function _handleNewPriceOracle(
   protocol: LendingProtocol,
-  event: NewPriceOracle
+  newPriceOracle: Address
 ): void {
-  protocol._priceOracle = event.params.newPriceOracle.toHexString();
+  protocol._priceOracle = newPriceOracle.toHexString();
   protocol.save();
 }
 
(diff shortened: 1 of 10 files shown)


0xbe1 (commented):
@tnkrxyz I kinda prefer `function _handleAccrueInterest(params: AccrueInterestParams)` over `function _handleAccrueInterest(a, b, c, d, e, f, ...)`

This applies to all common functions. What do you think?

Other than that, LGTM

tnkrxyz:
I agree that is generally a better approach - but I don't like the fact that I have to create a Data class like MarketListData for each handler. What do you think of _handleAccrueInterest({totalBorrows, AccrueInterest})? With this, we will need validation inside each handler for presence and correct type of params.

0xbe1:
> @0xbe1 I agree that is generally a better approach - but I don't like the fact that I have to create a Data class like MarketListData for each handler. What do you think of _handleAccrueInterest({totalBorrows, AccrueInterest})? With this, we will need validation inside each handler for presence and correct type of params.

hmm that's also a good point 🤷 assemblyscript is very verbose.

LGTM

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -169,14 +169,16 @@ export class UpdateMarketData {
 // event.params.cToken:
 // event.params.oldCollateralFactorMantissa:
 // event.params.newCollateralFactorMantissa:
-export function _handleNewCollateralFactor(event: NewCollateralFactor): void {
```
0xbe1, line 172: remove `NewCollateralFactor` from import

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -200,9 +202,9 @@ export function _handleNewCollateralFactor(event: NewCollateralFactor): void {
 // event.params.newLiquidationIncentiveMantissa
 export function _handleNewLiquidationIncentive(
   protocol: LendingProtocol,
-  event: NewLiquidationIncentive
```
0xbe1, line 203: remove `NewLiquidationIncentive` from imports

0xbe1 (commented):
basically remove all unused imports

dmelotik:
@tnkrxyz nice job! I think the method headers should be updated in `src/mapping.ts`.

Also looks like their was a syncing issue with the aurigami test.

Other than that lgtm

tnkrxyz:
> > @0xbe1 I agree that is generally a better approach - but I don't like the fact that I have to create a Data class like MarketListData for each handler. What do you think of _handleAccrueInterest({})? With this, we will need validation inside each handler for presence and correct type of params.
> 
> hmm that's also a good point 🤷 assemblyscript is very verbose.

@0xbe1, I have tried to pass an object as an argument:

```
export function _handleAccrueInterest(
  updateMarketData: UpdateMarketData,
  accrueInterestParams: {
    comptrollerAddr: Address;
    interestAccumulated: BigInt;
    totalBorrows: BigInt;
  },
  event: ethereum.Event
): void {...}
```
But `graph build` complains that `ERROR TS1109: Expression expected`, seem to me due to assemblyScript more limited than typescript. So what I have in mind doesn't work. Do you think it is worthwhile to create a Data class (like MarketListData) for each handler function to contain the params?

@dmelotik , Yes, I will be updating the header to remove unused imports. aurigami subgraph failing is weird, because it doesn't seem to relate to my changes at all: `failed to process trigger: block #61007515 (0x8a70…09de), transaction 276c3690a9721a11a01eba68381f647e987bf56313c21d9dd97f03419c3e1164: Found no transaction for event`. I am pushing another deploy without my changes to see if it works.

0xbe1:
@tnkrxyz 

aurigami - it seems the error relates to Aurora issue with The Graph, do you mind reaching out The Graph team at #subgraph-development channel on graphprotocol server?

if it is worthwhile - yeah if you want it to work it has to be in the Data class way. I am still not sure if it is worthwhile, but I think it is good to go without these Data class

tnkrxyz:
@0xbe1 Sure, I will reach out to the graph team (not yet on their discord server)

Let me confirm if I understand you correctly - I will keep params of the handler functions as they are now in my commit and no need to convert the params to data class. Am I right?

0xbe1:
@tnkrxyz yeah you're right

tnkrxyz:
This PR is now ready to be merged. The issue with aurigami subgraph remains (I am getting access to the thegraph discord server & report the issue there), but it has nothing to do with our code.

this-username-is-taken:
> This PR is now ready to be merged. The issue with aurigami subgraph remains (I am getting access to the thegraph discord server & report the issue there), but it has nothing to do with our code.

Would you be able to open an issue to track the aurigami issue? Otherwise great job on this! And appreciate @dmelotik @0xbe1 for looking this over!

tnkrxyz:
Tracked with issue #343

PR #318, opened 2022-06-14
Yeti finance

amritkumarj:
Deployed subgraph : https://thegraph.com/hosted-service/subgraph/amritkumarj/yeti-finance

Yeti finance


subgraphs/yeti-finance/README.md
```diff
@@ -0,0 +1 @@
+Yeti Finance
```
this-username-is-taken, line 1: You should complete the README like the other subgraphs
e.g. https://github.com/messari/subgraphs/blob/master/subgraphs/liquity/README.md

this-username-is-taken:
Implementation is incomplete but merging for future reference

PR #321, opened 2022-06-14
Add ERC20 Subgraph

harsh9200:


subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  " Transaction hash of the transaction that emitted this event "
+  hash: String!
+
+  " Event log index. For transactions that don't emit event, create arbitrary index starting from 0 "
+  logIndex: Int!
+
+  " The token this transaction belongs to "
+  token: Token!
+
+  nonce: Int!
```
this-username-is-taken, line 185: add comment `" Nonce of the transaction that emitted this event "`

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  " Name of the token, mirrored from the smart contract "
+  name: String!
+
+  " Symbol of the token, mirrored from the smart contract "
+  symbol: String!
+
+  " The number of decimal places this token uses, default to 18 "
+  decimals: Int!
+
+  " If token transfers are paused, derived from paused event "
```
this-username-is-taken, line 21: `pause` doesn't seem to be a standard ERC20 event. What are some tokens that have `pause` event?

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  isPaused: Boolean
+
+  " Token Holders Balance "
+  holdersBalance: [AccountBalance!]! @derivedFrom(field: "token")
+
+  " Number of token holders "
+  holderCount: BigInt!
+
+  " Total number of token transfer transactions "
```
this-username-is-taken, line 30: probably want to be specific here about whether we are talking about events or transactions. If 1 tx contains 10 token transfers, I think we should consider it as 10 transfers (i.e. transferCount += 10)

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  transferCount: BigInt!
+
+  " Total number of token mint transactions "
+  mintCount: BigInt!
+
+  " Total number of token burn transactions "
+  burnCount: BigInt!
+
+  " Total token supply "
+  totalSupply: BigDecimal!
```
this-username-is-taken, line 40: is `currentSupply` or `liquidSupply` something we can track as well? i.e. (sum of all mint amounts - sum of all burns)

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  holdersBalance: [AccountBalance!]! @derivedFrom(field: "token")
+
+  " Number of token holders "
+  holderCount: BigInt!
+
+  " Total number of token transfer transactions "
+  transferCount: BigInt!
+
+  " Total number of token mint transactions "
+  mintCount: BigInt!
```
this-username-is-taken, line 34: should we track mint/burn events as well?

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  id: ID!
+
+  " Token this snapshot is associated with "
+  token: Token!
+
+  " # of accounts holding this token "
+  dailyHolderCount: Int!
+
+  " # of unique daily active users "
+  dailyActiveUsers: Int!
```
this-username-is-taken, line 78: should we rename users to holders since we are using holders elsewhere?

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  " Total number of token transfer transactions "
+  transferCount: BigInt!
+
+  " Total number of token mint transactions "
+  mintCount: BigInt!
+
+  " Total number of token burn transactions "
+  burnCount: BigInt!
+
+  " Total token supply "
```
this-username-is-taken, line 39: Should we have a `liquidSupply` or `currentSupply` to track the supply balance? This will be useful in the daily snapshots

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+  hourlyHolderCount: Int!
+
+  " # of unique hourly active users "
+  hourlyActiveUsers: Int!
+
+  " # of cumulative unique users "
+  cumulativeUniqueUsers: Int!
+
+  " Total number of transactions occurred in an hour. Transactions include all entities that implement the Event interface. "
+  hourlyTransactionCount: Int!
```
this-username-is-taken, line 122: since tokens have very basic events (and don't have calls to handle), maybe we should just rename all transactions to events?

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,293 @@
...
+##################################
+######## Account MetaData ########
+##################################
+
+# An account is a unique Ethereum address
+type Account @entity {
+  " Address of the account "
+  id: ID!
+
+  " Token Transfers by this account "
```
this-username-is-taken: you mean `to` this account? doesn't `by` and `from` mean the same thing?

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,292 @@
...
+
+  ##### Snapshots #####
+
+  " Daily usage metrics for this token "
+  dailyUsageMetrics: [UsageMetricsDailySnapshot!]! @derivedFrom(field: "token")
+
+  " Hourly usage metrics for this token "
+  hourlyUsageMetrics: [UsageMetricsHourlySnapshot!]! @derivedFrom(field: "token")
+
+  dailyTokenSnapshot: [TokenDailySnapshot!]! @derivedFrom(field: "token")
```
this-username-is-taken, line 59: since there is one field in here, how about we combine it with usage metrics and call them `tokenSnapshot`? e.g. `tokenDailySnapshot` and `tokenHourlySnapshot`

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,292 @@
...
+
+  " Token balances that this account holds "
+  balances: [AccountBalance!]! @derivedFrom(field: "account")
+
+  " Token balances snapshot that this account holds "
+  balancesSnapshot: [AccountBalanceDailySnapshot!]! @derivedFrom(field: "account")
+}
+
+type AccountBalance @entity {
+  " { Address Of the Account } - {Address of the Token}"
```
this-username-is-taken, line 251: nit: spaces between curly braces and the first/last word

subgraphs/erc20/schema.graphql
harsh9200: yeah that make sense

subgraphs/erc20/schema.graphql
harsh9200: yeah, it's not a standard ERC20 event, I have removed this from the schema

subgraphs/erc20/schema.graphql
harsh9200: yeah I think so.

subgraphs/erc20/schema.graphql
harsh9200: what is the difference between the `totalSupply` and the `currentSupply`?

subgraphs/erc20/schema.graphql
harsh9200: yeah, we can do that

subgraphs/erc20/schema.graphql
this-username-is-taken: I'd imagine some of the token that got minted have never been moved (e.g. locked). So would be considered illiquid supply. But I guess those would be difficult to track

subgraphs/erc20/schema.graphql
```diff
@@ -0,0 +1,269 @@
...
+  " Timestamp of this event "
+  timestamp: BigInt!
+}
+
+##################################
+######## Account MetaData ########
+##################################
+
+# An account is a unique Ethereum address
+type Account @entity {
```
this-username-is-taken, line 210: I wonder if instead of representing account holding as `Account` and `AccountBalance`, we should make an account to be token-specific (i.e. an `Account` for a specific token only shows the balances of a particular token).

I have a feeling that having a ton of accounts x a ton of tokens might be impossible to index.

this-username-is-taken: I think we can start with a simplified version, see how long it takes to index that, and then iterate based on the findings

PR #322, opened 2022-06-15
Updated deployment scripts and added documentation.

steegecs:


.github/workflows/deployment.yml
```diff
@@ -19,4 +19,4 @@ jobs:
           HOSTED_SERVICE_ACCESS_TOKEN: ${{ secrets.HOSTED_SERVICE_ACCESS_TOKEN }}
           CHANGED_FILES: "${{ steps.files.outputs.all }}"
           ABSOLUTE_PATH: ${{ github.workspace }}
-          GRAPH_DEPLOYMENT_LOCATION: messari
+          GRAPH_DEPLOYMENT_LOCATION: steegecs
```
this-username-is-taken, line 22: change back?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,9 @@
+name: Deployment
```
this-username-is-taken, line 1: remove?

this-username-is-taken:
Great work!

.github/workflows/deployment.yml
steegecs: Good catch. Changed back.

.github/workflows/test.yaml
steegecs: Removed

Issue #242, opened 2022-05-31
Compound: rates don't change in snapshots


menaitm:
The chart on the Messari dashboard shows the rates are fixed throughout historical snapshots. But they should be changing over time ([see rates on Compound](https://compound.finance/markets/ETH)).

[https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/compound-ethereum&tab=pool&poolId=0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/compound-ethereum&tab=pool&poolId=0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5)

![image](https://user-images.githubusercontent.com/106617039/171232085-d78c27ff-5b09-4817-b5f4-edce9415f829.png)

Problem occurs for both the `hourly` and `daily` snapshots.

Replicate with the query:
```
{  			
  markets(
      first:1
      orderBy:totalValueLockedUSD
      orderDirection:desc
  ) {
      name
      dailySnapshots (
          first: 5
          orderDirection: desc
          orderBy: timestamp
      ) {
          blockNumber
          timestamp

          # Below should change over time but do not
          rates(orderBy:id) {
            id
            rate
          }
      }    
  }
}
```

this-username-is-taken:
We are in code freeze for Graph Hack (https://thegraph.com/graph-hack/2022). Will release the new versions early next week.

menaitm:
@this-username-is-taken thank you.

The Compound reward rates are also affected (show 0% for everything). Is 1.5.5 expected to patch this too?
eg. https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fname%2Fmessari%2Fcompound-ethereum&tab=pool&poolId=0x4ddc2d193948926d02f9b1fe9e1daa0718270ed5&view=marketHourlySnapshots-rewardAPR

menaitm:
I'll track reward rates in your separate issue here: https://github.com/messari/subgraphs/issues/288

Closing this issue once `1.5.5` is released

dmelotik:
[...]

PR #323, opened 2022-06-15
Fix Compound Sync Failure

dmelotik:
Fixing error created from https://github.com/messari/subgraphs/issues/242

Not able to fork the subgraph since the state was not saved at failure.

it is deployed to my studio and the status can be seen here https://okgraph.xyz/?q=Qmbdr7ZXRuaBia4BWbKyHJLBQvLLvU3LHZZTKcrBRXcCk6


this-username-is-taken:
Great!

PR #331, opened 2022-06-16
Updated liquidate event

this-username-is-taken:


0xbe1 (commented):
lgtm

schema-lending.graphql
```diff
@@ -994,6 +994,9 @@ type Liquidate implements Event @entity {
   " Address that carried out the liquidation "
   from: String!
 
+  " Address that got liquidated "
+  liquidated: String!
```
0xbe1, line 998: does `liquidatee` sound less ambiguous? `liquidated` could also mean the liquidated asset aka collateral

this-username-is-taken: I don't think `liquidatee` is a real word :)

this-username-is-taken: Also, `liquidated` is a bit ambiguous

0xbe1: or just `liquidatedAccount` or even `victim` 😆

PR #332, opened 2022-06-16
Yearn v2: Schema update 1.3.0

harsh9200:
- Hard-coded problematic stable coins (LUSD Stable) values to One USD. 
- Fixed zero protocol revenue bug.
- Updated schema to 1.3.0


this-username-is-taken:
LGTM!

PR #338, opened 2022-06-17
Extended lending schema to include account/position data

this-username-is-taken:
To compare diff:

https://raw.githubusercontent.com/messari/subgraphs/master/schema-lending.graphql
https://raw.githubusercontent.com/messari/subgraphs/b7f5412c347eef3928d0d6f9a4b7fcf95c7489b3/schema-lending-extended.graphql

paste into:
https://www.diffchecker.com/

## Use Cases

1. View the current positions of an account (e.g. how much money do I have in Aave?)
2. View all historical positions of an account (e.g. what does my lending/borrowing behavior look like?)
3. View all the open positions in a market

## Open Questions

- Should we calculate USD prices in positions? These can be computed offchain more easily


schema-lending-extended.graphql
```diff
@@ -0,0 +1,1146 @@
...
+
+enum InterestRateSide {
+  " Interest rate accrued by lenders "
+  LENDER
+
+  " Interest rate paid by borrowers "
+  BORROWER
+}
+
+enum PositionSide {
```
dmelotik, line 112: could you add this to the `_reference_` subgraph?

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1146 @@
...
+
+  " Revenue claimed by suppliers to the protocol. LPs on DEXs (e.g. 0.25% of the swap fee in Sushiswap). Depositors on Lending Protocols. NFT sellers on OpenSea. "
+  cumulativeSupplySideRevenueUSD: BigDecimal!
+
+  " Gross revenue for the protocol (revenue claimed by protocol). Examples: AMM protocol fee (Sushi’s 0.05%). OpenSea 10% sell fee. "
+  cumulativeProtocolSideRevenueUSD: BigDecimal!
+
+  " All revenue generated by the protocol. e.g. 0.30% of swap fee in Sushiswap, all yield generated by Yearn. "
+  cumulativeTotalRevenueUSD: BigDecimal!
+
```
dmelotik, line 192: missing `totalPoolCount`

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1146 @@
...
+  " Sum of all historical borrows/mints in USD (i.e. total loan origination). "
+  cumulativeBorrowUSD: BigDecimal!
+
+  " Total assets liquidated on a given day, in USD. "
+  dailyLiquidateUSD: BigDecimal!
+
+  " Sum of all historical liquidations in USD "
+  cumulativeLiquidateUSD: BigDecimal!
+}
```
dmelotik, line 441: missing `withdrawUSD` and `RepayUSD`

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1146 @@
...
+
+type Account @entity {
+  " { Account address } "
+  id: ID!
+  
+  " Number of open positions this account has "
+  positionCount: Int!
+
+  " All open positions that belong to this account "
+  positions: [Position!]! @derivedFrom(field: "account")
```
dmelotik, line 775: will we delete position entities once they are closed? I don't know how else we would remove from this derived array then

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1146 @@
...
+type PositionSnapshot @entity {
+  " { Position ID }-{ # of days since Unix epoch time } "
+  id: ID!
+
+  " Position of this snapshot "
+  position: Position!
+
+  " Token balance in this position, in native amounts "
+  balance: BigInt!
+
```
dmelotik, line 859: thoughts on `amountUSD` here?

dmelotik (commented):
Aside from missing the latest commits to the regular lending schema it LGTM

schema-lending-extended.graphql
this-username-is-taken: Good point, will remove

this-username-is-taken:
> Aside from missing the latest commits to the regular lending schema it LGTM

Thanks for looking it over!

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
+  ##### Snapshots #####
+
+  " Daily usage metrics for this protocol "
+  dailyUsageMetrics: [UsageMetricsDailySnapshot!]! @derivedFrom(field: "protocol")
+
+  " Hourly usage metrics for this protocol "
+  hourlyUsageMetrics: [UsageMetricsHourlySnapshot!]! @derivedFrom(field: "protocol")
+
+  " Daily financial metrics for this protocol "
+  financialMetrics: [FinancialsDailySnapshot!]! @derivedFrom(field: "protocol")
```
KremerMichael, line 205: Should we make this `dailyFinancialMetrics` for consistency with `dailyUsage Metrics` ?

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
```
KremerMichael, line 292: Same Daily comment

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
+  " Creation block number "
+  createdBlockNumber: BigInt!
+
+  ##### Account/Position Data #####
+
+  " All open positions in this market "
+  positions: [Positions!]! @derivedFrom(field: "market")
+
+  " Number of open positions in this market "
+  positionCount: Int!
```
KremerMichael, line 566: Maybe overkill: It may be worth it to break this up into deposit & borrow position counts

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
+  market: Market!
+
+  " The hash of the transaction that opened this position "
+  hash: String!
+
+  " Side of the position (either lender or borrower) "
+  side: PositionSide!
+
+  " Whether this position has been closed or not "
+  closed: Boolean!
```
KremerMichael, line 858: Can we have closed timestamp/blocknumber so we can measure the duration of positions.

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
+
+type Account @entity {
+  " { Account address } "
+  id: ID!
+  
+  " Number of open positions this account has "
+  positionCount: Int!
+
+  " All open positions that belong to this account "
+  positions: [Position!]!
```
KremerMichael, line 802: Just sanity checking that old & closed positions will be in here. If I were to ping all positons/events for an Account, I'd have a full view of their history?

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1173 @@
...
+  side: PositionSide!
+
+  " Whether this position has been closed or not "
+  closed: Boolean!
+
+  " The token in the position "
+  token: Token!
+
+  " Token balance in this position, in native amounts "
+  balance: BigInt!
```
KremerMichael, line 864: How are we handling this changing over time?

KremerMichael: It may make the schema overly circular/bloated, but should we hold the events of a position with this entity?

KremerMichael (commented):
Can't wait to dig into all of this!

schema-lending-extended.graphql
this-username-is-taken: hmm, good point. There are some historical reasons to this. We only had daily values before, so they were named `usageMetrics` and `financialMetrics`. Later people wanted hourly data for usage, so I made `usageMetrics` into daily and hourly. But no one asked for hourly financial metrics so I left as is

schema-lending-extended.graphql
this-username-is-taken: What would be some use cases for this? There are some performance reasons for this design but would want to weight the trade-off

schema-lending-extended.graphql
this-username-is-taken: Good point!

schema-lending-extended.graphql
this-username-is-taken: `PositionSnapshot` has `balance` change over time

this-username-is-taken: Yeah, I've thought about that. What are the use cases you have in mind?

this-username-is-taken:
Seems like there is a demand for even more nuanced data. I think the easier thing to do is to include everything we want, see how the indexing time is, and then do ablation studies to find the middle ground if indexing time is too long

schema-lending-extended.graphql
KremerMichael: It's another way to show long/short bias. Instead of long/short bias in USD, now its long/short Bias in terms of users.

KremerMichael:
> Seems like there is a demand for even more nuanced data. I think the easier thing to do is to include everything we want, see how the indexing time is, and then do ablation studies to find the middle ground if indexing time is too long

It may not be in scope but the next big thing would be flashloans.

I will stare at the extended schema more but likely won't have a ton more of in depth feedback until it's deployed somewhere and I can start to mess with it

schema-lending-extended.graphql
KremerMichael: The first thing that comes to mind would be aggressively tracking high importance positions under stress. Lot's of people rn are watching the celsius positions continue to top up on collateral, this may be the easiest way to go about that. It is a niche use case though.

this-username-is-taken:
No problem. This will be an iterative process

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1239 @@
...
+
+  " Number of times this account has been liquidated "
+  liquidationCount: Int!
+
+  " All liquidation events where this account got liquidated "
+  liquidations: [Liquidate!]! @derivedFrom(field: "liquidatee")
+}
+
+type Position @entity {
+  " { Account address }-{ Market address }-{ Counter } "
```
0xbe1, line 878: is the counter increment only?

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1239 @@
...
+  id: ID!
+
+  " Account that owns this position "
+  account: Account!
+
+  " The market in which this position was opened "
+  market: Market!
+
+  " The hash of the transaction that opened this position "
+  hash: String!
```
0xbe1, line 888: say the user deposits 1ETH, then 3ETH, then 5ETH to compound - will that create 3 positions or just one?

schema-lending-extended.graphql
```diff
@@ -0,0 +1,1239 @@
...
+
+  " Side of the position (either lender or borrower) "
+  side: PositionSide!
+
+  " The token in the position "
+  token: Token!
```
0xbe1, line 894: is it theoretically derivable from `market`? or not

schema-lending-extended.graphql
0xbe1: how about `hashOpened`? to be consistent with `blockNumberOpened` etc

0xbe1 (commented):
I am thinking about a dedicated use case - "what addresses are underwater"?

To calculate this, we need to know the positions of all its collaterals and its borrowing assets. I am wondering which part in the schema keeps track of its collateral?

schema-lending-extended.graphql
this-username-is-taken: Yes

this-username-is-taken:
> I am thinking about a dedicated use case - "what addresses are underwater"?
> 
> To calculate this, we need to know the positions of all its collaterals and its borrowing assets. I am wondering which part in the schema keeps track of its collateral?

All positions with `PositionSide` being lender will give you the collaterals

0xbe1:
@this-username-is-taken Let's say in compound, I:

1. deposit 10 ETH and get 500 cETH
2. use 400 cETH as collateral to borrow 1000 USDC

So I have positions of:

1. 10 ETH (lender)
2. 400 cETH (lender) -- collateral
3. 1000 USDC (borrower)

is that correct? 🤔

this-username-is-taken:
> @this-username-is-taken Let's say in compound, I:
> 
> 1. deposit 10 ETH and get 500 cETH
> 2. use 400 cETH as collateral to borrow 1000 USDC
> 
> So I have positions of:
> 
> 1. 10 ETH (lender)
> 2. 400 cETH (lender) -- collateral
> 3. 1000 USDC (borrower)
> 
> is that correct? 🤔

I've updated the docs. Let me know if it answers your question. With regard to your example, why do we care about the cETH?

schema-lending-extended.graphql
this-username-is-taken: Good point. Added

schema-lending-extended.graphql
this-username-is-taken: Good point. `Market` provides much more complete data

PR #339, opened 2022-06-17
Curve hourly snapshot

sevenshi:


this-username-is-taken:
Is this ready or still a work in progress?

subgraphs/curve-finance/src/services/pools.ts
```diff
@@ -70,7 +70,8 @@ export function createNewPool(
   pool.gauge = gaugeAddress.toHexString();
   const inputTokens = getPoolCoins(pool);
   pool.coins = inputTokens;
-  pool.inputTokens = inputTokens.sort();
+  pool.inputTokens = inputTokens;
```
this-username-is-taken, line 73: Is this the only change? If so, can you revert all the other changes (including the commented line below)?

this-username-is-taken, line 73: Is this the only change? If so, can you revert all the other changes (including the commented line below)?

sevenshi: yes  ok

PR #356, opened 2022-06-19
curve hourly snapshot update

sevenshi:


this-username-is-taken:
see https://github.com/messari/subgraphs/pull/386

PR #381, opened 2022-06-21
Support dex schema v1.3.0 on dashboard

jihok:
most of these changes are from prettier while I was logging around to figure things out. Any removed imports were unused

### DIFF
--- a/dashboard/src/common/DashboardVersion.tsx
+++ b/dashboard/src/common/DashboardVersion.tsx
@@ -12,5 +12,5 @@ const DashboardTag = styled("div")`
 `;
 
 export const DashboardVersion = () => {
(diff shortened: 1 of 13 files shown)


this-username-is-taken:
@jihok  Can you also increment the dashboard version?

jihok:
Any ground rules on how we handle semver here? A minor bump makes sense to me, but seems we've only updated as patches so far so I bumped to 1.1.5

this-username-is-taken:
yeah, the versioning has been pretty adhoc so far. It's only for the end-user (subgraph developers) to know which version has been deployed as sometimes people might make feature request for something that's under development, or have a PR out but haven't been deployed.

I think as long as we make it different for every PR it'll be fine

PR #387, opened 2022-06-22
finish supporting schema 1.3.0

jihok:
support 1.3.0 for yield and lending. Lmk if there's anything else that needs to be updated

### DIFF
--- a/dashboard/src/common/DashboardVersion.tsx
+++ b/dashboard/src/common/DashboardVersion.tsx
@@ -12,5 +12,5 @@ const DashboardTag = styled("div")`
 `;
 
 export const DashboardVersion = () => {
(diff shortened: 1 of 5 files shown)


this-username-is-taken:
Great! Thanks.

Also `totalPoolCount` as shown here: https://github.com/messari/subgraphs/pull/330.

Not sure how you are diffing the schema versions but the 1.3.0 upgrade PR is missing lots of changes as we've iterated on it a couple times. For a complete set of changes, probably good to diff between the commit before the 1.3.0 PR vs the current graphql file, or you can refer to this doc, which should contain a complete list of all changes: https://github.com/messari/subgraphs/blob/master/CHANGES.md#schema-update-130

jihok:
Ah I was diffing from only the main PR. I'll get in the rest 👍

Issue #258, opened 2022-06-01
Moonwell rewards seem too high


dmelotik:
`rewardTokenEmissions` seem to be too high. Emissions are supposed to be normalized to a day, looks like it is normalized to a year.

![image](https://user-images.githubusercontent.com/60583747/171315800-2fedfc33-bb8f-4534-9839-944c7893527d.png)

See visualizer graph [here](https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fname%2Fmessari%2Fmoonwell-moonriver&tab=pool&poolId=0x36918b66f9a3ec7a59d0007d8458db17bdffbf21&view=marketDailySnapshots-rewardTokenEmissionsUSD%2520%5B1%5D)

dmelotik:
Seems to look right and is fixed in the pending version: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmXYwnpaLAo5d5XSHHnuguFY13QsXoviEKeM2hnjiQKzL3&tab=poolOverview

PR #413, opened 2022-06-23
Fix Moonwell Rewards

dmelotik:
Fixing #258 

Test subgraph: https://thegraph.com/hosted-service/subgraph/dmelotik/moonwell?version=pending


this-username-is-taken:
NIce catch!

PR #415, opened 2022-06-23
MakerDAO: Fix warnings

### LINKED ISSUE #244: Maker DAO build and deployment errors
Screen Shot 2022-05-31 at 1 22 11 PM

These errors show up when I built the subgraph. These are most likely due to these fields not being marked as required in the schema, and they are used in the program without either checking if null or asserting not null with '!'. I tried to fix these issue for deployment by adding the exclamation, however, the deployment still failed when I tried to deploy to Messari Hosted service.

harsh9200:
Fixes #244


this-username-is-taken:
Thanks!

PR #416, opened 2022-06-23
FIx bugs in Schema v2.3.0

nemani:
Bug List:
- Remove double Account Entity
- Use `Position` instead of `Positions` in array
- Use `Account` instead of `String` in Event and its Implementations


this-username-is-taken:
2.3.0 is fine as it wouldn't be actually usable without these changes 😅

PR #417, opened 2022-06-23
Extend Abracadabra Subgraph to use v2.0.1

nemani:
This PR attempts to extend Abracadabra Subgraph to support schema version > 2.0.0
Currently: 2.0.1


subgraphs/abracadabra/src/common/utils/arrays.ts
```diff
@@ -1,3 +1,33 @@
+export function removeFromArrayAtIndex<T>(x: T[], index: i32): T[] {
```
this-username-is-taken, line 1: This is nice! Would you be able to add these to the `_reference_` subgraph?

subgraphs/abracadabra/src/mapping.ts
```diff
@@ -193,53 +206,64 @@ export function handleLiquidation(event: LogRepay): void {
...
   liquidateEvent.amount = collateralAmount;
   liquidateEvent.amountUSD = collateralAmountUSD;
   liquidateEvent.profitUSD = collateralAmountUSD.minus(mimAmountUSD);
-  liquidateEvent.liquidatee = event.params.to.toHexString();
+  liquidateEvent.save();
+
+  let liqudidatedAccount = getOrCreateAccount(liquidateEvent.liquidatee!, event);
+  liqudidatedAccount.liquidateCount = liqudidatedAccount.liquidateCount + 1;
+  liqudidatedAccount.save();
+  addAccountToProtocol("LIQUIDATEE", liqudidatedAccount, event);
```
this-username-is-taken, line 225: Probably better to make these named constants

this-username-is-taken:
Everything looks good other than above. After you are done with the changes, would you be able to rename this `abracadabra-ext` while maintaining the 1.3.0 version as `abracadabra`? We want to maintain both versions going forward

subgraphs/abracadabra/src/common/utils/arrays.ts
nemani: Sure, I will create another PR for this!

subgraphs/abracadabra/src/mapping.ts
nemani: done, in a new commit.

subgraphs/abracadabra/src/common/utils/arrays.ts
nemani: https://github.com/messari/subgraphs/pull/493

PR #420, opened 2022-06-23
Schema Proposal: Add Side to Position ID, use BigInt for nonce;

nemani:
- Having the type of the postion in the id is beneficial as it allows for quickly finding the latest open position of each type example code:
```
// Finds the open lending position or opens a new one 
export function getOrCreateLendingOpenPosition(marketId: string, accountId: string, event: ethereum.Event): Position {
  let positionIdPrefix = `${accountId}-${marketId}-LENDER`;

  let account = getOrCreateAccount(accountId, event);
  for (let curr = 0; curr < account.openPositionCount; curr += 1) {
    let op = account.openPositions.at(curr);
    if (op.startsWith(positionIdPrefix)) {
      return Position.load(op)!;
    }
  }

  let count = 0;
  for (let curr = 0; curr < account.closedPositionCount; curr += 1) {
    let cp = account.closedPositions.at(curr);
    if (cp.startsWith(positionIdPrefix)) {
      count += 1;
    }
  }

  let newPositionId = `${accountId}-${marketId}-LENDER-${count}`;
```

- event.transaction.nonce is BigInt so we should use BigInt in nonce for event;

### DIFF
--- a/schema-lending-extended.graphql
+++ b/schema-lending-extended.graphql
@@ -931,7 +931,7 @@ type Account @entity {
 }
 
 type Position @entity {
-  " { Account address }-{ Market address }-{ Counter } "
+  " { Account address }-{ Market address }-{ Position Type }-{ Counter } "
   id: ID!
 
   " Account that owns this position "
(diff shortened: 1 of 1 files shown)


nemani:
@this-username-is-taken

this-username-is-taken:
Good change 👍

PR #423, opened 2022-06-23
schema v2.3.0:  `Liquidate` does not satisfy `Event`

nemani:


0xbe1:
@this-username-is-taken this can be closed

PR #426, opened 2022-06-24
Updated extended lending schema to be more fitting

this-username-is-taken:


schema-lending-extended.graphql
```diff
@@ -1164,26 +1126,21 @@ type Borrow implements Event @entity {
...
-  from: Account!
-
   " Block number of this event "
   blockNumber: BigInt!
 
   " Timestamp of this event "
   timestamp: BigInt!
 
-  ##### Lending-Specific #####
+  " Account that borrowed/minted tokens "
```
0xbe1, line 1135: why "minted" in "Borrow" event?

schema-lending-extended.graphql
```diff
@@ -1164,26 +1126,21 @@ type Borrow implements Event @entity {
...
+  account: Account!
 
-  " The market this transaction interacted with "
+  " The market tokens are borrowed/minted from "
```
0xbe1, line 1138: same question

0xbe1 (commented):
lgtm!

schema-lending-extended.graphql
this-username-is-taken: mint is equivalent to borrow in CDPs

PR #427, opened 2022-06-24
Add more aave-v2 networks / fix rev/tvl/borrowusd

dmelotik:
Deliverables:
- Fix TVL calculation
- Fix Revenue calculation
- Reorganize directory structure to make adding forks easier
- added other Aave v2 protocols
- separated Ethereum and Ethereum AMM implementations
- Added Permissioned protocols (ie, ARC / RWA)

Explanations:
- Revenue calculation: new revenue = totalBorrows * difference in liquidity Index from last transaction

Still Broken:
- How to make network configuration when multiple deployments are on the same network?
- Ensure Aave Reward price is calculated correctly


deployment/deploymentConfigurations.json
```diff
@@ -34,13 +34,43 @@
         }
       }
     },
-    "aave-v2": {
+    "aave-v2-forks": {
       "aave-v2": {
+        "mainnet-arc": {
+          "template": "aave.v2.template.yaml",
+          "messari": "messari/aave-v2-ethereum-arc",
```
this-username-is-taken, line 41: I think they should be more appropriately named `aave-v2-arc-ethereum` as technically `aave-v2-arc` can be deployed on different chains

this-username-is-taken: Alternatively we can just have `aave-arc-ethereum` and `aave-rwa-ethereum` as that seems to align with how they market themselves (no v2)

this-username-is-taken:
Great work Dylan! 🚀🚀🚀

Let me check with our research analysts on how they want the naming to be done. I'm going to merge in the PR now so that it's easier for you to build on top of.

PR #429, opened 2022-06-24
Add compound Price calc switch

dmelotik:
Add a switch to toggle whether to calculate all market prices on each AccrueInterest event or not

Maybe let's review, but wait to deploy b/c it will stop progress on all subgraphs currently syncing.

NOTICE: Maybe we want to approach this a different way using price oracle update events. great idea @0xbe1 

Update always test: 
- https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmW5WMTifq2NST1U3YSWKgZms9L4ix6i9oCzpZsHhNkLpp&tab=protocol 
- https://thegraph.com/hosted-service/subgraph/dmelotik/fuse-official?version=pending

Do not update always test: 
- https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmVsmAsnrkun68vJY2fhTE94JgQJiGD1rHMbbyJx2V9pCa&tab=protocol
- https://thegraph.com/hosted-service/subgraph/dmelotik/benqi


0xbe1:
@dmelotik can i get more context about the PR? Why we want to add the switch, what problem does it solve?

dmelotik:
> @dmelotik can i get more context about the PR? Why we want to add the switch, what problem does it solve?

The idea is that protocols that may have markets that that don't get touched very often may have a TVL that is not accurate b/c the price is not getting updated.

this-username-is-taken:
For example, this is here https://github.com/messari/subgraphs/pull/263#issuecomment-1158533494

tnkrxyz:
This seems to be reasonable trade-off for outdated price data.

dmelotik:
@tnkrxyz what dForce networks are having the stale price data issue? I still have to add dForce into this

tnkrxyz:
> @tnkrxyz what dForce networks are having the stale price data issue? I still have to add dForce into this

dForce mainnet.  Between 5/20 and 6/17, there was no UpdateInterest event emitted in the wstETH market, and stETH prices became stale. But there have been a few UpdateInterest events in last week or so (& ETH price more stable), the stale price problem become less severe/disappears.

this-username-is-taken:
Great work @dmelotik! I just merged this in with CI/CD off. Would you be able to manually deploy the subgraphs for the protocols in question?

dmelotik:
deployed dForce (mainnet), aurigami, and fuse (arbitrum) @this-username-is-taken

PR #432, opened 2022-06-25
Compound forks extended

0xbe1:
🫡 Reviewers:

- The current change only works for moonwell, see [deployment](https://okgraph.xyz/?q=0xbe1%2Fmessari-moonwell-moonriver). To make the rest work, I want an LGTM to move on refactoring the rest.
- Main logic are in `addPosition` and `substractPosition` methods of src/mapping.ts


this-username-is-taken:
can you put this in a separate folder? e.g. `compound-forks-ext`. This way we don't override our existing ones (we will maintain both at the same time).

0xbe1:
> can you put this in a separate folder? e.g. `compound-forks-ext`. This way we don't override our existing ones (we will maintain both at the same time). You can just copy paste the code.

Will do after we finish the pr review, otherwise it is impossible for reviewers to review the code.

this-username-is-taken:
ah, good point

this-username-is-taken:
where is this currently deployed?

0xbe1:
> where is this currently deployed?

updated description, check it out :)

subgraphs/compound-forks/protocols/moonwell/src/mapping.ts
```diff
@@ -193,25 +212,97 @@ export function handleNewReserveFactor(event: NewReserveFactor): void {
 export function handleMint(event: Mint): void {
   let minter = event.params.minter;
   let mintAmount = event.params.mintAmount;
-  _handleMint(comptrollerAddr, minter, mintAmount, event);
+  let contract = CToken.bind(event.address);
+  let balanceOfUnderlyingResult = contract.try_balanceOfUnderlying(
+    event.params.minter
+  );
+  let balanceOfUnderlying = BigInt.fromI32(-1);
```
this-username-is-taken, line 219: Should make `BigInt.fromI32(-1)` a named constant

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1640,6 +1870,217 @@ function getMarketDailySnapshotID(marketID: string, timestamp: i32): string {
...
+    position.balance = BIGINT_ZERO;
+    position.depositCount = 0;
+    position.withdrawCount = 0;
+    position.borrowCount = 0;
+    position.repayCount = 0;
+    position.liquidationCount = 0;
+    position.save();
+  }
+  position = position!;
+  if (balance.ge(BIGINT_ZERO)) {
```
this-username-is-taken, line 1925: I think it'll be easier to understand if you test for equality with the named constant

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1043,28 +1221,58 @@ function snapshotUsage(
...
       break;
     case EventType.Repay:
       dailySnapshot.dailyRepayCount += 1;
       break;
     case EventType.Liquidate:
       dailySnapshot.dailyLiquidateCount += 1;
+      if (newDAU) {
+        dailySnapshot.dailyActiveLiquidators += 1;
+      }
+      // TODO: handle liquidatee
```
this-username-is-taken, line 1262: Does this need to be handled as well?

this-username-is-taken (commented):
Looks great! Just the minor issues I pointed out. Then rename this to `compound-forks-ext` and we are good to go.

subgraphs/compound-forks/protocols/moonwell/src/mapping.ts
0xbe1: cool let me name it `REVERTED` :)

PR #437, opened 2022-06-25
Aave v2: Schema 1.3.0 update

harsh9200:
- Refactored Code.
- Added Price Module support.


subgraphs/aave-v2/src/common/initializers.ts
```diff
@@ -0,0 +1,424 @@
...
+  interestRate.type = rateType;
+
+  interestRate.save();
+
+  return interestRate;
+}
+
+export function getOrCreateMarket(
+  event: ethereum.Event,
+  marketId: string
```
dmelotik, line 181: market id should be the outputToken address

subgraphs/aave-v2/src/modules/Liquidate.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  liquidate.to = debtAsset;
+  liquidate.market = market.id;
+  liquidate.from = event.transaction.from.toHexString();
+  liquidate.hash = event.transaction.hash.toHexString();
+  liquidate.logIndex = event.logIndex.toI32();
+  liquidate.protocol = protocol.id;
+  liquidate.asset = collateralAsset;
+  liquidate.amount = amount;
+  liquidate.liquidatee = liquidator;
```
this-username-is-taken, line 38: this should be `liquidator`, you also need `liquidatee` (user)

subgraphs/aave-v2/src/common/initializers.ts
```diff
@@ -0,0 +1,424 @@
...
+    market._vToken = constants.ZERO_ADDRESS;
+    market.createdTimestamp = event.block.timestamp;
+    market.createdBlockNumber = event.block.number;
+
+    protocol.totalPoolCount += 1;
+    protocol.save();
+    market.save();
+  }
+
+  const currentPrice = getUsdPricePerToken(Address.fromString(market.id));
```
this-username-is-taken, line 243: why not use the native oracle?

this-username-is-taken:
Great work! @dmelotik let me know if you have any other questions or thoughts

subgraphs/aave-v2/schema.graphql
```diff
@@ -78,24 +156,140 @@ interface Protocol {
...
+  lendingType: LendingType
+
+  " Risk type of the lending protocol "
+  riskType: RiskType
+
+  " Tokens that can be minted. Only applies to CDP "
+  mintedTokens: [Token!]
+
+  " String of the address of the current price oracle on the protocol "
+  protocolPriceOracle: String!
```
dmelotik, line 235: should this be marked with an `_` or labeled as an add on?

subgraphs/aave-v2/src/common/constants.ts
```diff
@@ -137,22 +118,17 @@ export const MS_PER_YEAR = DAYS_PER_YEAR.times(new BigDecimal(BigInt.fromI32(24
 
 // This is the LendingPoolAddressesProviderRegistry contract address
 export const PROTOCOL_ADDRESS = "0x52D306e36E3B6B02c153d0266ff0f85d18BCD413";
-
 export const USDC_TOKEN_ADDRESS = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
-
 export const PRICE_ORACLE_ADDRESS = "0xa50ba011c48153de246e5192c8f9258a2ba79ca9";
```
dmelotik, line 122: Instead of hardcoding this we could call `getPriceOracle()` when the protocol is created

subgraphs/aave-v2/src/modules/Price.ts
```diff
@@ -0,0 +1,46 @@
+import * as utils from "../common/utils";
+import * as constants from "../common/constants";
+import { getPriceOracle } from "../common/initializers";
+import { Address, BigDecimal, BigInt, log } from "@graphprotocol/graph-ts";
+
+export function getAssetPriceInUSDC(tokenAddress: Address): BigDecimal {
```
dmelotik, line 6: If we get 0 or a negative asset price aave has a fallback oracle that we should implement: https://docs.aave.com/developers/v/2.0/the-core-protocol/price-oracle#getfallbackoracle

subgraphs/aave-v2/src/modules/Price.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (assetPriceInUSDC.equals(constants.BIGDECIMAL_ZERO)){
+    log.info(
+      "[AssetPriceInUSDC] Failed for {}",
+      [tokenAddress.toHexString()]
+    );
+  }
+
+  // Asset price in Eth/USDC priced in Eth = Asset price in in USDC
+  // return price per asset in USDC
+  return assetPriceInUSDC.truncate(3);
```
dmelotik, line 36: why truncate here?

subgraphs/aave-v2/src/mappings/lendingPoolConfigurator.ts
```diff
@@ -31,24 +24,29 @@ export function handleReserveInitialized(event: ReserveInitialized): void {
...
 
 export function handleCollateralConfigurationChanged(event: CollateralConfigurationChanged): void {
   // Adjust market LTV, liquidation, and collateral data when a reserve's collateral configuration has changed
-  const marketAddr = event.params.asset.toHexString();
-  log.info("MarketAddr in lendingPoolConfigurator.ts handleCollateralConfigurationChanged" + marketAddr, []);
-  const market = initMarket(event.block.number, event.block.timestamp, marketAddr) as Market;
+  const marketAddress = event.params.asset.toHexString();
+  const market = getOrCreateMarket(event, marketAddress);
+
   market.maximumLTV = new BigDecimal(event.params.ltv);
```
dmelotik, line 47: decimal place is messed up for this. seeing 8000%

subgraphs/aave-v2/src/modules/Liquidate.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  liquidate.to = debtAsset;
+  liquidate.market = market.id;
+  liquidate.from = liquidator;
+  liquidate.hash = event.transaction.hash.toHexString();
+  liquidate.logIndex = event.logIndex.toI32();
+  liquidate.protocol = protocol.id;
+  liquidate.asset = collateralAsset;
+  liquidate.amount = amount;
+  liquidate.liquidatee = event.transaction.from.toHexString();
```
dmelotik, line 38: from my understanding the liquidatee is this address: 
image
https://etherscan.io/tx/0xf712b360811a8e66c78437417f514643c0123e58db6d310d2a16000ccf057306#eventlog

dmelotik (commented):
Other than comments LGTM.

@this-username-is-taken let us know if I should fix these comments in my aave PRs or not. Does not matter to me, but I figured I would highlight them while I was looking over the code.

subgraphs/aave-v2/src/common/constants.ts
harsh9200: yeah, we are calling this [contract function](https://github.com/messari/subgraphs/pull/437/files#diff-5f5d7397e42bc85f6ccad939668677e03a5c641f711fd574d5a61f62c477f272R104-R106) when we initialize the context, this constant is just the default value if the RPC call fails.

subgraphs/aave-v2/src/modules/Revenue.ts
```diff
@@ -0,0 +1,178 @@
...
+  const staAmountUSD = market.inputTokenPriceUSD.times(
+    market.totalStableValueLocked
+      .toBigDecimal()
+      .div(constants.BIGINT_TEN.pow(token.decimals as u8).toBigDecimal())
+  );
+  const staFees = staAmountUSD.times(stableBorrowRate.rate);
+
+  // Add these values together, save to market and add protocol total
+  let oldCumulativeTotalRevenueUSD = market.cumulativeTotalRevenueUSD;
+  market.cumulativeTotalRevenueUSD = staFees.plus(varFees).truncate(3);
```
spitko, line 107: This is incorrect, you cannot calculate total fee amounts by multiplying the current interest rate with current balance. For variable debt you could get this from the variable debt index, but for stable debt there is no way to get this data as far as I know.

subgraphs/aave-v2/src/modules/Revenue.ts
```diff
@@ -0,0 +1,178 @@
...
+  protocol.cumulativeProtocolSideRevenueUSD = protoMinusMarketProtoRevenue.plus(
+    market.cumulativeProtocolSideRevenueUSD
+  );
+
+  let oldCumulativeSupplySideRevenueUSD = market.cumulativeSupplySideRevenueUSD;
+  market.cumulativeSupplySideRevenueUSD = market.cumulativeTotalRevenueUSD
+    .times(
+      constants.BIGDECIMAL_ONE.minus(
+        bigIntToBigDecimal(market.reserveFactor, 4)
+      )
```
spitko, line 143: Doesn't work since reserve factor can be changed.

subgraphs/aave-v2/src/modules/Revenue.ts
spitko: This also breaks the rest of the revenue (snapshot) fields because they're based on the assumption that this field is cumulative.

spitko (commented):
Revenue needs to be rethought, maybe look into the official subgraph and see how they handle it.

subgraphs/aave-v2/src/modules/Revenue.ts
this-username-is-taken: I'm gong to resolve these for now but @dmelotik feel free to refer back to these comments when you work on fixing these issues

PR #439, opened 2022-06-25
Extending DEX-AMM to add account/position data

this-username-is-taken:


schema-dex-amm-extended.graphql
```diff
@@ -611,62 +650,182 @@ type LiquidityPoolHourlySnapshot @entity {
...
+  outputTokenBalance: BigInt
+
+  " Total reward token accumulated under this position, in native amounts "
+  cumulativeRewardTokenAmounts: [BigInt!]
+
+  " Number of deposits related to this position "
+  depositCount: Int!
+
+  " All deposit events of this position "
+  deposits: [Deposit!]! @derivedFrom(field: "position")
```
nemani, line 739: There is no field `position` in `Withdraw` and `Deposit`. Needs to be updated.

schema-dex-amm-extended.graphql
```diff
@@ -167,9 +167,21 @@ interface Protocol {
   " Number of cumulative unique users "
   cumulativeUniqueUsers: Int!
 
+  " Number of cumulative liquidity providers "
```
nemani, line 170: These 4 fields should also be in `DexAmmProtocol` which implements interface `Protocol`

schema-dex-amm-extended.graphql
this-username-is-taken: updated

schema-dex-amm-extended.graphql
this-username-is-taken: updated

PR #441, opened 2022-06-26
compact class definition

0xbe1:


dmelotik:
very nice, this will trigger a deployment to all of the compound forks though..

0xbe1:
> very nice, this will trigger a deployment to all of the compound forks though..

oh no is there a way to avoid?

this-username-is-taken:
I disabled Actions when merging so we are all good

PR #445, opened 2022-06-27
ENS Governance++

mmnavarr:
# Title
Tidy up work for ENS Governance subgraph

## Description
- Changed address fields to `String` value for now to be consistent
- Minor updates to comments & GQL schema
- Added optional creation of entities via param on their getter fns


this-username-is-taken:
LGTM!

PR #451, opened 2022-06-29
Removing the non-derived field that can be queried using filters

this-username-is-taken:


schema-lending-extended.graphql
```diff
@@ -1001,11 +969,25 @@ type Position @entity {
   dailySnapshots: [PositionSnapshot!]! @derivedFrom(field: "position")
 }
 
+# Unlike other snapshots that are taken at a fixed time interval. Position
+# snapshots should be taken after every event, including the opening and
+# closing events. This will prevent an ever growing number of snapshots
+# for positions that are not moving. As we are only recording the balance
+# in token amounts instead of in USD, this will work well.
 # Note that we only take snapshot for open positions
 type PositionSnapshot @entity {
```
0xbe1, line 978: the entity doesn't come with a timestamp field, then how do we order them?

this-username-is-taken: which entity doesn't come with a timestamp field?

schema-lending-extended.graphql
```diff
@@ -1001,11 +969,25 @@ type Position @entity {
   dailySnapshots: [PositionSnapshot!]! @derivedFrom(field: "position")
```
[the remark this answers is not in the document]
this-username-is-taken: ah, nice catch!

schema-lending-extended.graphql
0xbe1: my bad, saw it

PR #457, opened 2022-06-29
Updated execution scripts to retry and log on HTTP errors. 

steegecs:


deployment/execution.js
```diff
@@ -57,23 +57,38 @@ async function runCommands(allScripts, results, args, callback) {
     let logs = ""
     var index = 0;
     var index2 = 0;
+    var httpCounter = 1;
     let protocols = Array.from( allScripts.keys() );
 
     function next() {
```
this-username-is-taken, line 63: curious why do you make this a recursive call instead of a for loop or while loop?

deployment/execution.js
```diff
@@ -57,23 +57,38 @@ async function runCommands(allScripts, results, args, callback) {
     let logs = ""
     var index = 0;
     var index2 = 0;
```
this-username-is-taken, line 59: you should use better names especially there are so many counters now. Something like `forkIndex` and `networkIndex` are easier to understand

deployment/execution.js
```diff
@@ -57,23 +57,38 @@ async function runCommands(allScripts, results, args, callback) {
...
         if (index < protocols.length) {
             exec(allScripts.get(protocols[index])[index2++], function(error, stdout, stderr) {
             logs = logs + "stdout: "  + stdout
             logs = logs + "stderr: "  + stderr
+            if (stderr.includes("HTTP error")) {
+                if (httpCounter >= 2) {
+                    index++;
+                    index2 = 0;
+                }
+                httpCounter++
```
this-username-is-taken, line 73: don't forget to run `prettier`

deployment/execution.js
steegecs: The recursion causes the execution of each script to be awaited for completion before continuing on to the next execution or writing of the logs.

deployment/execution.js
steegecs: Good call!

deployment/execution.js
steegecs: Ran prettier.

PR #470, opened 2022-06-30
Add Geist Finance to `aave-v2-forks`

dmelotik:
Deliverables:
- Add Geist to Aave v2 forks

Watching:
- Make sure GEIST price is correct for reward purposes


this-username-is-taken:
Great work! Thanks for submitting the new PR.

PR #473, opened 2022-07-01
Compound extended

0xbe1:
Syncing https://okgraph.xyz/?q=messari%2Fcompound-ethereum-extended


0xbe1:
prefer https://github.com/messari/subgraphs/pull/480

PR #474, opened 2022-07-01
Trader Joe - FIxed MasterChef indexing like sushiswap

steegecs:


this-username-is-taken:
Don't forget to run `prettier`

this-username-is-taken:
Actually, maybe see if you can set up `prettier` with git commit hook

PR #483, opened 2022-07-02
update compound v2 to reduce handlers

dmelotik:
Reducing abis and handlers by "hacking" a modified aBI explained in readme


0xbe1:
LGTM

PR #486, opened 2022-07-02
Add Badger DAO Subgraph

harsh9200:


this-username-is-taken:
Great work! Can you add CI/CD and also update the README with some of the implementation details that people should know about (e.g. which pools are tracked/skipped), reward emission methodology etc?

this-username-is-taken:
You also need to update `https://github.com/messari/subgraphs/blob/master/deployment/deploymentConfigurations.json`. Probably should sync up with @steegecs on the right way to add CI/CD.

PR #487, opened 2022-07-02
Bug in Aave V2 Mainnet

dmelotik:
Fix Issue with Aave V2 normal / ext on mainnet

Issue: There was a reward token at one point, which is not even a token, and when the invalid decimals were `9999` it was too large, so the invalid decimals needed to be changed to 0.

See stripped down test here: https://okgraph.xyz/?q=dmelotik%2Faave-v2-ethereum-redo

Note: testing all of my aave-v2 subgraphs under the names listed in the deployments

@this-username-is-taken  I think you should turn off deployments, because other aave fork subgraphs have errors that I don't want reset. DM me when this is merged and i will deploy only the mainnet subgraphs to messari


this-username-is-taken:
Great work on combing through all the issues! These are not easy to debug.

PR #493, opened 2022-07-03
Reference Subgraph: Add more Array Utils

nemani:
adds: removeFromArrayAtIndex and addToArrayAtIndex to the utils for the reference subgraph


this-username-is-taken:
Thanks!

PR #496, opened 2022-07-04
Add Aave V3 position data

spitko:
- Added back Account.openPositions array
- Added liquidityIndex to Position entity to track `LENDER` positions
- Position balance snapshots are created only after an update to the balance. This means the snapshot values do not necessarily include all the interest accrued/paid on the position. The interest amounts can still be calculated from the positions's events.

### DIFF
--- a/deployment/deploymentConfigurations.json
+++ b/deployment/deploymentConfigurations.json
@@ -150,6 +150,40 @@
         }
       }
     },
+    "aave-v3": {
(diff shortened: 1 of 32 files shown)


spitko:
Deployed to https://okgraph.xyz/?q=spitko%2Faave-v3-matic-extended

this-username-is-taken:
LGTM!

PR #534, opened 2022-07-07
Re-create Aave v2 forks back to 1.3.0

dmelotik:
This re creates the aave v2 forks without the extended schema. 

Note:
- This does not fix the current aave issue in `_handleReserveDataUpdated`

Testing: https://okgraph.xyz/?q=dmelotik%2Faave-v2-ethereum


this-username-is-taken:
Thanks!

PR #543, opened 2022-07-10
Lido Finance

fortysevenlabs:
Validation
* [Subgraph link for validation](https://thegraph.com/hosted-service/subgraph/fortysevenlabs/lido)

To review:
- [x] Rewards are minted in stETH. Do we want to calculate revenues with stETH price ([see this](https://github.com/messari/subgraphs/compare/master...fortysevenlabs:messari-subgraphs:master#diff-f425ab52dd0bdc8463cd9d36191e6cfe4e116c6e5ad90fd09a63da2e8b4f462dR165))
- [x] YearnlensOracle token prices accuracy (~currently commented out~, minor differences, ok to use, re-enabled)


subgraphs/lido/src/entityUpdates/financialMetrics.ts
```diff
@@ -0,0 +1,352 @@
...
+  amount: BigInt
+): void {
+  const pool = getOrCreatePool(block.number, block.timestamp);
+  const protocol = getOrCreateProtocol();
+
+  // inputToken is ETH, price with ETH
+  const amountUSD = bigIntToBigDecimal(amount).times(
+    getOrCreateToken(Address.fromString(ETH_ADDRESS), block.number)
+      .lastPriceUSD!
+  );
```
harsh9200, line 31: unused variable

subgraphs/lido/src/entityUpdates/financialMetrics.ts
```diff
@@ -0,0 +1,352 @@
...
+  // Pool
+  pool.cumulativeSupplySideRevenueUSD = pool.cumulativeTotalRevenueUSD.minus(
+    pool.cumulativeProtocolSideRevenueUSD
+  );
+  pool.save();
+
+  // Pool Daily
+  poolMetricsDailySnapshot.cumulativeSupplySideRevenueUSD =
+    pool.cumulativeSupplySideRevenueUSD;
+  poolMetricsDailySnapshot.dailySupplySideRevenueUSD = poolMetricsDailySnapshot.dailyTotalRevenueUSD.minus(
```
harsh9200, line 232: Some of the values in `dailySupplySideRevenueUsd` is zero ([Link](https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fname%2Ffortysevenlabs%2Flido&tab=pool&poolId=0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84&view=poolDailySnapshots-dailySupplySideRevenueUSD)). Can you add a condition here to only update the field when the revenue is positive?

harsh9200 (commented):
Great Work! As for the revenue, it should be calculated in `stETH` instead of `ETH`, it will be more accurate and logically correct.

fortysevenlabs:
> As for the revenue, it should be calculated in stETH instead of ETH, it will be more accurate and logically correct.
Exactly, thanks for the confirmation, will continue to price using ETH for input/deposit but price the output/stakingRewards in stETH for accuracy.

fortysevenlabs:
> As for the revenue, it should be calculated in stETH instead of ETH, it will be more accurate and logically correct.
Exactly, thanks for the confirmation, will continue to price using ETH for input/deposit but price the output/stakingRewards in stETH for accuracy.

subgraphs/lido/src/entityUpdates/financialMetrics.ts
fortysevenlabs: Pushed the change.

harsh9200:
> Actually, thinking further about this. Since the rewards are issued to the protocol from ETH2 deposit contract in the form of ETH (which is then used to rebase the user with stETH shares), we should be fine calculating revenues using ETH. stETH would have been necessary if we were tracking user's balances/value.

Sounds Good, LGTM!

this-username-is-taken:
Great work! Would you be able to add CI/CD to it? See here for details: https://github.com/messari/subgraphs/tree/master/deployment

this-username-is-taken:
Also, if you could update Lido here: https://github.com/messari/subgraphs#development-status as well

fortysevenlabs:
Will add/update:
- [ ] CI/CD
- [ ] Dev status

this-username-is-taken:
Thanks Tarun!

fortysevenlabs:
What's the merge and deploy plan?

Are you or your team validating [the graph](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/fortysevenlabs/lido-ethereum) as well?

fortysevenlabs:
Cool, that was fast, thanks for the merge 👍

PR #553, opened 2022-07-11
Convex Finance: QA Update

harsh9200:
- Added Missing pools support which is used to calculate token prices. (cause of low tvl) .
- Added `cvx` token as a rewardToken (There is no event for the rewarded amount, its calculated based on the logic the official subgraph is using).
- Added standard way to calculate reward token emission amount (present in _reference_ folder).


this-username-is-taken:
great work!

Issue #500, opened 2022-07-05
MakerDAO QA (Ethereum) Schema Version 1.2.1 Subgraph Version 1.0.1 Methodology Version 1.0.0


tnkrxyz:
I am documenting here the issues I found with the makerdao subgraph in the master branch (subgraphVersion 1.0.1, schema 1.3.0), as I work on developing the makedao extended subgraph:

# data issues
- [x] 1. ~~"from" and "liquidatee" in Liquidate entity are always the same, pointing to the account triggering vat.grab~~(`tx.transaction.from` logged for the function); according to the schema "from" - Address that carried out the liquidation; "liquidatee" - Address that got liquidated.
- [x] 2. ~~Cat_v2 contract in subgraph.yaml (line 193) is not using the correct contract address~~[0xa5679C04fc3d9d8b0AaB1F0ab83555b301cA70Ea](https://etherscan.io/address/0xa5679C04fc3d9d8b0AaB1F0ab83555b301cA70Ea#code) and still use v1 address of [0x78f2c2af65126834c51822f56be0d7469d7a523e](https://etherscan.io/address/0x78f2c2af65126834c51822f56be0d7469d7a523e#code), likely leading to incorrect mapping of liquidations
- [x] 3. ~~`Created` event from the `DSProxyFactory` contract (creating a proxy) should not trigger `updateUsageMetrics` (src/mappings/users/users.ts line 13), leading to incorrect usage metrics~~
- [x] 4. CdpManager
    - [x] ~~should `give` (transfer a cdp position) contract trigger `updateUsageMetrics`?~~
    - [x] ~~frob event should not trigger `updateUsageMetrics`, as it calls vat.frob, which again triggers `updateUsageMetrics`~~
- [x] 5 ~~pot contract has not been mapped, thus daisavingrate & supplysiderevenue has not been considered, leading to zero supplysiderevenue and missing deposit rates~~
- [x] 6 ~~Market.OutputToken: DAI (borrow) is not the outputToken~~
- [x] 7 ~~IDs for transaction entities (Borrow, Repay, Deposit, Withdraw, Liquidate) do not confirm to the schema specification of "{ Transaction hash }-{ Log index }"~~
- [x] 8 ~~incorrect inputTokenBalance for USDC-A market (#610)~~

# code issues
- [x] ~~deployment failure due to non-nullable fields error in graph-ts version 0.30;~~
[...]

PR #555, opened 2022-07-12
Fixing makerdao subgraph

tnkrxyz:
This PR fixes makerdao subgraph issues listed in #500, improve code readability, organization and documentation.
- [x] Debug deployment failure
- [x] Verify issues fixed (commented in #500)


subgraphs/makerdao/src/utils/numbers.ts
```diff
@@ -0,0 +1,43 @@
...
+  BIGINT_ZERO,
+  BIGDECIMAL_TWO,
+  BIGDECIMAL_ONE,
+  BIGDECIMAL_SIX,
+  BIGDECIMAL_TWELVE,
+  BIGDECIMAL_THREE,
+  BIGDECIMAL_ONE_THOUSAND,
+} from "../common/constants";
+
+export function bigIntToBDUseDecimals(quantity: BigInt, decimals: i32 = 18): BigDecimal {
```
this-username-is-taken, line 15: This is nice. Can you add this to the common lib in the reference subgraph?

subgraphs/makerdao/src/utils/numbers.ts
```diff
@@ -0,0 +1,43 @@
...
+  return quantity.divDecimal(
+    BigInt.fromI32(10)
+      .pow(decimals as u8)
+      .toBigDecimal(),
+  );
+}
+
+export function bigDecimalExponential(rate: BigDecimal, exponent: BigDecimal): BigDecimal {
```
this-username-is-taken, line 23: same here

subgraphs/makerdao/src/utils/numbers.ts
```diff
@@ -0,0 +1,43 @@
...
+    .times(exponent.minus(BIGDECIMAL_TWO))
+    .times(rate.times(rate).times(rate));
+  let fourthTerm = exponent
+    .div(BIGDECIMAL_TWELVE)
+    .times(exponent.minus(BIGDECIMAL_THREE))
+    .times(rate.times(rate).times(rate).times(rate));
+  return firstTerm.plus(secondTerm).plus(thirdTerm).plus(fourthTerm);
+}
+
+export function round(numberToRound: BigDecimal): BigDecimal {
```
this-username-is-taken, line 39: and here

subgraphs/makerdao/src/utils/numbers.ts
tnkrxyz: Will do after I get the subgraph to work correctly. Some of the functions I borrowed from @[tannishmango](https://github.com/tannishmango)'s implement

tnkrxyz: Looks like this function is already in there `_reference/src/common/utils/numbers.ts`: `bigIntToBigDecimal`

tnkrxyz:
A temporary deployment of the current implement is available at: https://api.studio.thegraph.com/query/29379/makerdao-ethereum-v2/v0.0.2

A few things to note:
- Since the lending schema expects supply side revenue from a market, a 'market' is created for the DAI saving contract POT, even though it is a real market.
- About 400 liquidations have negative profitUSD, one as high as -$17,381.73 (liquidate.id=0x8f06a94a522dff90a445eadc7569919abab4caee936e011aed1517acca2dc404-245). After double, triple check, as far as I can see the numbers are correct, but they seem odd. For example, for 0x8f06a94a522dff90a445eadc7569919abab4caee936e011aed1517acca2dc404-245, when it was flag for liquidation, 89.61790614077303 ETH was seized for a loan of $50,000. When the auction started, the bids went to 85.450979507487898432 ETH, 55.741671655855377187 ETH, and finally 42.361384320354736031 ETH after 25 rounds of bidding. With the price of ETH at $754.01, the collateral worth $31,940.91 and a net loss of -$17,381 to the liquidator.

this-username-is-taken:
Really good fixes. Thanks for being so meticulous with this one!

Issue #547, opened 2022-07-11
Solarbeam QA (Moonriver) Schema Version 1.3.0 Subgraph Version 1.0.0 Methodology Version 1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/solarbeam-moonriver&tab=protocol
Date Reviewed | July 11 2022
Schema Version | 1.3.0
Subgraph Version | 1.0.0
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1NOS8FPgnyJa-a3rKe_KcVbx5SZY-TZRa7JhU1z8QwPg/edit?usp=sharing

### Metrics To Review
**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | dailyVolumeUSD | Trend looks good, but there are certain days where the volume reported on the subgraph is over 100% difference than Solarbeam UI. This could be a Solarbeam UI problem, but difference is worth looking into
financialsDailySnapshots | dailySupplySideRevenueUSD | 2% fee applied supposed to be 0.2% fee
financialsDailySnapshots | cumulativeSupplySideRevenueUSD | Looks good, but daily is off
financialsDailySnapshots | dailyProtocolSideRevenueUSD | 0.5% fee applied supposed to be 0.05% fee
financialsDailySnapshots | cumulativeProtocolSideRevenueUSD | Looks good, but daily is off
financialsDailySnapshots | dailyTotalRevenueUSD | 2.5% fee applied supposed to be 0.25% fee
financialsDailySnapshots | cumulativeTotalRevenueUSD | Looks good, but daily is off

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
ALL POOLS | Reward Token | Missing SOLAR as a reward token
Solarbeam MOVR/xcKSM | Base Yield % | 930.00% being outputted, Solarbeam UI is 11.58%
SolarBeam MOVR/USDC | Base Yield % | 100010.56% being outputted, Solarbeam UI is 154.95%
Solarbeam ETH/WMOVR | Base Yield % | 19424.32% being outputted, Solarbeam UI is 43.40%
Solarbeam FRAX/ROME | ALL | This pool is the number one pool in TVL on the Solarbeam UI, but is not present in the subgraph's Pool Overview

**Pool Metrics**
[...]

PR #563, opened 2022-07-12
Solarbeam - Issue 547 - Fixed pricing, pool tvl, and update masterchef contracts 

steegecs:


this-username-is-taken:
Fixes issues in https://github.com/messari/subgraphs/issues/547

Issue #579, opened 2022-07-14
AAVE V3 QA (Avalanche, Optimism, Arbitrum, Harmony, Polygon, Fantom) Schema Version 2.0.1  Subgraph Version  1.0.2  Methodology Version 1.0.0


ishraq8:
<google-sheets-html-origin><style type="text/css"></style>

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/aave-v3-avalanche&tab=protocol |   |   |   |   |  
-- | -- | -- | -- | -- | -- | --
Date Reviewed | 7-Jul-22 |   |   |   |   |  
Schema Version | 1.3.0 |   |   |   |   |  
Subgraph Version | 1.0.0 |   |   |   |   |  
Methodology Version | 1.0.0 |   |   |   |   |  
  |   |   |   |   |   |  
Protocol |   |   |   |   |   |  
Section | Value | Status |   | Comments |   |  
usageMetricsDailySnapshots | dailyTransactionCount | NOT OK | Public data | Subgraph data is more than the snowtrace data. either multiplier or something else. Sourncing data: https://snowtrace.io/txs?a=0x794a61358D6845594F94dc1DB02A252b5b4814aD&ps=100&p=7 |   |  
usageMetricsDailySnapshots | dailyDepositCount | NOT OK | Public data | Subgraph data is more than the snowtrace data. either multiplier or something else. Sourncing data: https://snowtrace.io/txs?a=0x794a61358D6845594F94dc1DB02A252b5b4814aD&ps=100&p=7 |   |  
usageMetricsDailySnapshots | dailyWithdrawCount | NOT OK | Public data | Subgraph data is more than the snowtrace data. either multiplier or something else. Sourncing data: https://snowtrace.io/txs?a=0x794a61358D6845594F94dc1DB02A252b5b4814aD&ps=100&p=7 |   |  
usageMetricsDailySnapshots | dailyBorrowCount | NOT OK | Public data | Subgraph data is more than the snowtrace data. either multiplier or something else. Sourncing data: https://snowtrace.io/txs?a=0x794a61358D6845594F94dc1DB02A252b5b4814aD&ps=100&p=7 |   |  
[...]

PR #591, opened 2022-07-15
Aave V3: Convert interest rates to APY

spitko:
#579


this-username-is-taken:
Thanks for the change!

PR #622, opened 2022-07-19
Platypus Fix

aditya-todi:
- Handle unexpected assets in master chef contract.
- Moved alt pools initialisation from code to deployment config.
- Added steps to add new pool in readme.


this-username-is-taken:
Thanks for the fix!

PR #585, opened 2022-07-15
Initial OZ Governor subgraphs 

danielkhoo:
## Description
Following up on #558. This PR migrates some of our existing subgraphs to the new shared format. Also removes the parent governance directory in order to enable CI deployment.

For the initial test only `ens-governance` should deploy when this PR is merged

## Subgraph Links
https://thegraph.com/hosted-service/subgraph/danielkhoo/hop-governance
https://thegraph.com/hosted-service/subgraph/danielkhoo/ens-governance
https://thegraph.com/hosted-service/subgraph/danielkhoo/fei-governance
https://thegraph.com/hosted-service/subgraph/danielkhoo/silo-governance
https://thegraph.com/hosted-service/subgraph/danielkhoo/euler-governance

As a side note, I've found Tally https://www.tally.xyz/governance/eip155:1:0x323A76393544d5ecca80cd6ef2A560C6a395b7E3
to be a convenient sanity check for the subgraph numbers.

## Misc
Includes tiny fix for bug cause total proposals to be 2x the correct number.

PR #623, opened 2022-07-19
OZGovernor Subgraphs Batch 2

danielkhoo:
## Description

Building on the OpenZeppelin Governor template #585. 

Adds the following new subgraphs:
- Unlock Protocol
- Angle Finance
- Code4Rena (Polygon)

## Misc
- Updated the deployment naming convention from `mainnet` to `ethereum` 
- Updated the main README to reflect the new OZGovernor subgraphs


subgraphs/openzeppelin-governor/protocols/angle-governance/src/governor.ts
```diff
@@ -0,0 +1,127 @@
...
+// Helper function that imports and binds the contract
+function getGovernanceFramework(contractAddress: string): GovernanceFramework {
+  let governanceFramework = GovernanceFramework.load(contractAddress);
+
+  if (!governanceFramework) {
+    governanceFramework = new GovernanceFramework(contractAddress);
+    let contract = Governor.bind(Address.fromString(contractAddress));
+
+    governanceFramework.name = contract.name();
+    governanceFramework.type = "OZGovernor";
```
mmnavarr, line 110: We should pull this out to a `const` now that its being re-used across the subgraphs. Will likely be adding more of these framework types which can go in the same place

subgraphs/openzeppelin-governor/protocols/code4rena-governance/config/templates/code4rena-governance.template.yaml
```diff
@@ -0,0 +1,79 @@
...
+        - Vote
+        - TokenHolder
+        - Delegate
+      abis:
+        - name: ArenaGovernor
+          file: ./abis/code4rena-governance/ArenaGovernor.json
+      eventHandlers:
+        # Part of GovernorPreventLateQuorum 
+        # Our entities currently don't track deadline
+        # - event: LateQuorumVoteExtensionSet(uint64,uint64)
```
mmnavarr, line 29: I'd be curious to know what fields are involved in this event and how they would effect the proposal entity fields

Feels like we should be handling it.. In a universe where we have a robust notification system (soon) this feels like a case that would be of importance to us

mmnavarr: Goes for the below `ProposalExtended` event as well

mmnavarr (dismissed):
Minor feedback on `governanceFramework.type` options being moved to a const and some thoughts on the quorum extension

Would like to see the framework types migrated to const prior to merge but we should talk about quorum extension use case as a separate discussion (not blocking)

subgraphs/openzeppelin-governor/protocols/code4rena-governance/config/templates/code4rena-governance.template.yaml
danielkhoo: `LateQuorumVoteExtensionSet` changes the default number of blocks for an extension. imo it doesnt really serve any purpose to track.

But on second looks, i realise we can support `ProposalExtended` simply be updating the `endBlock`

danielkhoo: added tiny commit with extra handler for ProposalExtended

Issue #618, opened 2022-07-18
Uniswap V3 QA (Arbitrum) Schema Version 1.3.0 Subgraph Version 1.1.1 Methodology Version 1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/uniswap-v3-arbitrum&tab=protocol
Date Reviewed | July 18 2022
Schema Version | 1.3.0
Subgraph Version | 1.1.1
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1nhu2rqFYM8TuV9DlrksQTAOnP-3L-WsY/edit?usp=sharing&ouid=113156017090580515789&rtpof=true&sd=true

### Metrics To Review
**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | totalValueLockedUSD | The subgraph TVL is about 20% less than the TVL from DefiLlama. Furthermore, the trends look decent, except for between 02/06/2022 - 02/08/2022 there is a spike in the subgraph that is different than both DefiLlama and Uniswap UI. See protocol tab on evidence sheet.
financialsDailySnapshots | dailySupplySideRevenueUSD | QA'er made a note that the recent average daily revenue/volume on subgraph (Arbitrum) is higher than Dune's data (Uniswap V3 on Ethereum), which may warrant attention. I think this metric is most likely fine since the daily volume seems to be ok, but just want to make sure that we are accounting for Uniswap V3 having different fees for different pools. If this is the case, then this is probably ok.

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
All Pools | Fee Labels | Missing labels for pools with different fees since there can be pools of the same assets, but with different fees. For example, WBTC/ETH 0.3% and WBTC/ETH 0.05%
Some Pools | Naming for ETH | Pools with ETH are missing ETH in the pool name and in the input tokens. Right now it is just blank. For example, Uniswap V3 ETH / GMX is being shown as "Uniswap V3 /GMX" and the input tokens are being shown as ", GMX"
Some Pools | Naming for USDC and USDT | Pools with USDC and USDT seem to be having a weird name convention where in the pool name they are being called arbFiatToken and being output as "FIAT" in the input token section
All Pools | Base Yield | Yields are in the high hundreds and thousands for most pools, which is not correct. Would be great if true though.
[...]

PR #629, opened 2022-07-19
Issue (#618) - Uniswap V3 - Fixed pool naming to include fee percentage and WETH and fiat names

steegecs:


this-username-is-taken:
Oh, I meant including in the description as this https://github.com/messari/subgraphs/issues/618 which is really nice because it generates a link

Issue #587, opened 2022-07-15
Arrakis Finance QA (Ethereum) Schema Version 1.3.0 Subgraph Version 1.0.0 Methodology Version 1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=l0c4t0r/laboratory4&tab=protocol
Date Reviewed | July 14 2022
Schema Version | 1.3.0
Subgraph Version | 1.0.0
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1zlH-rlysSs5g7rb8ODd1cEhKL_IzuOm6/edit?usp=sharing&ouid=113156017090580515789&rtpof=true&sd=true

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
All Pools | Naming | The subgraph name does not include the pool commission rate of 0.01%, 0.05% etc. making it hard to differentiate. Would be nice if this was included.
Some Pools | TVL | Pools such as Gelato Un...DAI/FEI LP, Gelato Un...EI/USDC LP, Gelato Un...AR/USDC LP, Gelato Uniswap USDC/WETH LP, Gelato Uniswap GEL/WETH LP have differing TVLs compared to the Arrakis UI. See evidence sheet "Pool Overview" for more detail. Link to Arrakis - https://beta.arrakis.finance/vaults. I also checked the FEI/DAI contract and I am getting token balances that align with Arrakis UI and not the subgraph (4.6m vs 2.9m). Link to FEI/DAI etherscan under getUnderlyingTokenBalances - https://etherscan.io/address/0x3d1556e84783672f2a3bd187a592520291442539#readProxyContract

l0c4t0r:
Checking this now. It does seem like there is some issue with stale TVL

l0c4t0r:
Additional, the price of FEI from price lib seems to be under-reported ($0.35 vs ~$1)

l0c4t0r:
I have opened the above PR to address some of the issues mentioned above.  However, not all vault TVL are "fixed".  I elaborate on the reasons for the discrepancies here:

1. Price lib in it's current form doesn't always return the most accurate data. In this specific case, Yearn lens was returning $0.35 for FEI, while it should have been ~$1 (reported correctly by chainlink)
[...]

PR #632, opened 2022-07-20
Address Arrakis QA issues

l0c4t0r:
Partially addresses QA issues in #587 .

1.  Improved vault naming by adding fee tier suffix to vault name.
2. Yearn Lens was giving bad pricing data for FEI ($0.35 vs $1), skip Yearn Lens for FEI when getting prices.

This fixes the TVL for the vault with the greatest disparity.  There are still some differences in TVL values vs those reported  by Arrakis for smaller vaults that account for ~0.1% of Arrakis total TVL.  I have no address these in this PR as these are due to more systematic issues around how we update TVL, and price lib support.  I elaborate on these in #587


l0c4t0r:
Deployed at l0c4t0r/laboratory4 for verification.

this-username-is-taken:
Would you be able to change this to true so it'll auto-deploy?
https://github.com/messari/subgraphs/blob/fc9a2936ba7289aecd6d867301ff599a1ab6e1e0/deployment/deployment.json#L241

l0c4t0r:
Done!

this-username-is-taken:
Thanks for the fix!

Issue #556, opened 2022-07-12
Saddle Finance QA (Ethereum) Schema Version 1.3.0 Subgraph Version 1.1.0 Methodology Version 1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/saddle-finance-ethereum&tab=protocol
Date Reviewed | July 12 2022
Schema Version | 1.3.0
Subgraph Version | 1.1.0
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1d-BxTuefAKYALD6YjyI8aeV1K5PPj8OCxIUpR__7XF8/edit?usp=sharing

### Metrics To Review
**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | totalValueLockedUSD | Trend is similar to DefiLlama and token terminal, but they have the exact same trend. Subgraph values are slightly higher 5%-15% for recent data, which is ok, but older data shows more disparity (>40%)
financialsDailySnapshots | dailyVolumeUSD | There are quite large differences, with subgraph having generally higher values from 55% to 437% difference in volume. Erroneous peak in subgraph data on Apr 30 (>160 million in volume)
financialsDailySnapshots | cumulativeVolumeUSD | Off because daily is
financialsDailySnapshots | dailySupplySideRevenueUSD | Values are generally higher on token terminal, in the range of 20-45%. This might be a combination of how we we are calculating revenue and that daily volume is off
financialsDailySnapshots | cumulativeSupplySideRevenueUSD | Off because daily is
financialsDailySnapshots | dailyProtocolSideRevenueUSD | I think subgraph data is right here, but Saddle Finance claims this is 0 in their docs. I believe that the admin fee is being applied now, but need to find source that says that admin fees started on 1/19/2022 and double check. For now, I will go with the "official" documentation and make a note. 
financialsDailySnapshots | cumulativeProtocolSideRevenueUSD | Off because daily is
financialsDailySnapshots | dailyTotalRevenueUSD | Off because protocol and supply side are
financialsDailySnapshots | cumulativeTotalRevenueUSD | Off because daily is

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
ALL POOLS | Overall Pool Count | Missing 4 pools compared to what is showing on the Saddle UI.
[...]

PR #635, opened 2022-07-20
Saddle Finance: update dependency versions, fix inputTokens order 

spitko:
#556 

All fields related to pool.inputTokens should now be fixed, protocol.totalPoolCount should now match actual pool count.


this-username-is-taken:
Thanks for the fixes. Can you update https://github.com/messari/subgraphs/blob/master/deployment/deployment.json#L603-L615 to true so it'll deploy?

PR #639, opened 2022-07-20
Uniswap V3 - Added README comments for Optimism regenesis

steegecs:


subgraphs/uniswap-v3/README.md
```diff
@@ -38,19 +47,19 @@ Note that Pool Fee Tiers vary by pool and more tiers could be added by Governanc
...
 `Swaps`
 
 `Deposits`
 
 `Withdraws`
 
-###  Reward Token Emissions Amount
+### Reward Token Emissions Amount
 
 To be added
```
this-username-is-taken, line 60: should probably update this and Protocol Controlled Value (does not apply).

steegecs: Just add does not apply like so? Or do you think it's better to just remove them?

PR #640, opened 2022-07-21
add nft marketplace schema

0xbe1:
Minimal schema for NFT marketplace.

### DIFF
--- a/schema-nft-marketplace.graphql
+++ b/schema-nft-marketplace.graphql
@@ -61,6 +61,9 @@ type Marketplace @entity {
     " Number of collections listed on the marketplace. "
     collectionCount: Int!
 
+    " Trade count of the all collections on the marketplace. "
+    tradeCount: Int!
+
     " Cumulative trade volume (in ETH) "
     cumulativeTradeVolumeETH: BigDecimal!
 
@@ -72,9 +75,6 @@ type Marketplace @entity {
 
     " Sum of marketplaceRevenueETH and creatorRevenueETH. "
     totalRevenueETH: BigDecimal!
-
-    " Trade count of the all collections on the marketplace. "
-    totalTradeCount: Int!
 
     " Cumulative unique traders "
     cumulativeUniqueTraders: Int!
@@ -99,7 +99,7 @@ type Collection @entity {
     " Royalty fee rate in percentage. E.g. 2.5% should be 2.5 "
     royaltyFee: BigDecimal!
 
-    " Number of tokens listed for sale in the marketplace. "
+    " Number of tokens listed for trade in the marketplace. "
     listedCount: BigInt!
 
     " Floor price (in ETH) "
@@ -118,7 +118,7 @@ type Collection @entity {
     totalRevenueETH: BigDecimal!
 
     " Trade count of the collection on the marketplace. "
-    totalTradeCount: Int!
+    tradeCount: Int!
 
     " Trades of the collection. "
     trades: [Trade!]! @derivedFrom(field: "collection")
@@ -173,8 +173,8 @@ type MarketplaceDailySnapshot @entity {
     " Number of collections listed on the marketplace. "
     collectionCount: Int!
 
-    " Cumulative sale volume (in ETH) "
-    cumulativeSaleVolumeETH: BigDecimal!
+    " Cumulative trade volume (in ETH) "
+    cumulativeTradeVolumeETH: BigDecimal!
 
     " Revenue that goes to the marketplace protocol, aka protocol fee. "
     marketplaceRevenueETH: BigDecimal!
@@ -186,13 +186,19 @@ type MarketplaceDailySnapshot @entity {
     totalRevenueETH: BigDecimal!
 
     " Trade count of the all collections on the marketplace. "
-    totalTradeCount: Int!
+    tradeCount: Int!
 
     " Cumulative unique traders "
     cumulativeUniqueTraders: Int!
 
     " Daily active traders "
     dailyActiveTraders: Int!
+
+    " Number of traded collections of the day "
+    dailyTradedCollectionCount: Int!
+
+    " Number of traded items of the day "
+    dailyTradedItemCount: Int!
 }
 
 type CollectionDailySnapshot @entity {
@@ -230,5 +236,8 @@ type CollectionDailySnapshot @entity {
     totalRevenueETH: BigDecimal!
 
     " Trade count of the collection on the marketplace. "
-    totalTradeCount: Int!
+    tradeCount: Int!
+
+    " Number of traded items of the day "
+    dailyTradedItemCount: Int!
 }
--- a/schema-nft-marketplace.graphql
+++ b/schema-nft-marketplace.graphql
@@ -129,6 +129,12 @@ type Collection @entity {
     " Trade count of the collection on the marketplace. "
     tradeCount: Int!
 
+    " Buyer count. "
+    buyerCount: Int!
+
+    " Seller count. "
+    sellerCount: Int!
+
     " Trades of the collection. "
     trades: [Trade!]! @derivedFrom(field: "collection")
 }
--- a/schema-nft-marketplace.graphql
+++ b/schema-nft-marketplace.graphql
@@ -135,11 +135,14 @@ type Collection @entity {
 
 " Trades exist such as a combination of taker/order and bid/ask. "
 type Trade @entity {
-    " { Transaction hash }-{ Log index } "
+    " { maker txn hash }-{ taker txn hash } "
     id: ID!
 
-    " Transaction hash of the trade. "
-    transactionHash: String!
+    " Transaction hash of the maker. "
+    makerTransactionHash: String!
+
+    " Transaction hash of the taker. "
+    takerTransactionHash: String!
 
     " Event log index. "
     logIndex: Int!
--- a/schema-nft-marketplace.graphql
+++ b/schema-nft-marketplace.graphql
@@ -1,0 +1,160 @@
+# Subgraph Schema: NFT Marketplace
+# Version: 0.0.1
+
+enum Network {
+  ARBITRUM_ONE
+  ARWEAVE_MAINNET
+  AURORA
+  AVALANCHE
+  BOBA
+  BSC # aka BNB Chain
+  CELO
+  COSMOS
+  CRONOS
+  MAINNET # Ethereum Mainnet
+  FANTOM
+  FUSE
+  HARMONY
+  JUNO
+  MOONBEAM
+  MOONRIVER
+  NEAR_MAINNET
+  OPTIMISM
+  OSMOSIS
+  MATIC # aka Polygon
+  XDAI # aka Gnosis Chain
+}
+
+enum NftStandard {
+    ERC721
+    ERC721A
+    ERC1155
+}
+
+enum SaleStrategy {
+    " Strategy that executes an order at a fixed price that can be taken either by a bid or an ask. "
+    StandardSale
+
+    " Strategy that executes an order at a fixed price that can be matched by any tokenId for the collection. "
+    AnyItemFromCollection
+
+    " Strategy that executes an order at a fixed price that can be matched by any tokenId in a set of tokenIds. "
+    AnyItemFromSet
+
+    " Strategy to launch a Dutch Auction for a token where the price decreases linearly until a specified timestamp and end price defined by the seller. "
+    DutchAuction
+
+    " Strategy to set up an order that can only be executed by a specific address. "
+    PrivateSale
+}
+
+type Marketplace @entity {
+    " Smart contract address of the protocol's main contract (Factory, Registry, etc) "
+    id: ID!
+
+    " Name of the NFT marketplace, for example LooksRare "
+    name: String!
+
+    " The blockchain network this subgraph is indexing on. "
+    network: Network!
+
+    " Number of collections listed on the marketplace. "
+    collectionCount: Int!
+
+    " Cumulative sale volume (in ETH) "
+    cumulativeSaleVolumeETH: BigDecimal!
+
+    " Cumulative sale volume (in USD)"
+    cumulativeSaleVolumeUSD: BigDecimal!
+
+    " Revenue that goes to the marketplace protocol, aka protocol fee. "
+    marketplaceRevenue: BigDecimal!
+
+    " Revenue that goes to creator, aka royalty fee. "
+    creatorRevenue: BigDecimal!
+
+    " Sum of marketplaceRevenue and creatorRevenue. "
+    totalRevenue: BigDecimal!
+
+    " Trade count of the all collections on the marketplace. "
+    totalTradeCount: Int!
+}
+
+type Collection @entity {
+    " Contract address. "
+    id: ID!
+
+    " Collectio name, mirrored from the smart contract. "
+    name: String!
+
+    " Collection symbol, mirrored from the smart contract. "
+    symbol: String!
+
+    " Total supply of the collection, mirrored from the smart contract. "
+    totalSupply: BigInt!
+
+    " NFT Standard the collection uses. "
+    nftStandard: NftStandard!
+
+    " Royalty fee rate in percentage. E.g. 2.5% should be 2.5 "
+    royaltyFee: BigDecimal!
+
+    " Number of tokens for sale in the marketplace. "
+    listedCount: BigInt!
+
+    " Floor price (in ETH) "
+    floorPriceETH: BigDecimal!
+
+    " Floor price (in USD) "
+    floorPriceUSD: BigDecimal!
+
+    " Cumulative trade volume (in ETH) "
+    cumulativeTradeVolumeETH: BigDecimal!
+
+    " Cumulative trade volume (in USD)"
+    cumulativeTradeVolumeUSD: BigDecimal!
+
+    " Revenue that goes to the marketplace protocol, aka protocol fee. "
+    marketplaceRevenue: BigDecimal!
+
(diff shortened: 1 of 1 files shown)


schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    cumulativeSaleVolumeETH: BigDecimal!
+    " Cumulative sale volume (in USD)"
+    cumulativeSaleVolumeUSD: BigDecimal!
+    " Revenue that goes to the marketplace protocol, aka protocol fee "
+    marketplaceRevenue: BigDecimal!
+    " Revenue that goes to creator, aka royalty fee"
+    creatorRevenue: BigDecimal!
+    " Sum of marketplaceRevenue and creatorRevenue "
+    totalRevenue: BigDecimal!
+    totalTransactionCount: Int!
```
mrbrianhobo, line 91: i'm assuming the transaction count refers to the sale count? or would this also include erc721/1155 transfers?

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+
+enum NftStandard {
+    ERC721
+    ERC721A
+    ERC1155
+}
+
+enum SaleStrategy {
+    " Strategy that executes an order at a fixed price that can be taken either by a bid or an ask. "
+    StandardSaleForFixedPrice
```
mrbrianhobo, line 36: maybe can just be named `StandardSale` since the fixed price is mentioned in the comment and the only sale strategy that doesn't have a fixed price is `DutchAuction` (if i'm not mistaken)

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    " Strategy that executes an order at a fixed price that can be matched by any tokenId for the collection. "
+    AnyItemFromCollectionForFixedPrice
```
mrbrianhobo, line 38: same as above comment

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    " Strategy that executes an order at a fixed price that can be matched by any tokenId in a set of tokenIds. "
+    AnyItemInASetForFixedPrice
```
mrbrianhobo, line 40: same as above comment but maybe `AnyItemFromSet` to match `AnyItemFromCollection` wording

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    tokenId: BigInt!
+    " Price (in ETH) "
+    priceETH: BigDecimal!
+    " Price (in USD) "
+    priceUSD: BigDecimal!
+    strategy: SaleStrategy!
+    " Maker account address "
+    maker: String!
+    " Taker account address "
+    taker: String!
```
mrbrianhobo, line 108: i wonder if it makes more sense to have a `buyer` and `seller` field vs `maker` and `taker` since maker and taker doesn't give much information as to who is on the giving/receiving end of the NFT transaction. is there any situation that i'm not thinking of where the maker/taker is useful?

e.g. alice is the seller and bob is the buyer
- alice lists an NFT for a fixed price, bob buys it
  - alice is the maker and bob is the taker 
- bob bids on an NFT from alice, alice accepts it
  - alice is the taker and bob is the maker

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    PrivateSale
+}
+
+type Marketplace @entity {
+    id: ID!
+    name: String!
+    network: Network!
+    collectionCount: Int!
+    " Cumulative sale volume (in ETH) "
+    cumulativeSaleVolumeETH: BigDecimal!
```
this-username-is-taken, line 53: I think we should accumulate volume per token/currency, this way we don't need to worry about USD conversion. Otherwise price calculation can slow down the subgraph by a lot.

An example could be:
```
type VolumeByToken @entity {
  token: Token!
  volume: BigInt! # In token's native amount
}

cumulativeSaleVolumeByToken: [VolumeByToken]!
```
By doing this, we separate the volume by the denominated token and do USD conversion offline.

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    DutchAuction
+    " Strategy to set up an order that can only be executed by a specific address. "
+    PrivateSale
+}
+
+type Marketplace @entity {
+    id: ID!
+    name: String!
+    network: Network!
+    collectionCount: Int!
```
this-username-is-taken, line 51: can you add a new line between every two fields? It's quite difficult to read like this...

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,109 @@
...
+    cumulativeSaleVolumeETH: BigDecimal!
+    " Cumulative sale volume (in USD)"
+    cumulativeSaleVolumeUSD: BigDecimal!
+    " Revenue that goes to the marketplace protocol, aka protocol fee "
+    marketplaceRevenue: BigDecimal!
+    " Revenue that goes to creator, aka royalty fee"
+    creatorRevenue: BigDecimal!
+    " Sum of marketplaceRevenue and creatorRevenue "
+    totalRevenue: BigDecimal!
+    totalTransactionCount: Int!
```
this-username-is-taken, line 62: We should comment every field. This one in particular deserves a comment as to what is considered a "transaction" here. e.g. "sale transactions"?

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,160 @@
...
+    " Trade count of the all collections on the marketplace. "
+    totalTradeCount: Int!
+}
+
+type Collection @entity {
+    " Contract address. "
+    id: ID!
+
+    " Collectio name, mirrored from the smart contract. "
+    name: String!
```
this-username-is-taken, line 88: Most of the fields here including `name` and `symbol` are not required by the ERC721 interface. Are we going to give them a default value if there isn't one? Or should we leave them as optional?
https://eips.ethereum.org/EIPS/eip-721

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,160 @@
...
+    collectionCount: Int!
+
+    " Cumulative sale volume (in ETH) "
+    cumulativeSaleVolumeETH: BigDecimal!
+
+    " Cumulative sale volume (in USD)"
+    cumulativeSaleVolumeUSD: BigDecimal!
+
+    " Revenue that goes to the marketplace protocol, aka protocol fee. "
+    marketplaceRevenue: BigDecimal!
```
this-username-is-taken, line 71: This should also be an array if we aggregate revenue per listing token/currency, in which case it should be `BigInt` and comment should indicate it is in native amounts.

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,160 @@
...
+
+    " Revenue that goes to creator, aka royalty fee. "
+    creatorRevenue: BigDecimal!
+
+    " Sum of marketplaceRevenue and creatorRevenue. "
+    totalRevenue: BigDecimal!
+
+    " Trade count of the collection on the marketplace. "
+    totalTradeCount: Int!
```
this-username-is-taken, line 127: probably want something like this here:
`trades: [Trade!]! derivedFrom "collection"`

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,160 @@
...
+}
+
+" Trades exist such as a combination of taker/order and bid/ask. "
+type Trade @entity {
+    id: ID!
```
this-username-is-taken, line 132: probably a comment on `id` explaining how it's defined

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,222 @@
...
+    royaltyFee: BigDecimal!
+
+    " Number of tokens listed for sale in the marketplace. "
+    listedCount: BigInt!
+
+    " Floor price (in ETH) "
+    floorPriceETH: BigDecimal!
+
+    " Cumulative trade volume (in ETH) "
+    cumulativeTradeVolumeETH: BigDecimal!
```
this-username-is-taken, line 209: can you also add `dailyTradeVolumeETH`?

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,222 @@
...
+    marketplaceRevenueETH: BigDecimal!
+
+    " Revenue that goes to creator, aka royalty fee. "
+    creatorRevenueETH: BigDecimal!
+
+    " Sum of marketplaceRevenueETH and creatorRevenueETH. "
+    totalRevenueETH: BigDecimal!
+
+    " Trade count of the all collections on the marketplace. "
+    totalTradeCount: Int!
```
this-username-is-taken, line 186: can you also add `cumulativeUniqueTraders` and `dailyActiveTraders`?

this-username-is-taken: `cumulativeUniqueTraders` should be added to `Marketplace` as well

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,222 @@
...
+    cumulativeSaleVolumeETH: BigDecimal!
+
+    " Revenue that goes to the marketplace protocol, aka protocol fee. "
+    marketplaceRevenueETH: BigDecimal!
+
+    " Revenue that goes to creator, aka royalty fee. "
+    creatorRevenueETH: BigDecimal!
+
+    " Sum of marketplaceRevenueETH and creatorRevenueETH. "
+    totalRevenueETH: BigDecimal!
```
this-username-is-taken, line 183: I wonder if we should make these fields BigInt

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,222 @@
...
+    id: ID!
+
+    " Block number where the snapshot is taken. "
+    blockNumber: BigInt!
+
+    " Block timestamp when the snapshot is taken. "
+    timestamp: BigInt!
+
+    " Number of collections listed on the marketplace. "
+    collectionCount: Int!
```
this-username-is-taken, line 171: maybe also add a `tradedCollectionCount` or `activeCollectionCount` to indicate the # of collections that's being actively traded on a given day. Can also add something similar like `tradedItemCount` for total # items.

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,222 @@
...
+    marketplaceRevenueETH: BigDecimal!
+
+    " Revenue that goes to creator, aka royalty fee. "
+    creatorRevenueETH: BigDecimal!
+
+    " Sum of marketplaceRevenue and creatorRevenue. "
+    totalRevenueETH: BigDecimal!
+
+    " Trade count of the collection on the marketplace. "
+    totalTradeCount: Int!
```
this-username-is-taken, line 221: maybe also add a `tradedItemCount` here for the unique # of items being traded on a given day?

this-username-is-taken:
LGTM!

this-username-is-taken (dismissed):

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,255 @@
...
+    blockNumber: BigInt!
+
+    " Block timestamp when the snapshot is taken. "
+    timestamp: BigInt!
+
+    " Royalty fee rate in percentage. E.g. 2.5% should be 2.5 "
+    royaltyFee: BigDecimal!
+
+    " Number of tokens listed for sale in the marketplace. "
+    listedCount: BigInt!
```
harsh9200, line 230: why don't we add the following fields?
```
   "  # of unique users holding this collection/NFT "
    holdersCount: Int!

   "  # of unique users buying this collection/NFT "
    buyersCount: Int!

   " # of unique users selling this collection/NFT "
    sellersCount: Int!
```

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,255 @@
...
+
+    " Floor price (in ETH) "
+    floorPriceETH: BigDecimal!
```
harsh9200, line 233: why are we not interested in storing prices in USD?

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,255 @@
...
+    timestamp: BigInt!
+
+    " Block number where the trade is executed. "
+    blockNumber: BigInt!
+
+    " Collection involved "
+    collection: Collection!
+
+    " Token ID of the traded NFT. "
+    tokenId: BigInt!
```
harsh9200, line 160: why don't we create a separate entity for NFT, which can contain all the metadata-related information, and information about the owner, creator, number of transfers, and holding period(for how long the user is holding the NFT) of the NFT.

schema-nft-marketplace.graphql
this-username-is-taken: We will do holder in a separate subgraph. But it does make sense to separate buyers and sellers

schema-nft-marketplace.graphql
this-username-is-taken: Because it's kinda a pain. We either need to calculate USD prices which will slow down indexing by a lot, or we need to store each currency denomination separately, which complicates the schema. 99.9% of the stuff are denominated in ETH anyways, so I think it's fine using ETH and omitting NFTs that are prices in other tokens.

schema-nft-marketplace.graphql
this-username-is-taken: yes, we will have a separate subgraph for NFT holder data and NFT metadata

schema-nft-marketplace.graphql
harsh9200: Normally, a marketplace uses at max 3 different tokens for trading, I don't think it will slow down indexing that much. btw are we only targeting `mainnet` marketplaces? If not, NFT marketplaces on different chains are not dependent on ETH.

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,255 @@
...
+    creatorRevenueETH: BigDecimal!
+
+    " Sum of marketplaceRevenueETH and creatorRevenueETH. "
+    totalRevenueETH: BigDecimal!
+
+    " Trade count of the all collections on the marketplace. "
+    tradeCount: Int!
+
+    " Cumulative unique traders "
+    cumulativeUniqueTraders: Int!
```
mrbrianhobo, line 204: does this make more sense to be `cumulativeUniqueOwners` or is that different from what you mean by `cumulativeUniqueTraders` (to me, this feels like this is anyone who has interacted with a collection on the marketplace which might be what we want to capture)?

mrbrianhobo, line 204: i wonder if it makes more sense to name this `cumulativeUniqueUsers` instead of traders?

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,255 @@
...
+
+    " Daily active traders "
+    dailyActiveTraders: Int!
```
mrbrianhobo, line 207: same as above

schema-nft-marketplace.graphql
this-username-is-taken: I think traders is more explicit and technically accurate. We are not measuring any non-trading usage such as listing.

mrbrianhobo:
schema looks good to me 💯

schema-nft-marketplace.graphql
0xbe1: since our first phase involves mainly: opensea, looksrare and x2y2, they are all majorly mainnet marketplace so we are good :) we can always update the schema when needed

dmelotik: @0xbe1 do you think we should drop the ETH, this was it is more generic and can apply to other chains in the future? it could look like:

```
" Floor price (in native asset) "
floorPrice: BigInt!
```
If we did it like this we could have a field in the `MarketPlace` called `nativeToken` that contains a type like `Token` in other subgraphs.. Food for thought

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,261 @@
...
+    creatorRevenueETH: BigDecimal!
+
+    " Sum of marketplaceRevenueETH and creatorRevenueETH. "
+    totalRevenueETH: BigDecimal!
+
+    " Cumulative unique traders "
+    cumulativeUniqueTraders: Int!
+}
+
+type Collection @entity {
```
dmelotik, line 92: is it possible / make sense for this schema to have a field for `uniqueOwners`, `supplyMinted`, and maybe we can create a metric that finds the percent of supplyMinted / totalSupply ??

dmelotik (commented):
nice work !

schema-nft-marketplace.graphql
this-username-is-taken: Yeah, I think this is a good idea. But I guess it should still be bigDecimal

schema-nft-marketplace.graphql
```diff
@@ -0,0 +1,261 @@
+# Subgraph Schema: NFT Marketplace
+# Version: 0.0.1
```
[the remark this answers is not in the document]
this-username-is-taken: From past experience, I think probably a good idea to start with some thing less than `1.0.0`, and only change to `1.0.0` after we are done with the first subgraph, otherwise there might be more changes along the way and it'd be difficult to track versions

schema-nft-marketplace.graphql
this-username-is-taken: that should belong in the holder subgraph imo

schema-nft-marketplace.graphql
0xbe1: @dmelotik good idea!

though i prefer going by ETH in this iteration for simplicity, as there are not many marketplace that matters living outside mainnet

PR #644, opened 2022-07-21
Create Notional Methodology

SerenityFund:
Notional Finance methodology.


this-username-is-taken:
see https://github.com/messari/subgraphs/pull/656

PR #660, opened 2022-07-23
LooksRare

0xbe1:
The PR uses an updated schema that can be found in another PR https://github.com/messari/subgraphs/pull/688.

Deployment: check out https://okgraph.xyz/?q=messari%2Flooksrare-ethereum


schema-nft-marketplace.graphql
```diff
@@ -26,236 +26,239 @@ enum Network {
 }
 
 enum NftStandard {
-    ERC721
-    ERC721A
```
0xbe1, line 30: I find that ERC721A is not part of openzepplin https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/token - should we add it here?

schema-nft-marketplace.graphql
```diff
@@ -26,236 +26,239 @@ enum Network {
...
 
-    " Trades of the collection. "
-    trades: [Trade!]! @derivedFrom(field: "collection")
+  " Trades of the collection. "
+  trades: [Trade!]! @derivedFrom(field: "collection")
 }
 
 " Trades exist such as a combination of taker/order and bid/ask. "
 type Trade @entity {
-    " { maker txn hash }-{ taker txn hash } "
```
0xbe1, line 144: the ID is wrong, so i fix it

schema-nft-marketplace.graphql
```diff
@@ -26,236 +26,239 @@ enum Network {
...
 
-    " Price (in ETH) "
-    priceETH: BigDecimal!
+  " Price (in ETH) "
+  priceETH: BigDecimal!
 
-    " Stretegy that the trade is executed. "
-    strategy: SaleStrategy!
+  " The amount of tokens to transfer for the tokenId. It is set at 1 except for ERC1155 batch. "
+  amount: BigInt!
```
0xbe1, line 172: added `amount` because a ERC1155 trade could involve more than 1 token

schema-nft-marketplace.graphql
```diff
@@ -26,236 +26,239 @@ enum Network {
...
 
-    " Collection symbol, mirrored from the smart contract. Leave null if not available. "
-    symbol: String
+  " Collection symbol, mirrored from the smart contract. Leave null if not available. "
+  symbol: String
 
-    " Total supply of the collection, mirrored from the smart contract. "
-    totalSupply: BigInt!
+  " Total supply of the collection, mirrored from the smart contract. "
+  totalSupply: BigInt
```
0xbe1, line 103: set optional since `totalSupply` is not part of the ERC721/ERC1155 standard - it only exists in extension interfaces

mrbrianhobo:
> listedTokens: LooksRare stores orders off-chain, therefore I cannot access tokens that are currently on-sale through the smart contract.

since this is the case for the `listedTokens` field, i don't think we can extract this information with any marketplace so i would suggest removing this field. i forgot that orders are stored off-chain and originally thought that there might be an event emitted when listing an NFT (but that doesn't make sense). 

> floorPriceETH: For the similar reason, there is no way to tell the floor price of tokens that are on-sale

instead of floor price referring to tokens listed on-sale, i was thinking we could use this field for the floor price of sold tokens (fulfilled orders on-chain). i also agree with the sentiment that a historical all-time floor price for a collection doesn't make much sense, so i'd opt for it to be moved to the daily snapshot entity or renaming it something like `floorPriceETH7d` or `floorPriceETH24h` to make it clear what the current floor price of a collection is over a time period.

subgraphs/looksrare/src/mapping.ts
```diff
@@ -0,0 +1,456 @@
...
+    marketplace.save();
+  }
+  return collection;
+}
+
+function getOrCreateMarketplace(marketplaceID: string): Marketplace {
+  let marketplace = Marketplace.load(marketplaceID);
+  if (!marketplace) {
+    marketplace = new Marketplace(marketplaceID);
+    marketplace.name = "LooksRare";
```
this-username-is-taken, line 340: We should have a slug as well

0xbe1: added

mrbrianhobo (approved):
everything looks good 👍

subgraphs/looksrare/schema.graphql
```diff
@@ -0,0 +1,285 @@
...
+
+  " Cumulative unique traders "
+  cumulativeUniqueTraders: Int!
+}
+
+type Collection @entity {
+  " Contract address. "
+  id: ID!
+
+  " Collectio name, mirrored from the smart contract. Leave null if not available. "
```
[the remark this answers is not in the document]
this-username-is-taken: Good eye 👍

PR #672, opened 2022-07-26
Add polygon support for price-lib

l0c4t0r:
Add polygon support for price lib via:

- quickswap router
- sushiswap router

Keeping this as draft while I verify prices are being pulled correctly.


l0c4t0r:
This is now ready to be reviewed.
WBTC price was being reported inaccurately due to low liquidity of the WBTC/WMATIC pool (which is used to as part of the path to derive token price in the default method).
I have added a method to override the path to use for tokens on specific networks, which I have documented in the readme.  While not bulletproof I believe this is sufficient for now in order to bring basic pricing functionality to polygon.

To test prices are being retrieved correctly, #736 contains the price-lib changes from this PR for the Arrakis Finance subgraph, and is current deployed on l0c4t0r/laboratory3. Token prices can be viewed here:
https://api.thegraph.com/subgraphs/id/QmVcF8H7sLbGMs2dTWjCauF2gRDaCwqzT1dTsXb4ZyMDgi/graphql?query=query+MyQuery+%7B%0A++tokens%28where%3A+%7Bsymbol_not%3A+%22G-UNI%22%2C+symbol_not_contains%3A+%22RAKIS%22%7D%29+%7B%0A++++id%0A++++symbol%0A++++name%0A++++lastPriceUSD%0A++++lastPriceBlockNumber%0A++%7D%0A%7D&operationName=MyQuery

PR #679, opened 2022-07-27
Merge Compound Extended to Main

dmelotik:
Merging Compound Forks Extended onto main subgraphs.

Note: Most compound forks are not setup for extended schema.

The subgraphs that are:
- Moonwell > has not been deployed to Messari as an extended schema yet, so I updated the versions
- Compound > Updated the versioning and grafting from https://okgraph.xyz/?q=messari%2Fcompound-ethereum-extended to keep state / update versions

`rari-fuse` and `bastion-protocol` will not merge so we can still see the version that was QA'd. Those will update once the QA is complete.

The subgraphs that are not setup for the extended schema will not be redeployed, because they are already the same as the current non-extended version


this-username-is-taken:
This is awesome! Thanks for the details.

Issue #389, opened 2022-06-22
Future schema upgrades


this-username-is-taken:
- Add strategy entity and harvest events to yield aggregators.
- Add underlying tokens to input tokens in yield aggregators.
- Add more reward token types (in Aave-v3, reward emissions are per token: aToken/stableDebtToken/variableDebtToken).

steegecs:
- Add number of days or hours since unix timestamp as a field for entities that concatenate them to some prefix in the id (i.e. LiquidityPoolDailySnapshots).

dmelotik:
- Lending protocol: add `positionCounts` to snapshots (market/usage)

tnkrxyz:
Make these two fields of the Market entity in the lending protocol optional, outputToken is already optional.
- outputTokenSupply
- outputTokenPriceUSD

steegecs:
Yield Aggregatort
- outputTokenSupply is not required in the Vault entity but required in the Vault snapshots
- UsageMetricsDaily snapshot has totalPoolCount but the hourly snapshot does not.

dmelotik:
- transaction counts on a market-level 
- user/unique user counts on market-level

dmelotik:
- add a field at the protocol level called "protocol" to store the name without any versions. ie, "aave" instead of "aave-v2"

dmelotik:
- specify if `withdraw.amount / amountUSD` should include or exclude withdrawal fee if applicable.

Based on convo with @bye43 this should be excluded

this-username-is-taken:
add oracle to lending protocols

this-username-is-taken:
add aggregated token data info in dex subgraphs (similar to uniswap)

dmelotik:
Consider allowing multiple input tokens in lending protocols. For the case of Compound V3 markets

dmelotik:
Consider having rewards on the protocol level. In the case of TrueFi that is what is happening. @corerouter feel free to add any other comments or notes on this front! Thanks for bringing it up

dmelotik:
How can we show variable vs stable debt in lending positions? prolly by separating the two into different positions, but we need schema changes to allow for this i think

steegecs:
- For DEXs, does it make sense to have `from` and `to` fields in the swaps? A user should always be swapping with a pool, though, they may interact with a Router as a proxy. So I am thinking about removing `from` and `to` and just having `user`, `pool`, and maybe `router` in the case of swaps. I think this is more clear.

[...]

PR #686, opened 2022-07-28
Aave v3: Fix stable mode borrow balance, close borrower position on liquidation

spitko:
Also changed aToken transfer logic so that transfer volume is not included in market/protocol deposit and withdraw volume.

### DIFF
--- a/subgraphs/aave-v3/protocols/aave-v3/src/entities/event.ts
+++ b/subgraphs/aave-v3/protocols/aave-v3/src/entities/event.ts
@@ -50,7 +50,8 @@ export function createDeposit(
   event: ethereum.Event,
   reserve: Address,
   user: Address,
-  amount: BigInt
+  amount: BigInt,
+  isTransfer: boolean = false
 ): Deposit {
   if (amount.le(BIGINT_ZERO)) {
     log.critical("Invalid deposit amount: {}", [amount.toString()]);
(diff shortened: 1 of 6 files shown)


subgraphs/aave-v3/protocols/aave-v3/src/entities/event.ts
```diff
@@ -232,6 +232,12 @@ export function createLiquidate(
     market,
     PositionSide.LENDER
   );
+  const borrowerPosition = getOrCreateUserPosition(
+    event,
+    userAccount,
+    getMarket(debtAsset),
+    PositionSide.BORROWER
+  );
```
spitko, line 240: Not sure what to do with the borrower position during liquidation, right now there's no reference to the liquidation, it is just closed.

this-username-is-taken: Good point. Do you think it'd be helpful if we changed `Liquidate` entity to include both borrower position and lender position?

spitko: I do think it would be helpful, but it would also potentially make queries more complicated. Still, might be something to think about for next schema update?

this-username-is-taken: yep, included it in here: https://github.com/messari/subgraphs/issues/389

this-username-is-taken: How would you make the change without adding this position then?

spitko: I agree that it is probably best to include the second position under `Liquidate`, either in a separate field or make Liquidate.position an array?

this-username-is-taken: yeah, let's do a separate field as discussed in Discord. Will improve this in the next schema upgrade (post Mainnet): https://github.com/messari/subgraphs/issues/389.

this-username-is-taken:
Nice update!

PR #690, opened 2022-07-29
[Gov] Aave governance V2

danielkhoo:
## Description 

This PR includes the initial subgraphs for Aave Governance V2:

- Aave Governance
https://thegraph.com/hosted-service/subgraph/messari/aave-governance

image
image

- dYdX Governance 
https://thegraph.com/hosted-service/subgraph/messari/dydx-governance
image
image

## Difference from OpenZeppelinGovernor
The only schema difference is that the `Proposal` entity need an extra `executor` field
```graphql
type Proposal @entity {
  ...
  "Address of the proposal's executor contract"
  executor: String!
  ...
}
```


subgraphs/aave-governance/protocols/aave-governance/src/aave-governance-v-2.ts
```diff
@@ -0,0 +1,142 @@
...
+
+// Helper function that imports and binds the contract
+function getGovernanceFramework(contractAddress: string): GovernanceFramework {
+  let governanceFramework = GovernanceFramework.load(contractAddress);
+
+  if (!governanceFramework) {
+    governanceFramework = new GovernanceFramework(contractAddress);
+    let contract = AaveGovernanceV2.bind(Address.fromString(contractAddress));
+    governanceFramework.name = contract.NAME();
+    governanceFramework.type = GOVERNANCE_NAME;
```
mmnavarr, line 108: The `type` in the framework is different than our previous OZ subgraphs

We should have an OZ-specific const e.g. `OZ_FRAMEWORK_TYPE` but then use a new `AAVE_V2_FRAMEWORK_TYPE` (or something similar, not strongly opinionated on name) for here

subgraphs/aave-governance/protocols/aave-governance/src/aave-governance-v-2.ts
```diff
@@ -0,0 +1,142 @@
...
+    governanceFramework.version = "1";
+
+    governanceFramework.contractAddress = contractAddress;
+    governanceFramework.tokenAddress =
+      "0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9";
```
mmnavarr, line 113: Feels dirt hard-coding these in here. Wish the governance contract contained this information -_-

subgraphs/aave-governance/protocols/aave-governance/src/aave-governance-v-2.ts
mmnavarr: ooo I see, Aave governance has its own constant file

The governance name should ideally be unique to each governance entity (Aave in this case) and the framework type const shared for all framework entities using said framework.

This is something I see we may be doing wrong in the OZ implementations as well

governance_name != governance_framework_type

Something worth discussing offline

subgraphs/aave-governance/protocols/dydx-governance/src/dydx-governor.ts
```diff
@@ -0,0 +1,142 @@
...
+// Helper function that imports and binds the contract
+function getGovernanceFramework(contractAddress: string): GovernanceFramework {
+  let governanceFramework = GovernanceFramework.load(contractAddress);
+
+  if (!governanceFramework) {
+    governanceFramework = new GovernanceFramework(contractAddress);
+    let contract = DydxGovernor.bind(Address.fromString(contractAddress));
+    governanceFramework.name = contract.EIP712_DOMAIN_NAME();
+    governanceFramework.type = GOVERNANCE_NAME;
+    governanceFramework.version = "1";
```
mmnavarr, line 109: Does Aave governance have a concept of version?

mmnavarr: I don't see it in the contract which may make "1" a little misleading

"N/A" would probably be a little more accurate. Alternatively, could make version optional but that would be breaking change against OZ schema

danielkhoo: It doesnt use versions. I agree `N/A` is good default for unused entity fields

subgraphs/aave-governance/protocols/aave-governance/src/aave-governance-v-2.ts
danielkhoo: Isn't that exactly what we're doing here?
 governanceFramework.name is the name from the contract i.e. `Aave Governance v2` or `dYdX Governance`
 governanceFramework.type is our constant name for the type `AaveGovernanceV2`
 
 I think only confusion is the variable name `GOVERNANCE_NAME` should be `GOVERNANCE_TYPE`

subgraphs/aave-governance/protocols/aave-governance/src/aave-governance-v-2.ts
danielkhoo: Yeah its annoying that the contract doesn't include it. OTOH this `getGovernanceFramework` function is already serving as the adapter for minor contract idiosyncrasies so its not the worse place for hardcoding values.

mmnavarr (dismissed):

subgraphs/aave-governance/protocols/aave-governance/config/templates/aave-governance.template.yaml
```diff
@@ -0,0 +1,102 @@
...
+        - event: DelegateChanged(indexed address,indexed address,uint8)
+          handler: handleDelegateChanged
+        - event: DelegatedPowerChanged(indexed address,uint256,uint8)
+          handler: handleDelegatedPowerChanged
+        - event: Transfer(indexed address,indexed address,uint256)
+          handler: handleTransfer
+      file: ./protocols/aave-governance/src/aave-token-v-2.ts
+  - kind: ethereum/contract
+    name: StakedTokenV2Rev3
+    network:  {{ network }}
```
mmnavarr, line 77: Extra space

mmnavarr (approved):

mmnavarr (approved):
LGTM

PR #695, opened 2022-07-31
fix: total pool count

aditya-todi:
usageMetricsDailySnapshots: totalPoolCount.

Fixed: Total Pool Count is now updated for daily usage snapshots.
There are 26 pools in the protocol, including 3 which are now shifted to withdraw pools. Withdraw pool’s TVL is not counted towards protocol’s TVL.


this-username-is-taken:
LGTM!

PR #714, opened 2022-08-03
refactor(#patch); network; clean up old code

dmelotik:
Changes:
- Change `STATS` to `Stats`
- remove old naming and change to `Network`
- Update Schema to `1.1.1` > removing `Stats.median` in hopes of improving performance

Testing: https://okgraph.xyz/?q=dmelotik%2Fnetwork-mainnet


this-username-is-taken:
Don't forget to increment subgraph version

this-username-is-taken:
Another thing I noticed is that the current version syncs very slowly (I think the ethereum one did 10% in 12 days or something like that). You think there is any way we can speed things up?

dmelotik:
> Another thing I noticed is that the current version syncs very slowly (I think the ethereum one did 10% in 12 days or something like that). You think there is any way we can speed things up?

This seems much slower than the first version. That synced up in around 3-4 weeks. This is way slower. And I did change anything with handlers or contract calls.

The slowest part is the array sorting https://github.com/messari/subgraphs/blob/master/subgraphs/network/src/helper.ts#L403 which happens on each block 10x...

It needs to happen each block so that `median` can be grabbed. Thoughts on saving this operation until a `dailySnapshot` is full so it happens once per ethereum day? I think that is an okay tradeoff, not sure how important current day median is.

Another thought.. Since the array is already sorted and we are just adding an element to the end to sort wouldn't it be pretty fast?

this-username-is-taken:
Honestly I'm fine with removing median if it takes so much time to compute

dmelotik:
@this-username-is-taken sounds good, I will remove and hopefully that helps

dmelotik:
@this-username-is-taken still pretty slow. I think it may have to do with adding data to 12 different stats entities on each block...

this-username-is-taken:
Hmm, let me merge it in and see how things are

PR #715, opened 2022-08-03
docs(); contributing; add contributing guidelines

dmelotik:


docs/Contributing.md
```diff
@@ -0,0 +1,84 @@
...
+git pull --rebase origin master
+```
+
+This will initiate a rebase. Sometimes you will have to resolve conflicts. And then `git add .` and `git rebase --continue` or `git rebase --skip`.
+
+Following this method, after rebasing you can do the command `git push --force` to update your remote repo.
+
+## Naming Convention
+
+### Pull Requests
```
jaimehgb, line 52: I would suggest adding a comment encouraging to keep PRs small. When they are smaller they are easier to review. If they get too big it increases the chances of the reviewer losing focus/getting tired while reviewing all the files.
Might delay the review process, or worse, get an approval without being properly reviewed.

In case a PR feels too big it might be because of it doing too many things. Try to:
- keep formatting fixes alone, with no other code changes
- isolate bug fixing in their own PRs without mixing them with feature additions
- single PR per feature
- etc

docs/Contributing.md
```diff
@@ -0,0 +1,84 @@
...
+
+> See "Naming Conventions" for how to name your PR
+
+4. Once your PR is ready for review go ahead and click "Ready for review". As a courtesy it is nice to let the reviewer know that it is ready through a DM.
+
+At this point you will just need to follow the "Reviewing" process until your PR is merged.
+
+## Reviewing
+
+### Reviewee
```
jaimehgb, line 23: How should the reviewer be chosen?

docs/Contributing.md
```diff
@@ -0,0 +1,84 @@
...
+- Do not make a new PR for changes, this will make it harder for the reviewer to track their progress.
+
+### Reviewer
+
+- This part is straight forward, and if you are reviewing you know what to do.
+- It is courtesy to let the developer know that you are done with the review.
+
+> This is an iterative process that takes time, so don't think it will always be easy.
+
+## Rebasing
```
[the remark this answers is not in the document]
jaimehgb: It's definitely easier, but I would suggest rebasing when possible.

Merges leave ugly "scar" commits with descriptions like:

`Merge branch 'master' of github.com:messari/subgraphs`

Also, such commits make understanding git history more difficult to follow. Tools like git blame, git log, and git bisect become much more cumbersome, since some changes will appear in the historical tree more than once.

docs/Contributing.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Subgraphs: change impacts an entire fork directory / multiple protocols
+- Dashboard: change impacts the entire dashboard / multiple objects
+
+### 2️⃣
+
+- Subgraphs: change impacts a single protocol
+- Dashboard: change impacts a single object
+
+### 3️⃣
```
jaimehgb, line 70: I like the idea of the impact level identifier. How do you feel about making them more descriptive? 

For example, using semver wording instead of numbers? (#patch, #minor, #major)
Or something along the lines?

Imagine a change which has no impact on code logic but affects many files from different protocols? Like formatting all files after adding some linter o something? Would it be a `1` or a `3`? It could be a 1 because it affects a lage portion of the repo, but also a 3 because no functionality was changed. I feel that by using words we can make it more explicit.

this-username-is-taken:
There are also some conventions for open source projects that we can consider/incorporate:

https://www.conventionalcommits.org/en/v1.0.0-beta.2/
https://github.com/googleapis/release-please

docs/Contributing.md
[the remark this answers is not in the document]
MichaelC1999: Thats a good point. Ive always done it that way but for the sake of keeping things clean ill start rebasing

docs/Contributing.md
dmelotik: good points! I added them

docs/Contributing.md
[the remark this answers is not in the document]
dmelotik: @MichaelC1999 gitLevel++;

docs/Contributing.md
dmelotik: @jaimehgb what do you think about keeping the numbers but the meanings correspond to semver wording. So 1 = #major 2 = #minor and 3 = #patch ? The idea would be to keep the PR names shorter

dmelotik: or we could use words like `feat` `fix` `docs` seen here https://www.conventionalcommits.org/en/v1.0.0-beta.2/#summary

docs/Contributing.md
```diff
@@ -0,0 +1,84 @@
...
+
+- Subgraphs: change impacts a single network of a single protocol
+- Dashboard: change impacts is a minor enhancement / bug fix
+
+### How to name a PR
+
+"`impact-level` `subgraph-name`: `description`"
```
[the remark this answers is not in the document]
dmelotik: The only thing with this is I can see it getting messy. Sometimes changes in the forks that affect multiple subgraphs can result in multiple version changes that are not all the same... @MichaelC1999 thoughts on this?

docs/Contributing.md
dmelotik: made a comment, right now there isn't a standard. usually vincent assigns it and it's someone who has previously worked on the code or something similar

docs/Contributing.md
jaimehgb: If the concern is PR name lenght we can use github labels instead 😅 

Apart from giving it a clearer meaning, part of the reason I personally prefer words over numbers is to avoid having to mentally map:
>2 -> oh, this was minor, right?
>or was it?
It might sound silly but 🤷‍♂️ 

And if we start adding other levels, like they mention about `other` in the link you share it might get messy.

I'm not so strongly opinionated on the wording we use though. Using conventionalcommits' schema sounds good to me too :D

dmelotik: @jaimehgb i didn't go with the comments I made lol. lmk your thoughts on the current change though

jaimehgb: lmao

It looks good to me, pretty descriptive 🚀 
Do you want to keep semver together with the rest of names? Since it can be inferred from each

dmelotik: @jaimehgb what do you mean by keep it together? Sorry I don't quite understand what you mean

jaimehgb: Oh, re-reading myself I didn't explain very well.

I was asking if you wanted to keep both semver (major, minor, patch) and the rest of identifiers (fix, feat, docs, etc...) for PR naming. Since you can infer semver from the others it might be redundant:
- fix -> patch
- feat -> minor
- docs -> _none_
- ...

Althugh now that I think about it, thinking out loud, some of them might not be so direct (chore, style, perf). So disregard my question 😅

docs/Contributing.md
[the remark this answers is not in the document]
MichaelC1999: That makes sense. For the validation dashboard Ive usually just named the branch/pr after the version and have the commits list off what was changed. I could add the dashboard version to the description instead.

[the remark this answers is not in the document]
dmelotik: I think it would still be fair to add that to the description of the PR

docs/Contributing.md
dmelotik: okay lol, I like the descriptiveness of the schema i came up with but I don't like the aesthetic, but I think I can look past that lol

jaimehgb (approved):

this-username-is-taken:
Thanks for putting this together!

PR #716, opened 2022-08-03
Uniswap-forks - Updated token weights and removed some imports

steegecs:


this-username-is-taken:
You should probably update the subgraph version as well

steegecs:
> You should probably update the subgraph version as well

Updated!

this-username-is-taken:
Thanks!

Issue #612, opened 2022-07-18
Sushiswap (All: Arbitrum, Fuse, Polygon, Avalanche, BSC, Celo, Fantom, Gnosis, Mainnet, Moonriver, moonbeam) Schema Version 1.3.0 Subgraph Version  1.1.2 Methodology Version 1.0.0


ishraq8:
Overall Comments: 1) All subgraphs have issues for `Base Yield` on pool overview page based on sanity check. 2) Avalanche, Celo and Gnosis have issues of `totalvaluelockedUSD` and `dailyvolumeUSD` when compared to Defilama and Sushi analytics dashboard. 3) Fantom, Celo, Avalanche and Fuse have issues with `dailyvolumebytokenUSD` charts showing zero inside the pool section. 4. Ethereum, BSC and Polygon are still indexing (blocks and percentage indicated in their section). 

Excel file for all issues combined: https://docs.google.com/spreadsheets/d/1-z6Eyl0cKodxq7fmnhS7Pt_F56ci6qW-zBi-KkPvugw/edit#gid=0

**Arbitrum**

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/sushiswap-arbitrum&tab=protocol
-- | --
Date Reviewed | July 15 2022
Schema Version | 1.3.0
Subgraph Version | 1.1.2
Methodology Version | 1.0.0
Evidence | https://docs.google.com/spreadsheets/d/1CJzTV7LA2BsH1iMpY5x7Y3diTCbP9HbxvW-WJrdE7a0/edit#gid=617964580

Pool Overview
All Pools |   |   |  
-- | -- | -- | --
All | Base Yield % | NOT OK | Yield is showing APY in the 4 digits. This is excluding liquidity mining. Sanity check suggests something is wrong. Could it be pulling farm data?

**Fuse**

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/sushiswap-fuse&tab=protocol
-- | --
Date Reviewed | July 15 2022
Schema Version | 1.3.0
Subgraph Version | 1.1.2
Methodology Version | 1.0.0
Evidence | https://docs.google.com/spreadsheets/d/1lch_ID4FZMOQLGEtQf0fwju2RgB-yVXWCHr6WJwxAJQ/edit?usp=drive_web&ouid=111561086681119932380

Pool Overview
[...]

PR #728, opened 2022-08-04
Sushiswap Gnosis - Fix issue #612

steegecs:
Updated the minimum liquidity for pricing threshold and update the reference token for Sushiswap xDAI


this-username-is-taken:
Nice catch!

this-username-is-taken:
Next time you should make sure there is a link to the issue in either the description or the comment. Like this: https://github.com/messari/subgraphs/issues/612

steegecs:
Is it supposed to just get highlighted in blue when I put the hash in front? It does want to do it automatically in the title, but it does in comments.

PR #730, opened 2022-08-04
fix: uniswap-forks; pool snapshots cumulative values

jaimehgb:
This PR aims to fix 2 issues:

### 1
Hourly and daily cumulative revenue values are only updated on Swap events, but snapshots might be created too on Mint/Burn.

It might happen that at during hour/day there are no swaps, but there are lp mints or burns. If this happens, we initialize a snapshot with a 0 cumulativeRevenue value (it would be updated on swap). Since there are no swaps during that hour it stays at 0.

### 2

We seem to be using protocol cumulative metrics for pool snaphots? I would assume we want pool cumulative metrics instead, since protocol cumulatives are tracked on the financials snapshots.


steegecs:
This looks right and good to me!

this-username-is-taken:
@jaimehgb  Appreciate giving another set of eyes on this and finding all the issues!

PR #731, opened 2022-08-04
Pre-commit hook to not add unstaged files

jaimehgb:


steegecs (commented):
Looks all good to me!

PR #735, opened 2022-08-05
OpenSea v2 Subgraph (Wyvern Exchange)

mrbrianhobo:
TODO items:
- handle bundle sales (and marketplace revenue/royalties in cases of bundle sales)
- calculate final match price for dutch auction sales


0xbe1:
@mrbrianhobo there are a few TODOs left - do you want me to review now, or later when TODOs addressed?

0xbe1 (dismissed):
LGTM! Good to learn that the call contains all data we need

0xbe1:
also, would be great to add README explaining the high level idea, and why

subgraphs/opensea-v2/README.md
```diff
@@ -0,0 +1,240 @@
...
+    | Function Selector | Text Signature |
+    | --- | --- |
+    | `0x23b872dd` | `transferFrom(address,address,uint256)` |
+    | `0x42842e0e` | `safeTransferFrom(address,address,uint256)` |
+    | `0xf242432a` | `safeTransferFrom(address,address,uint256,uint256,bytes)` |
+    | `0xfb16a595` | `matchERC721UsingCriteria(address,address,address,uint256,bytes32,bytes32[])` |
+    | `0xc5a0236e` | `matchERC721WithSafeTransferUsingCriteria(address,address,address,uint256,bytes32,bytes32[])` |
+    | `0x96809f90` | `matchERC1155UsingCriteria(address,address,address,uint256,uint256,bytes32,bytes32[])` |
+
+    Additional function selectors can be found using the [Ethereum Signature Database](https://www.4byte.directory/).
```
this-username-is-taken, line 195: oh wow, this is cool!

subgraphs/opensea-v2/src/mappings.ts
```diff
@@ -0,0 +1,480 @@
...
+  getSaleStrategy,
+} from "./helpers";
+import {
+  checkCallDataFunctionSelector,
+  decodeSingleNftData,
+  getFunctionSelector,
+  guardedArrayReplace,
+} from "./utils";
+
+// struct Order {
```
this-username-is-taken, line 30: Maybe also add a link here so people can find source: https://github.com/ProjectWyvern/wyvern-ethereum/blob/bfca101b2407e4938398fccd8d1c485394db7e01/contracts/exchange/ExchangeCore.sol#L92

subgraphs/opensea-v2/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+  replacement.reverse();
+  mask.reverse();
+
+  let bigIntArray = BigInt.fromUnsignedBytes(array);
+  let bigIntReplacement = BigInt.fromUnsignedBytes(replacement);
+  let bigIntMask = BigInt.fromUnsignedBytes(mask);
+
+  bigIntReplacement = bigIntReplacement.bitAnd(bigIntMask);
+  bigIntArray = bigIntArray.bitOr(bigIntReplacement);
+  // return changetype<Bytes>(Bytes.fromBigInt(bigIntArray).reverse());
```
this-username-is-taken, line 74: remove?

subgraphs/opensea-v2/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+export function checkCallDataFunctionSelector(callData: Bytes): boolean {
+  let functionSelector = getFunctionSelector(callData);
+  return (
+    functionSelector == MATCH_ERC721_TRANSFER_FROM_SELCTOR ||
+    functionSelector == MATCH_ERC721_SAFE_TRANSFER_FROM_SELCTOR ||
+    functionSelector == MATCH_ERC115_SAFE_TRANSFER_FROM_SELCTOR
+  );
+}
+
+export function guardedArrayReplace(
```
this-username-is-taken, line 48: Should add a comment with a link to https://github.com/ProjectWyvern/wyvern-ethereum/blob/bfca101b2407e4938398fccd8d1c485394db7e01/contracts/exchange/Exchange.sol#L20

subgraphs/opensea-v2/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+export function decode_matchERC721UsingCriteria_Method(
+  callData: Bytes
+): DecodedTransferResult {
+  let functionSelector = getFunctionSelector(callData);
+  let dataWithoutFunctionSelector = Bytes.fromUint8Array(callData.subarray(4));
+  let dataWithoutFunctionSelectorWithPrefix = ETHABI_DECODE_PREFIX.concat(
+    dataWithoutFunctionSelector
+  );
+
+  let decoded = ethereum
```
this-username-is-taken, line 87: nice!

subgraphs/opensea-v2/src/mappings.ts
```diff
@@ -0,0 +1,480 @@
...
+  let priceETH = volumeETH;
+
+  // buyer is buyOrder.maker (addrs[1])
+  let buyer = call.inputs.addrs[1].toHexString();
+  // seller is sellOrder.maker (addrs[8])
+  let seller = call.inputs.addrs[8].toHexString();
+
+  // No event log index since this is a contract call
+  let tradeID = call.transaction.hash.toHexString();
+  let trade = new Trade(tradeID);
```
this-username-is-taken, line 178: what if multiple trades in a single tx?

this-username-is-taken:
not sure if you ran into this, but seems helpful for debugging: https://standardcrypto.retool.com/embedded/public/12c77f62-cd06-499a-9ef5-c6bec457443c
Found here: https://mirror.xyz/blog.standard-crypto.eth/OLhItPXcXD7sNX7pX3dflnWF6FW7Bi77KiquWxwZCNI

this-username-is-taken:
Overall great work! The documentation is definitely very helpful. Probably worth giving a presentation sometimes going over decoding call data. I think people will find that useful.

One thing probably want to look into is the revenue data, seems like a bug:
image

subgraphs/opensea-v2/src/mappings.ts
mrbrianhobo: i think single sales only contain 1 trade in the transaction, but i would need to handle this case for bundle sales 
note: since this uses call handlers/there is no event log index

mrbrianhobo: changed to `{transactionHash}-{methodId}-{tokenId}`

this-username-is-taken: ah right. This is call handler so it's tricky.

What I meant is someone can group multiple function calls (i.e. multiple calls to atomicMatch) in the same tx. Or is that considered bundle sale?

Issue #707, opened 2022-08-01
Venus Protocol QA (BSC) Schema Version 1.3.0 Subgraph Version 1.0.6 Methodology Version	1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/venus-protocol-bsc&tab=protocol
Date Reviewed | August 1 2022
Schema Version | 1.3.0
Subgraph Version | 1.0.6
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1uRTeDj5l9ItwThie-6rbI1-g5uYouIivueY3urS6LGI/edit#gid=617964580

### Metrics To Review
**Comments** - Revenue values look correct based on the borrow and liquidation metrics, but those metrics (daily borrow and liquidate) look off for certain dates. So did not make any comments on the revenue and will recheck it once borrows and liquidate is addressed (I am guessing the current implementation for revenue is correct though).

**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | dailyBorrowUSD | All time high at 80 billion daily borrow on 04/21/2021 seems too high and most likely needs a double check, as Compound's highest dailyBorrowUSD value is roughly 3b.
financialsDailySnapshots | cumulativeBorrowUSD | If daily off, cumulative also off. Cumulative sum correct from a mathematical perspective
financialsDailySnapshots | dailyDepositUSD | Spike on 04/21/2021 of 200b in deposits. This seems too high and needs second look. Compound's highest deposit daily value was around 4.5b for comparison.
financialsDailySnapshots | cumulativeDepositUSD | Daily is off, also cumulative is off
financialsDailySnapshots | dailyLiquidateUSD | Spike on 05/19/2021 of $287,917,021.674 in liquidations. This seems too high and needs second look. Comparatively, Compound's daily liquidation on this date was under 150b.
financialsDailySnapshots | cumulativeLiquidateUSD | If daily is off, this is off

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
Venus CAKE | Naming | Nuance thing, but Cake should be CAKE

this-username-is-taken:
Really high deposit/borrow is reasonable on a low fee chain (similar to BENQI) since people can "wash"-deposit/borrow. Also bots might do this for arbs/mev. In any case, we should check to confirm.

[...]

PR #741, opened 2022-08-05
fix(#patch); venus; address venus qa

dmelotik:
Fixing #707 

Changes:
- Force `Cake` to have the symbol `CAKE`
- Update venus to version `1.1.1`


dmelotik:
moving to #749

PR #746, opened 2022-08-08
add seaport

0xbe1:
deployment: https://okgraph.xyz/?q=messari%2Fopensea-seaport-ethereum


subgraphs/seaport/schema.graphql
```diff
@@ -0,0 +1,272 @@
...
+  marketplaceRevenueETH: BigDecimal!
+
+  " Revenue that goes to creator, aka royalty fee. "
+  creatorRevenueETH: BigDecimal!
+
+  " Sum of marketplaceRevenueETH and creatorRevenueETH. "
+  totalRevenueETH: BigDecimal!
+
+  " Cumulative unique traders "
+  cumulativeUniqueTraders: Int!
```
this-username-is-taken, line 92: I think we should add `collections: [Collection!]! derivedFrom marketplace` and `snapshots: [MarketplaceDailySnapshot!]! derivedFrom marketplace` here.

subgraphs/seaport/schema.graphql
```diff
@@ -0,0 +1,272 @@
...
+  tradeCount: Int!
+
+  " Buyer count. "
+  buyerCount: Int!
+
+  " Seller count. "
+  sellerCount: Int!
+
+  " Trades of the collection. "
+  trades: [Trade!]! @derivedFrom(field: "collection")
```
this-username-is-taken, line 136: Also `snapshots: [CollectionDailySnapshot!]! derivedFrom collection` here

this-username-is-taken: For all NFT marketplace subgraphs

0xbe1:
@this-username-is-taken let me add these fields for all nft marketplace in one PR

subgraphs/seaport/src/mapping.ts
```diff
@@ -0,0 +1,691 @@
...
+  ) {}
+}
+
+export function handleOrderFulfilled(event: OrderFulfilled): void {
+  let offerer = event.params.offerer;
+  let recipient = event.params.recipient;
+  let offer = event.params.offer;
+  let consideration = event.params.consideration;
+
+  // I have a mixed feeling about OrderFulfilled event
```
mrbrianhobo, line 77: If not the `OrderFulfilled` event, what other events or information is available?

subgraphs/seaport/src/mapping.ts
```diff
@@ -0,0 +1,691 @@
...
+  // On the up side, it contains literally all data we need to build up a Trade entity
+  // One the down side, it is very generic and it requires a lot of hidden knowledge to decipher its actual meaning
+  // 1. `offer` doesn't tell you explicitly that if it is a bid offer or an ask offer
```
mrbrianhobo, line 80: In the `Trade` entity currently, it doesn't know the difference between a bid or ask right? As long as theres a buyer (NFT recipient) and seller (NFT seller), we can properly create a trade.

Or is `OrderFulfilled.recipient` not necessarily the NFT recipient which is why knowing the difference between bid/ask important?

subgraphs/seaport/src/mapping.ts
```diff
@@ -0,0 +1,691 @@
...
+  } else {
+    if (isERC1155Result.value) {
+      return NftStandard.ERC1155;
+    }
+  }
+
+  return NftStandard.UNKNOWN;
+}
+
+// Use best effort to figure out the following data that construct a `Sale`:
```
mrbrianhobo, line 442: 👍 doing the best with what we have

I'm wondering does would Dune queries for Seaport be helpful for navigating through buyer/seller or fees (I think these are the more ambiguous fields to resolve)? 
ref: https://dune.com/Marcov/opensea-seaport

mrbrianhobo: not sure exactly how they distinguish between types of fees or between buyer/seller

mrbrianhobo (commented):
Looks really good to me 👍 

Are there a list of edge cases and transactions for those cases you have on hand? Maybe it might be worth documenting so they can be found easily?

subgraphs/seaport/src/mapping.ts
mrbrianhobo: Is there an alternative you are considering?

0xbe1: unfortunately there's no alternative that i am aware of

subgraphs/seaport/src/mapping.ts
0xbe1: > is OrderFulfilled.recipient not necessarily the NFT recipient which is why knowing the difference between bid/ask important?

yeah, this is what I find

subgraphs/seaport/src/mapping.ts
0xbe1: yeah tbh i have no clue what they do under the hood

https://dune.com/queries/911780/1594223

this table `seaport."view_transactions"` has both `buyer` and `seller` column, Dune did some thing to figure them out before giving the data to their users

0xbe1:
> Looks really good to me 👍
> 
> Are there a list of edge cases and transactions for those cases you have on hand? Maybe it might be worth documenting so they can be found easily?

check out comments of method `handleOrderFulfilled` - I have tried my best to put almost an essay there 😆

PR #754, opened 2022-08-08
fix(#minor); uniswap forks; updated reference tokens and stable pools for Sushiswap BSC and Polygon - Issue #612

steegecs:
Issue #612 

- The final things left from this issue are with Sushiswap BSC and Polygon. 
- The pricing was way off because the reference token was getting priced poorly on both. 
- Updated reference token and or stable oracle pools to improve pricing. 
- Polygon was using the wrong oracle pools, and BSC needed to use BNB as its reference token since Eth was low volume.


this-username-is-taken:
Nice digging!

PR #756, opened 2022-08-08
fix: husky pre-commit hook

jaimehgb:
It seems we didn't need to get too smart with the pre-commit hook running the prettier.

`npx pretty-quick --staged` is smart enough to know which of the staged files it modifies, and stages them again after prettying.


steegecs:
:) Yeah this works for me.

PR #755, opened 2022-08-08
fix(#minor); uniswap forks; Set rewards related fields to not populate unless staking is enabled.

steegecs:
**Context**
rewardTokenEmissionAmount, rewardTokenEmissionsUSD, and stakedOutputTokenAmount are optional fields in the DEX schema. Currently, these fields are being returned with values, regardless of whether or not the LP tokens minted from these pools can be used to earn rewards.

**Problem**
We want these fields to be equal to null when these LP tokens cannot be used to earn rewards. 

**Proposal**
Removed the instantiation of these fields at the time of creation of the liquidity pool entity. Then, these fields will only be populated if there is an analogue MasterChef pool for the LP token.

Issue #740, opened 2022-08-05
Dex AMM undefined or empty fields 


liambeckman199:
While querying a number of Dex AMM subgraphs on various chains, I have noticed the following fields in particular either undefined or empty:

**- RewardemissionsUSD 
- RewardTokenEmissionsAmount
- Protocol Name and Network
- outputToken** 

I'm assuming this is a pending out of sync issue, but it would helpful to know if the team has noticed issues with any of these fields and has plans on addressing them. A subgraph that showcases these issues is Curve Finance on Ethereum Mainnet.

this-username-is-taken:
Thanks for flagging. Tagging @harsh9200 for Curve. @liambeckman199 can you list the other subgraphs where you are seeing this issue?

steegecs:
Here is a little report from looking through all the dex-amms besides curve! 

By always present, always null, or sometimes present and sometimes null in the report below, I mean that it is actually a null field and not just a '0' value. I did not notice any cases where the protocol name or network are missing, but there were a couple examples of outdated names. 

There is a fair bit of discrepancy between the protocols in how they handle the stakedOutputTokenAmount, rewardTokenEmissionsUSD, and rewardTokensEmissionsAmount. I think this is because, sometimes people chose to not add a value to these fields unless that pool is actually receiving rewards. 

Also, I saw that the outputToken for Balancer and Beethoven is always null. 

*Bancor V3
- rewardTokenEmissionUSD: always present
- rewardTokenEmissionsAmount: always present
- stakedOutputToken: always present
- outputToken: always present
- ProtocolName: present
- ProtocolNetwork: present but uses mainnet instead of ethereum

*Balancer V2 (ALL)
- rewardTokenEmissionUSD: always null
- rewardTokenEmissionsAmount: always null
- stakedOutputTokenAmount: always present
- outputToken: always null
- ProtocolName: present
- ProtocolNetwork: present but uses matic instead of polygon for the polygon deployment.

*Beethoven X Optimism
- rewardTokenEmissionUSD: always null
- rewardTokenEmissionsAmount: always null
- stakedOutputTokenAmount: always present
- outputToken: always null
- ProtocolName: present
- ProtocolNetwork: present

*Saddle Finance (ALL)
- rewardTokenEmissionUSD: sometimes present sometimes null
- rewardTokenEmissionsAmount: sometimes present sometimes null
- stakedOutputTokenAmount: sometimes present sometimes null
- outputToken: always present
- ProtocolName: present
[...]

PR #762, opened 2022-08-09
docs(); schema; updated Reward Token information for DEXs

steegecs:
- Added more in-depth information to the schema documentation about fields related to rewards calculations for dexes
- We want to keep fields like (`rewardTokenEmissionsAmount`, `rewardTokenEmissionsUSD`, `stakedOutputTokenAmount`, `rewardToken`) null unless the related pool is used in staking. 
- Even if the pool is no longer earning rewards, we still want to leave these fields as non-null values.
- Response to issue #740. Related PR #755


jaimehgb (approved):

this-username-is-taken:
Thanks for the update!

PR #763, opened 2022-08-09
feat(#major); CI/CD; major refactor of code and added tests for deployment.json and CI/CD. 

steegecs:
**Context**
This PR comes with an upgrade to the CI/CD scripts plus a set of tests to go along with it in response to deployment.json upgrade in PR #979.

**Changes**
- [x] Major refactor of CI/CD deployment scripts. 
- [x] Added capability to deploy to subgraph-studio and cronos-portal.
- [x] Much easier to build subgraphs. It now takes two arguments max for building!
- [x] You can specify an alternate deployment location for singular deployments. 
- [x] Removed deploy on merge functionality
- [X] Better argument names for CLI script. 
- [x] Added comprehensive tests for deployment.json that execute before each build and deployment. Tests whether all of the data is present or expected. https://github.com/steegecs/subgraphs/blob/deploy-json-new/deployment/test/json-test.js
- [x] Added a set of tests for command line scripts for deployment functionality. https://github.com/steegecs/subgraphs/blob/deploy-json-new/deployment/test/deployment-tests.txt
- [x] Added tests within deployment class during preparation to check for valid arguments. 
- [x] Update workflow for checking builds on push to Messari repo. 

README.md describing how to use new CI/CD scripts. Also, recommend looking through the deployment-tests.txt file.
https://github.com/steegecs/subgraphs/blob/deploy-json-new/deployment/README.md


this-username-is-taken:
would you also add a field called `status` with options either `prod` or `dev`?

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+
+  // This checks if the deployment json file is valid for the deployments that you want to execute.
+  isValidJsonData() {
+    let scope = this.getDeploymentScope();
+    if (scope == "network") {
+      this.checkProtocolLevelData(this.protocol);
+      if (this.type == "deploy" || this.type == "check") {
+        this.checkNetworkLevelData(this.protocol, this.network);
+      } else if (this.type == "build") {
+        this.checkNetworkLevelBuildDataOnly(this.protocol, this.network);
```
jaimehgb, line 288: This function looks like it can be simplified.

- You use this in every branch. Might move it to its own function.
- Replace the elses with early returns.
- Calls to `this.checkProtocolLevelData(this.protocol);` are in all branches but the forks one. You might try to generalize it too.

jaimehgb (commented):
It feels way cleaner this way. I like the approach 🚀 

I've left a few comments. Most of them regarding style and readability. I know I'm a bit picky. Many of these things will be caught earlier by a linter though, so I won't be this annoying for very long 😅

deployment/README.md
```diff
@@ -1,160 +1,41 @@
...
-    - `npm install minimist`
-  - mustache
-    - `npm install mustache`
+# Deploys uniswap-v2 to ethereum in steegecs hosted service.
 
-```
-# Deploys uniswap-v2 from the uniswap-forks to mainnet in my hosted service.
-npm run deploy --SUBGRAPH=uniswap-forks --PROTOCOL=uniswap-v2 --NETWORK=mainnet --LOCATION=steegecs
+npm run deploy --PROTOCOL=uniswap-v2 --NETWORK=ethereum --TARGET=steegecs --SERVICE=h --TYPE=deploy
+npm run deploy --PROTOCOL=uniswap-v2 --NETWORK=ethereum --TARGET=steegecs --SERVICE=h
```
dmelotik, line 6: omitting `--TYPE=deploy` will also deploy? If that is the case I could imagine that it could possible happen due to error. I think it would be better to make it explicit

deployment/README.md
```diff
@@ -1,160 +1,41 @@
...
-          - deploy-on-merge (Optional)
-            - true/false/null
-            - Set to true if you want to deploy upon merging your branch with remote repo
-            - Does not deploy unless true
-          - prepare:constants (Optional)
-            - true/false/null
-            - Set to true if you are configuring constants in the configurations/ folder.
-            - Does not use unless true
+- The CI/CD deployment scripts and actions allow you to deploy, build, or check the readyness for deployment of multiple subgraphs at a time.
+- A response telling you if any information is missing or misplaced in the deployment.json, or if you specified invalid parameters or combinations in the script will be given. Errors will occur if your file structure does not conform to the standard -- See uniswap-forks as an example
```
dmelotik, line 37: Good feature

deployment/deployment.js
```diff
@@ -1,240 +1,25 @@
...
-    for (const protocol in protocolNetworkMap[args.subgraph]) {
-      for (const network in protocolNetworkMap[args.subgraph][protocol]) {
-        let template =
-          protocolNetworkMap[args.subgraph][protocol][network]["template"];
-        let location = "";
-        let prepareConstants =
-          protocolNetworkMap[args.subgraph][protocol][network][
-            "prepare:constants"
-          ];
+  throw "Please check package.json scripts in local subgraph folder. This error is being thrown becuase it is missing a parameteter in the 'deploy' script. You can find an updated version of the scripts in the deployments folder at the head of the directory.";
```
dmelotik, line 17: thoughts on having some sort of logic to tell which item is missing (can be the first one it runs into) then notifies that? Maybe not worth the effort though

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+          protocol +
+          " --NETWORK=" +
+          network
+      );
+    }
+    scripts.push("graph codegen");
+
+    // We don't want to deploy if we are building or just testing.
+    if (this.type == "deploy" || this.type == "") {
+      scripts.push(this.getDeploymentScript(location));
```
dmelotik, line 97: okay so looks like this is the case. I would argue you have to explicitly state `deploy` to deploy

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+
+  // Checks if the input arguments are valid for the deployment location.
+  checkValidService() {
+    switch (this.service) {
+      case "subgraph-studio":
+      case "studio":
+      case "s":
+      case "decentralized-network":
+      case "d":
+        return true;
```
dmelotik, line 133: why have these extra `return true;` statements when you only need one at the end in this case?

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+    }
+  }
+
+  // Makes sure the version specified in the json file is valid.
+  checkValidVersion(version) {
+    this.checkVersionLengthIsTwo(version);
+    this.checkVersionIsNumber(version);
+  }
+
+  checkVersionLengthIsTwo(version) {
```
dmelotik, line 156: should probably read `checkVersionLengthIsThree()` based off @jaimehgb comment

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+    if (version.split(".").length - 1 != 2) {
+      throw "See deployment.json: (1) version is not valid - must be in format x.x.x (e.g. 1.3.1)";
```
dmelotik, line 158: Again being picky, but you could be more descriptive and say that the length is not three or something along those lines

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+  }
+
+  checkVersionIsNumber(version) {
+    let array = version.split(".");
+    if (
+      !array.every((element) => {
+        return !isNaN(element);
+      })
+    ) {
+      throw "See deployment.json: (2) version is not valid - must be in format x.x.x (e.g. 1.3.1)";
```
dmelotik, line 169: same here but with numbers, otherwise it is the same error message as above

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+    }
+    this.checkValidVersion(
+      this.data[protocol]["networks"][network][this.getServiceByAlias()][
+        "version"
+      ]
+    );
+  }
+
+  checkAuthorization() {
+    if (!this.access && this.getServiceByAlias() == "cronos-portal") {
```
dmelotik, line 260: is `args.access` something you have to specify in deployment.json? If so we wouldn't want the cronos access token there

dmelotik (dismissed):
👏 @steegecs !  Your code was really easy to read, sorry for being picky, but this is a pretty important piece of code lol

deployment/README.md
```diff
@@ -1,160 +1,41 @@
...
+- --TYPE=`build` --TYPE executes a clean build by first removing generated, build, subgraph.yaml, and configure.ts files and folders.
+- --TYPE=`deploy` command excutes the same step as `build`, but it goes all the way to deployment.
+- --TYPE=`check` command does everything `deploy` does except actually deploy the subgraph. What is different from `build` is that is runs all the same checks as `deploy`. Some do not apply to `build` - such as whether or not there is a valid deployment service in the deployment.json.
+  - This command will execute on each push to a remote branch in the Messari repository for relevant deployments that have been inpacted by the code changes.
```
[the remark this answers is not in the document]
steegecs: I was actually meaning to say that the `check` type will run on any push (except for master). I actually removed the functionality to deploy on merge (push to master).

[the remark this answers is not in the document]
steegecs: I'll make this a little more clear though.

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+  // Checks if you are wanting to deploy a single network, all networks for a protocol, or all protocols and networks for a fork
+  prepareScripts() {
+    let scope = this.getDeploymentScope();
+    if (scope == "network") {
+      this.generateScripts(
+        this.protocol,
+        this.network,
+        this.getTemplate(this.protocol, this.network)
+      );
+    } else if (scope == "protocol") {
```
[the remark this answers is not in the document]
steegecs: Yes - that does make sense! Manage the complexity budget :)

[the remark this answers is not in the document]
steegecs: Although, in this instance I am not returning anything?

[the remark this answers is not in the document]
steegecs: Are you thinking a switch would be better here or something?

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+      for (const protocol in forkProtocols) {
+        for (const network in this.data[protocol]["networks"]) {
+          this.generateScripts(
+            protocol,
+            network,
+            this.getTemplate(protocol, network)
+          );
+        }
+      }
+    }
```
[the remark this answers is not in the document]
steegecs: Screen Shot 2022-09-01 at 4 45 22 PM

Yes, there is already a check in the processing stage. (I realized I needed to add one more condition).

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+  checkValidVersion(version) {
+    this.checkVersionLengthIsTwo(version);
+    this.checkVersionIsNumber(version);
+  }
+
+  checkVersionLengthIsTwo(version) {
+    if (version.split(".").length - 1 != 2) {
+      throw "See deployment.json: (1) version is not valid - must be in format x.x.x (e.g. 1.3.1)";
+    }
+  }
```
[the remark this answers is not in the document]
steegecs: XD. Haha, yeah... that makes sense. Although, this is a point I wanted to discuss. Since we may not be including the version in the deployment.json, I might scrap the version checks.

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+
+  // Checks if the protocol level data in the deployment json file is present and/or valid.
+  checkProtocolLevelData(protocol) {
+    if (this.data[protocol] == undefined) {
+      throw (
+        "Check --PROTOCOL spelling as argument or in deployment.json: protocol= " +
+        protocol
+      );
+    } else if (Object.keys(this.data[protocol]) == []) {
```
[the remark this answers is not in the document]
steegecs: Sounds good! Will fix.

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+      throw (
+        "See deployment.json: No networks are defined for: protocol=" + protocol
+      );
+    }
+  }
+
+  // Checks if the network level data necessary to build this subgraph is present and/or valid.
+  checkNetworkLevelBuildDataOnly(protocol, network) {
+    if (this.data[protocol]["networks"][network] == undefined) {
```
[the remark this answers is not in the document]
steegecs: I just considered that in all likelihood this would fail, because people forgot to add the key in the deployment.json. However, I see that a blank string is a reasonable check as well, so I will add it.

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+    } else if (
+      this.data[protocol]["networks"][network]["files"]["template"] == undefined
+    ) {
+      throw (
+        "See deployment.json: template is missing for: protocol=" +
+        protocol +
+        " network=" +
+        network
+      );
+    }
```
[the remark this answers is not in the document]
steegecs: I have decimated all `elses`

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+        protocol +
+        " " +
+        network
+      );
+    }
+  }
+
+  // Checks if the network level data necessary to build and deploy this subgraph is present and/or valid.
+  checkNetworkLevelData(protocol, network) {
+    if (this.data[protocol]["networks"][network] == undefined) {
```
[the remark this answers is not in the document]
steegecs: Ah! Good call.

deployment/deploymentClass.js
```diff
@@ -0,0 +1,448 @@
...
+      for (const network in this.data[this.protocol]["networks"]) {
+        if (this.type == "deploy" || this.type == "check") {
+          this.checkNetworkLevelData(this.protocol, network);
+        } else if (this.type == "build") {
+          this.checkNetworkLevelBuildDataOnly(this.protocol, network);
+        }
+      }
+    } else if (scope == "fork") {
+      let forkProtocols = this.getAllForks();
+      if (forkProtocols == []) {
```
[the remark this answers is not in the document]
steegecs: Fixed!

deployment/deploymentClass.js
steegecs: Fixed!

deployment/deploymentClass.js
steegecs: What do you mean by generalizing on the last point? Protocol level data is still checked when the scope is 'forks'.  By generalizing, do you mean combining the protocol and network level data checks?

deployment/README.md
steegecs: Okay - I can see how that would be more safe. Its a trade off for convenience, but I like can change it.

deployment/deploymentClass.js
steegecs: Sure!

deployment/deploymentClass.js
steegecs: Sure!

deployment/README.md
```diff
@@ -1,160 +1,41 @@
 ## Deployment Instructions:
```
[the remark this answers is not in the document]
steegecs: Sure - added that to the readme.

deployment/deploymentClass.js
```diff
@@ -166,34 +170,35 @@ class Deployment {
...
     ) {
-      throw "See deployment.json: (2) version is not valid - must be in format x.x.x (e.g. 1.3.1)";
+      throw "See deployment.json: (2) version is not valid - make sure to use integers between the periods - must be in format x.x.x (e.g. 1.3.1)";
     }
   }
 
   // Checks if the protocol level data in the deployment json file is present and/or valid.
   checkProtocolLevelData(protocol) {
-    if (this.data[protocol] == undefined) {
+    if ([undefined, ""].includes(this.data[protocol])) {
```
jaimehgb, line 179: You can do `if (!this.data[protocol])`.

Both `undefined` and `""` evaluate to false.

deployment/deploymentClass.js
jaimehgb: Oh disregard the last point. Looking at it now it might not make much sense. Looks way better already, so not necessary either :D

deployment/deploymentClass.js
steegecs: Actually, when I have tried this. And after testing again. "" does not evaluate as false with this expression.

jaimehgb: How strange 🤔 
Being how JS is I believe anything 😅 

Theres this thing of using `!!` when you want to evaluate some non-boolean as a boolean. And `===` whenever you need to compare, to make sure you are always comparing what you intend to compare.

Let's leave it as it is then.

Screenshot 2022-09-02 at 17 55 44

this-username-is-taken:
> > would you also add a field called `status` with options either `prod` or `dev`?
> 
> @this-username-is-taken Do you think we should put this at the protocol level like it has been in the README or at the network level?

I think this is a good opportunity to move it to the network level so we have more granular control 👍

steegecs:
> > > would you also add a field called `status` with options either `prod` or `dev`?
> > 
> > 
> > @this-username-is-taken Do you think we should put this at the protocol level like it has been in the README or at the network level?
> 
> I think this is a good opportunity to move it to the network level so we have more granular control 👍

And I am thinking to leave this outside of the service instead of inside the particular service configs
Screen Shot 2022-09-02 at 12 06 47 PM
Screen Shot 2022-09-02 at 12 07 12 PM

jaimehgb (approved):

deployment/deployment.js
```diff
@@ -1,240 +1,28 @@
...
+const deploymentJsonData = require("./deployment.json");
 const args = require("minimist")(process.argv.slice(2));
 
-const protocolNetworkMap = JSON.parse(JSON.stringify(protocolNetworkData))[
-  "subgraphs"
-];
-
-let allScripts = new Map();
-let results = "RESULTS:\n";
+testDeploymentJson(deploymentJsonData);
```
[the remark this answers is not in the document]
steegecs: Updated that naming of these files and function!

deployment/test/json-test.js
```diff
@@ -0,0 +1,332 @@
...
+}
+
+function checkDuplicateIDs(deploymentJsonMap) {
+  let deploymentIDs = new Set();
+  let protocols = new Set();
+
+  // Check for duplicate deployment IDs
+  for (const [protocol, protocolData] of Object.entries(deploymentJsonMap)) {
+    protocols.add(protocol);
+    for (const [deployment, deploymentData] of Object.entries(
```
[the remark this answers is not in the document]
steegecs: I left this test in, since there is still some utility for duplicates across protocols. I will be changing it as we discussed to use a validation tool in the near future.

jaimehgb (approved):

PR #769, opened 2022-08-09
fix(#minor); Solarbeam; fixed masterchef rewards calculations

steegecs:
**Context** 
The contract that was being used to calculate staking rewards was incorrect. This was causing the staking and rewards to not be present in the liquidity pools. 

**Solution**
Added the new ABI and made the necessary changes to the configurations to use the SolarDistributorV2 Contract. Additionally, these staking pools may earn multiple rewards tokens, so I updated the indexing to account for this.


this-username-is-taken:
parenthesis but braces, not too much an issue but im ocd 😅

this-username-is-taken:
Fixes for https://github.com/messari/subgraphs/issues/547

this-username-is-taken:
Great work! Don't forget to link the issue next time.

Issue #659, opened 2022-07-22
Abracadabra Ethereum Extended


BroderickBonelli:
This subgraph (https://thegraph.com/hosted-service/subgraph/dmelotik/abracadabra-ethereum-extended) has issues w/ the borrow amounts as well. Some values are negative and there seems to be an issue w/ the decimals. Some balances are too large such as this yvUSDC balance.
<img width="1420" alt="Abracadabra Ethereum Extended  Borrows" src="https://user-images.githubusercontent.com/79765357/180504686-6371f012-d645-400f-aee0-44aa7c1f8815.png">

Issue #658, opened 2022-07-22
Abracadabra Money Avalanche Ext 


BroderickBonelli:
There appear to be errors with borrow balances. Under the Position field, many balances are negative where side = 'Borrower'.
<img width="1223" alt="Abracadabra Extended Schema Borrows" src="https://user-images.githubusercontent.com/79765357/180494267-64600a60-88fa-457e-9d87-034ba40677d5.png">

this-username-is-taken:
I believe this has been fixed in https://thegraph.com/hosted-service/subgraph/messari/abracadabra-money-avalanche?version=pending:

![image](https://user-images.githubusercontent.com/2354636/185533457-a2e2e16d-26c0-4bef-8858-4e8b3735d6fa.png)

Would you be able to confirm: @BroderickBonelli

BroderickBonelli:
Hey @this-username-is-taken, confirming this looks correct now. Thanks!

this-username-is-taken:
Great! Thanks

PR #775, opened 2022-08-10
fix(#patch); abracadabra; fix positional balances 

dmelotik:
Fix postional balances to ensure they are accurate and non negative.

Relating to #658 #659 

Need to rebase once QA is merged


dmelotik:
abandon and using #781 instead

PR #777, opened 2022-08-11
fix(#minor); Uniswap V3; response to Issue #789

steegecs:
**Context**
There is a basket of pools instantiated at the beginning of the Optimism deployment because of a re-genesis. The pools instantiated did not have the proper name convention with the swap fee tagged on the end. It would show up like "Uniswap V3 wETH/OP" instead of "Uniswap V3 wETH/OP 0.3%"

Additionally, there were issue with the tracking of the output tokens as NFTs. There were negative values for a few liquidity pools. This is because the code had been tracking the positions as closing and opening at the time of Deposits and Withdraws, however, this is not always the case as the positions can just be modified. Additionally, because some of these values were coming in negative, it was throwing off pricing for some liquidity pools, because pricing was not occurring if there were no open positions for the pool.

Finally, the cumulativeUniqueUsers had a higher value than on Dune Analytics. Upon further inspection, I was passing event.transaction.from and event.params.sender in the Mint and Burn events to get the EOAs that initiated the event. I discovered this may not always be reliable, because sometimes the transaction can be initiated from another contract. In the most recent commit, I utilize the owner parameter in the Mint and Burn events to get the right address reliably for the Deposit and Withdraw entities and also to accurately record usage metrics of EOAs.

**Solution**
Updated the instantiation of the pools created before regenesis to have the proper pool names. Also, I removed the tracking of the output tokens as NFTs. I think it is best to not track them now and then update them later when we include positions data. 

This PR is in response to Issue #789


subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -140,7 +140,7 @@ export function findUSDPricePerToken(
       let poolAmounts = getLiquidityPoolAmounts(poolAddress);
       let pool = getLiquidityPool(poolAddress);
 
-      if (pool.outputTokenSupply!.gt(BIGINT_ZERO)) {
+      if (pool.totalValueLockedUSD!.gt(BIGDECIMAL_ZERO)) {
```
jaimehgb, line 143: Is this change safe?
I understand this change is because we are no longer tracking the output token, so need to change the condition. But would it behave as expected if the pool consists of 2 non-whitelisted tokens? or the totalValueLockedUSD is 0 for some other reason (low liquidity, etc)?

jaimehgb (approved):
Looks good to me! Thanks for the description of the changes, I would have freaked out if I didn't know why the removal of output tokens :P

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: There are still checks for minimum liquidity for the paired token, so I still think it is a safe change. All this this condition was was a sanity check that there were any liquidity positions, but I think it was unnecessary, unless used for error logs. 

Actually on second thought though, checking tvl might not even be necessary since it still asks that the value of the paired token is greater than the minimum liquidity threshold.

jaimehgb (approved):

this-username-is-taken:
Great work @steegecs!

PR #778, opened 2022-08-11
fix(#minor); Uniswap V3; Removed unused fields from the liquidity pool entity

steegecs:
**Context** 
There are some fields related to rewards emissions, and the price of the output token which are not applicable to Uniswap V3. There are no rewards emission for Uniswap V3, and because the protocol mints NFTs that represent positions, we do not price the OutputTokens (NFTs).


this-username-is-taken:
Maybe this is a good opportunity to git-ignore the `subgraph.yaml`

PR #781, opened 2022-08-11
fix(#patch); abracadabra; fix positional balances

dmelotik:
Fix postional balances to ensure they are accurate and non negative. Upgrade to version `1.2.5`

Relating to https://github.com/messari/subgraphs/issues/658 https://github.com/messari/subgraphs/issues/659


this-username-is-taken:
Great work!

PR #782, opened 2022-08-11
fix(#minor); Uniswap V3; Removed unused fields from the liquidity pool entity - RE: PR 778

steegecs:
**Context**
There are some fields related to rewards emissions, and the price of the output token which are not applicable to Uniswap V3. There are no rewards emission for Uniswap V3, and because the protocol mints NFTs that represent positions, we do not price the OutputTokens (NFTs).

I am not sure why this happened, but when I force pushed in PR #778 I accidentally push no commits and then it just merged in. This is a redo of that PR.

### DIFF
--- a/subgraphs/uniswap-v3/.gitignore
+++ b/subgraphs/uniswap-v3/.gitignore
@@ -1,1 +1,3 @@
 .DS_Store
+subgraph.yaml
+configure.ts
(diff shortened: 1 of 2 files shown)


steegecs:
> 

Should I make the subgraph.yaml and configure.ts ignore global?

this-username-is-taken:
interesting, you are right

this-username-is-taken:
> > 
> 
> Should I make the subgraph.yaml and configure.ts ignore global?

let's keep them local for now and up to the subgraph owner to decide. I think there are still a number of subgraphs that don't generate `subgraph.yaml` using mustache

PR #786, opened 2022-08-12
chore(); subgraphs; add manifest to gitignore

dmelotik:
Added subgraph.yaml to main `.gitignore`


.gitignore
```diff
@@ -2,6 +2,7 @@ generated/
 build/
 Reference/
 .vscode/
+subgraph.yaml
```
jaimehgb, line 5: Do you want this to ignore all `subgraph.yaml` files in the project or just at the root directory?

If the latter, I think it should be `**/subgraph.yaml`.

If that's the idea, can it happen that some protocol contains a single subgraph.yaml file without it being templatized? If so, do we want to ignore that one too?

dmelotik: there would be no subgraph.yaml at the root directory?

There are 2 subgraphs that are not templateized yet, but could be. I just left the `subgraph.yaml` in there for those. we could "un-.gitignore" those `!subgraph.yaml` in their respective folders?

jaimehgb: Oh, never mind. I thought this way it was only going to ignore `subgraph.yaml` if placed at the root. My bad.

> There are 2 subgraphs that are not templateized yet, but could be. I just left the subgraph.yaml in there for those. we could "un-.gitignore" those !subgraph.yaml in their respective folders?

If they are already commited it will be fine. It's when we start a new one that we'll need to be explicit to not forget to commit those subgraph.yaml

deployment/deployment.json
```diff
@@ -7,35 +7,35 @@
           "steegecs": "steegecs/abracadabra-money-arbitrum",
           "messari": "messari/abracadabra-money-arbitrum",
           "dmelotik": "dmelotik/abracadabra-arbitrum",
-          "deploy-on-merge": true
+          "deploy-on-merge": false
```
jaimehgb, line 10: Changing this file was intended?

subgraphs/ellipsis-finance/subgraph.template.yaml
```diff
@@ -1,400 +0,0 @@
-{{#graftEnabled}}
```
jaimehgb, line 1: Why remove this template file?

deployment/deployment.json
dmelotik: I just wanted to ensure we wouldn't deploy subgraphs and lose progress since there are changes in individual folders

subgraphs/ellipsis-finance/subgraph.template.yaml
dmelotik: This was a duplicate in the subgraphs root folder. There is a copy under protocols/templates that you can see.

deployment/deployment.json
jaimehgb: Ah right, that makes sense

jaimehgb (dismissed):

dmelotik:
looksrare, opensea-v2, x2y2, and yeti-finance need to be converted to our current ci/cd

this-username-is-taken:
Great initiative!

PR #801, opened 2022-08-14
feat(#major); DEX-AMM-Extended Schema; New fields proposal

steegecs:
Added Fields: 
- UsageMetricsHourly - Added same changes made to usage metrics daily.
- LiquidityPool - Added metrics about positions.
- LiquidityPoolSnapshots - Added stats to hourly snapshots. Added positions data to all liquidity pool snapshots.
- Positions - Added ticks to track the upper and lower bound of concentrated liquidity.

It would be interesting to track these ticks as entities like they have done in the official Uniswap V3 subgraph. Positions in Uniswap V3 can have their ticks changed without necessarily closing the position.
Screen Shot 2022-08-14 at 10 14 04 AM

Would we also want to track descriptive statistics about the way liquidity is concentrated in the pool within the subgraph as well?

Screen Shot 2022-08-14 at 11 07 01 AM


KremerMichael:
LFG, stats is looking dope. Are we thinking about adding snapshots around events for the other schemas? 

ie: adding deposit, liquidate, borrow, etc... snapshots to lending schemas?

PR #806, opened 2022-08-15
fix:(#patch): trader-joe; Added bonus rewards to masterchef

### LINKED ISSUE #712: Trader Joe QA (Avalanche) Schema Version 1.3.0 Subgraph Version 1.0.1 Methodology Version	1.0.0
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/trader-joe-avalanche&tab=protocol
Date Reviewed | August 2 2022
Schema Version | 1.3.0
Subgraph Version | 1.0.1
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/1h_sEU5Ke4dTNuzGs6GkAw7o3SkgAC80_gx2axHwRpzk/edit?usp=sharing

### Metrics To Review
**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | totalValueLockedUSD | 12/01/2021 and 01/01/2022 are looking roughly 20% off to me and there is about a 50% difference on 11/01/2021. The dates before then match well though.
usageMetricsDailySnapshots | dailyTransactionCount | I think that these metrics are generally correct, so I am guessing that they are accurate, but I do not have any exact date to compare to. Uniswap V3 on mainnet has about 38k transactions while TJ has 164K on 01/06/202, which seems like it could be a little suspicious. Granted though, swapping on AVAX is cheaper, so this could be completely accurate, but would not hurt to double check.
usageMetricsDailySnapshots | dailyDepositCount | Following same logic as overall transaction count
usageMetricsDailySnapshots | dailyWithdrawCount | Following same logic as overall transaction count
usageMetricsDailySnapshots | dailySwapCount | Following same logic as overall transaction count

**Pool Metrics**
Pool | Section | Metric | Issue
-- | -- | -- | --
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | cumulativeSupplySideRevenueUSD | Looks like this is not adding correctly. The dailySupplySideRevenueUSD for 01/07/2022 is 243,677.032 and difference in cumulative values between 01/07 and 01/06 is 105,866,601.894 - 105,031,817.123 = 834,784.771
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | cumulativeProtocolSideRevenueUSD | This also looks like it is not adding correctly. The dailyProtocolSideRevenueUSD for 01/07/2022 is 48,735.406 and the difference in cumulative vales between 01/07 and 01/06 is 21,173,320.379 - 21,006,363.425 = 166,956.954
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | cumulativeTotalRevenueUSD | Same problem as cumulative metrics for supply side and protocol side revenue
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | dailyVolumeByTokenUSD - USDC | This is 0, would expect there to be values here
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | stakedOutputTokenAmount - USD Coin/Wrapped AVAX LP | TJ allows for LP tokens to be staked for JOE emissions, so there should be output for this metric. Currently, nothing being outputted.
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | rewardTokenEmissionAmount | No output. There should be output, as TJ has JOE emissions. There is reward information on the pool overview page, so not sure why it is not being reflected here. Looks like there might be an error though, as there is a warning at the top of the page stating: "LiquidityPoolDailySnapshot-rewardTokenEmissionsAmount should be an array, but has a null value in its timeseries data (first instance in LiquidityPoolDailySnapshot 0xa389f9430876455c36478deea9769b7ca4e3ddb1-18864)"
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | rewardTokenEmissions - USD | No output. There should be output, as TJ has JOE emissions.
Trader Joe USD Coin/Wrapped AVAX | liquidityPoolDailySnapshots | rewardAPR | No output. There should be output, as TJ has JOE emissions.

jaimehgb:
Trader joe masterchef pools have an additional feature called `Rewarders`. They are contracts pre-funded with an arbitrary token to give to LP stakers in addition to the masterchef default token. Every time a masterchef pool is added or modified it can include a rewarder.

The relationship between rewarders and pools is 1:1. The same rewarder cannot be assigned to more than one pool (technically it can happen, because nothing in their contract checks for it, but they explicitly mention that adding it more than once will mess up rewards).

The rewarder will be issuing rewards to LP stakers until it runs out of funds or until someone manually changes the rewarder rate to 0.

This PR adds support for traderjoe masterchef rewarders. And should probably close #712 .

### Some more details on the implementation.

- A new _MasterChefRewarder entity has been added to track the rewarder address, reward rate, and associated pool.
- Added an optional field `rewarder` to `_MasterChefStakingPool` to keep track of the rewarder assigned to it.
- Every time a mastrerchef pool is added or updated, if it includes a rewarder, a rewarder entity is created and the staking pool is assigned the rewarder address. Not further action is taken, since rewardRates are calculated on each deposit/withdrawal.
- Every deposit/withdrawal to a MC staking pool triggers an update of the default masterchef data. At the end of this process I've added the update of bonus rewards in case there is a rewarder for that pool.
- If the balance of the rewarder is 0, until it is refilled, its rewards are set to 0. I'd like your input here @steegecs or @this-username-is-taken , because that adds an additional contract call on each deposit/withdrawal. But omitting it entirely would potentially leave rewarder rates with positive values when they are not giving anything anymore. Maybe could only check the balance once per day or something like that? Or one additional call is not so big of a deal?


subgraphs/uniswap-forks/protocols/trader-joe/src/common/constants.ts
```diff
@@ -9,5 +9,6 @@ export const PROTOCOL_METHODOLOGY_VERSION = "1.0.0";
 export const PROTOCOL_NAME = "Trader Joe";
 export const PROTOCOL_SLUG = "trader-joe";
 
-export const TRADER_JOE_AVALANCHE_REWARD_TOKEN_RATE =
-  BigInt.fromI64(30000000000000000000);
```
jaimehgb, line 13: This int was overflowing 😅

steegecs:
In the past I have used the method of only updating rewards every few blocks. Like for the Sushiswap MasterChefV1 handler, I only updated the rewards rates every five blocks, because you have to do multiple contract calls to calculate the rewards. I think it is a good idea to just update over a specified interval. You'll definitely want to make it so the call only runs once per block at most.

this-username-is-taken:
yeah, we can update once every hour to speed up indexing

jaimehgb:
This is finally ready for review 🚀

steegecs:
`If the balance of the rewarder is 0, until it is refilled, its rewards are set to 0. I'd like your input here @steegecs or @this-username-is-taken , because that adds an additional contract call on each deposit/withdrawal. But omitting it entirely would potentially leave rewarder rates with positive values when they are not giving anything anymore. Maybe could only check the balance once per day or something like that? Or one additional call is not so big of a deal?`

Yeah - I think once per day sounds good. If we want to be more granular, I don't think every 6-12 hours would impact the speed much either.

subgraphs/uniswap-forks/protocols/trader-joe/src/common/handlers/handleRewarder.ts
```diff
@@ -0,0 +1,279 @@
...
+  event: ethereum.Event,
+  rewarder: _MasterChefRewarder
+): boolean {
+  // only check rewarder balance once per day
+  const ONE_DAY = BigInt.fromI32(24 * 3600);
+  if (event.block.timestamp.minus(rewarder.hasFundsAt).lt(ONE_DAY)) {
+    return rewarder.hasFunds;
+  }
+
+  let bal = fetchRewarderBalance(rewarder);
```
steegecs, line 128: Already included!

subgraphs/uniswap-forks/protocols/trader-joe/src/common/handlers/rewarderRate/rewardRateCalculator.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (tokenPerSecCall.reverted) {
+      return null;
+    }
+
+    return tokenPerSecCall.value;
+  }
+
+  // calculateRewarderRate will happen once every 6 hours.
+  // There are 2 ways of calculating the reward rate:
+  //   1. If the rewarder contract exposes `tokenPerSec`, we'll use that.
```
steegecs, line 52: Do you know if the rewarder ever exposes a tokenPerBlock() that you could utilize?

steegecs (approved):

subgraphs/uniswap-forks/protocols/trader-joe/src/common/handlers/rewarderRate/rewardRateCalculator.ts
jaimehgb: I haven't seen any but they have an example in their reference rewarder implementations. So they might appear at some point. I wanted to leave it out of the PR for now because translating `perBlock` to `perSecond` would involve a few more changes. 

I can add it after, and is a change that should not require a full resync, grafting should do.

subgraphs/uniswap-forks/protocols/trader-joe/src/common/handlers/handleRewarder.ts
jaimehgb: What do you mean? 😅

steegecs: My bad, I made this comment before I saw you already include the balancer checker every 24 Hours. Updated the comment above, but forgot this one. Just ignore!

subgraphs/uniswap-forks/protocols/trader-joe/src/common/handlers/rewarderRate/rewardRateCalculator.ts
steegecs: Okay sounds fine to me.

this-username-is-taken:
Great work guys!

PR #815, opened 2022-08-16
feat(#minor): mm-finance; added matic mm-finance subgraph

jaimehgb:
Added support for mm-finance polygon subgraph. Deploys on merge because it syncs super fast.

For deploy (and readme subgraph link) to work, the subgraph needs to be created first at the hosted service. I've named it`mm-finance-polygon`. I've tried to do it from the graph-cli but can't get it to work, might not be possible? If you could check @this-username-is-taken 💪 

Demo of the sync: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/jaimehgb/mm-finance-polygon&tab=protocol


this-username-is-taken:
Great work!

PR #816, opened 2022-08-16
feat(#minor); Uniswap V3; add Celo network

steegecs:
**Context** 
Celo network has been added to Uniswap V3. It was added a little over a month ago. The protocol has fairly low TVL currently at about 1.1 million. As a result, I have set a much lower minimum liquidity threshold for tracking token prices at about $35,000 USD. We may want to lower this depending on how well it tracks data relating to USD value, however, this can sometimes leading to a lot of abnormal pricing behavior in the Uniswap V3 subgraphs. 

**To do**
Add Uniswap V3 Celo endpoint to the Messari Hosted Service.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -698,7 +698,7 @@
         },
         "celo": {
           "template": "uniswap.v3.template.yaml",
-          "deploy-on-merge": false,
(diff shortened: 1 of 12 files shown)


this-username-is-taken:
🚀🚀🚀

PR #817, opened 2022-08-17
fix(); Compound forks; use getMarket(Hourly/Daily)SnapshotID function

0xbe1:


this-username-is-taken:
nice catch!

dmelotik:
@0xbe1 this should either be #patch and change and increment the version or nothing in (), but the change lgtm.

dmelotik (changes requested):
changes req

0xbe1:
@dmelotik updated pr description

PR #818, opened 2022-08-17
fix(#patch); cream finance; handle oracle that returns price in eth

0xbe1:
Syncing deployment: https://okgraph.xyz/?q=messari%2Fcream-finance-ethereum

The current issue with cream finance being:
- We use compound fork framework to handle cream finance
- The framework expects oracle to return price in USD, however cream finance uses an oracle that returns price in ETH instead

To work around the issue, we could:
- Derive ETH price
- Multiply the oracle price by ETH price before passing the number down to the compound fork framework

**NOTICE** This patch only fixes cream finance on Ethereum. Let us verify the patch works before fixing this for the rest networks (BNB, Polygon, etc)

### DIFF
--- a/subgraphs/compound-forks/protocols/cream-finance/src/mapping.ts
+++ b/subgraphs/compound-forks/protocols/cream-finance/src/mapping.ts
@@ -19,7 +19,11 @@ import {
   NewReserveFactor,
 } from "../../../generated/templates/CToken/CToken";
 import { LendingProtocol, Token } from "../../../generated/schema";
-import { cTokenDecimals, BIGINT_ZERO } from "../../../src/constants";
+import {
+  cTokenDecimals,
+  BIGINT_ZERO,
+  exponentToBigDecimal,
+} from "../../../src/constants";
 import {
   ProtocolData,
   _getOrCreateProtocol,
@@ -48,6 +52,7 @@ import { CToken as CTokenTemplate } from "../../../generated/templates";
 import { ERC20 } from "../../../generated/Comptroller/ERC20";
 import { getNetworkSpecificConstant } from "./constants";
 import { PriceOracle } from "../../../generated/templates/CToken/PriceOracle";
+import { ethereum } from "@graphprotocol/graph-ts";
 
 // Constant values
 let constant = getNetworkSpecificConstant();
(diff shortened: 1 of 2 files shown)


subgraphs/compound-forks/protocols/cream-finance/src/mapping.ts
```diff
@@ -261,12 +269,49 @@ export function handleAccrueInterest(event: AccrueInterest): void {
...
+  // to work around, for cream finance specifically, we get eth price first,
+  // multiple the oracle price by eth price, then pass it down to common logic.
+  //
+  // to compute eth price, we derive it from usdt price in eth.
+  // let's say if we know 1 usdt = 0.001 eth, then we know 1 eth = 1000 usd.
+  // to get usdt price, we simply do 10^(36-6)/getUnderlyingPrice(crUSDT), in which 6 = decimals(USDT)
+  //
+  // you may ask, why USDT, not USDC or DAI?
+  // this is bc crUSDT was deployed the earliest on mainnet compared to crUSDC and crDAI.
+  let usdtPriceResult = oracleContract.try_getUnderlyingPrice(
```
dmelotik, line 284: should this specfic logic be wrapped by an "if ethereum network" since that mkt only exists on ethereum?

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1877,7 +1877,8 @@ export function getOrCreateMarketHourlySnapshot(
   return snapshot;
 }
 
-function getTokenPriceUSD(
+// Return price in either USD or ETH, depending on oracle each fork uses
+function getTokenPrice(
```
dmelotik, line 1881: shouldn't this still be `getTokenPriceUSD()`? since the CallResult you are passing down is normalized to USD no matter what..

dmelotik (commented):
Should this increment the cream version?

subgraphs/compound-forks/src/mapping.ts
0xbe1: ah good catch!

subgraphs/compound-forks/protocols/cream-finance/src/mapping.ts
0xbe1: @dmelotik  how about me adding the crSTABLE address to config? so that it easily covers all network :)

dmelotik: I'm slightly confused. Would you have one for each network and have logic to switch based on the network?

dmelotik:
I think we forgot about this PR, but the issue was addressed in #967. @0xbe1 if you agree with this I think this can be closed

0xbe1:
> I think we forgot about this PR, but the issue was addressed in #967. @0xbe1 if you agree with this I think this can be closed

cool let's close then

PR #828, opened 2022-08-18
fix(#patch); rari fuse; fix holey array issue

dmelotik:
Fixing this issue / move to version `1.1.2`:

image

Test: https://okgraph.xyz/?q=dmelotik%2Ffuse-mainnet


this-username-is-taken:
Good job!

PR #833, opened 2022-08-18
fix (#minor); makerdao; fixed deployment fail

tnkrxyz:
- grafting on last successful deployment
- added DAI abi, just in case
- fixed a bug in setting daily financial protocol side revenue


subgraphs/makerdao/src/common/helpers.ts
```diff
@@ -355,7 +355,7 @@ export function updateFinancialsSnapshot(
   let newProtocolSideRevenueUSD = BIGDECIMAL_ZERO;
   if (newTotalRevenueUSD.gt(BIGDECIMAL_ZERO)) {
     financials.dailyTotalRevenueUSD = financials.dailyTotalRevenueUSD.plus(newTotalRevenueUSD);
-    let newProtocolSideRevenueUSD = newTotalRevenueUSD.minus(newSupplySideRevenueUSD);
```
this-username-is-taken, line 358: this bug you can't graft right? you'd need to reindex from scratch

Issue #821, opened 2022-08-17
AAVE V2 Avalanche and Polygon (Schema Version 2.0.1, Subgraph Version	 1.2.6, Methodology Version 1.0.0)


ishraq8:
**Avalanche**

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/aave-v2-avalanche&tab=protocol
-- | --
Date Reviewed | 17-Aug-22
Schema Version | 2.0.1
Subgraph Version | 1.2.6
Methodology Version | 1.0.0

Protocol |   |   |   |  
-- | -- | -- | -- | --
Section | Value | Status |   | Comments
financialsDailySnapshots | totalValueLockedUSD | not ok | Public data | Prior to april 20, 2022 the graph shows double TVL to defillama
financialsDailySnapshots | totalBorrowBalanceUSD | not ok | Public data | before april 20, 2022, borrowed balance seems to be amount from expected TVL. need to double check if this is related to `totalvaluelockedUSD` issue
financialsDailySnapshots | totalDepositBalanceUSD | not ok | Sanity check | Prior to april 20, 2022 the graph shows double TVL to defillama
financialsDailySnapshots | cumulativeLiquidateUSD | not ok |   | Not adding up liquidate data. Output is zero
usageMetricsDailySnapshots | cumulativeUniqueLiquidatees | not ok |   | shows zero when there are liquidations.
-- | -- | -- | -- | --

**Polygon**

Still indexing, will take time. Skipping for Mainnet launch and will QA once done.

dmelotik:
@ishraq8 I actually believe defiLlama is wrong prior to 4/20 (lol). I am cross-referencing with Aave's historical API calls (see api schema [here](https://aave-api-v2.aave.com/#/data/get_data_liquidity_v2))

[...]

PR #834, opened 2022-08-19
fix(#patch); aave v2 forks; address avax QA

dmelotik:
Addressing: #821 

Upgrade to version `1.2.10`


this-username-is-taken (approved):
LGTM!

Issue #773, opened 2022-08-10
AAVE V2 (mainnet, arc, rwa and AMM) Schema Version 2.0.1, Subgraph Version  1.2.6, Methodology Version  1.0.0


ishraq8:
Summary: `cumulativeLiquidateUSD` not adding up in all subgraphs and double check if `isCollateral` should be false in Positions tab. Some subgraphs have very low usage so no liquidations, double check ARC and RWA if subgraph is accurately reading this zero. 

**Aave v2 Mainnet**

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/aave-v2-ethereum-extended
-- | --
Date Reviewed | 26-Jul-22
Schema Version | 2.0.1
Subgraph Version | 1.2.6
Methodology Version | 1.0.0

Protocol |   |   |   |  
-- | -- | -- | -- | --
Section | Value | Status |   | Comments
financialsDailySnapshots | cumulativeLiquidateUSD | not ok |   | Not adding up liquidate data. Output is zero

Postions |   | not ok | Etherscan |  The data from etherscan shows that stETH tokens were enabled as collateral. However, the  'isCollateral' column in the subgraph data shows says FALSE. Should it be false? Double check The ranking in balance column is not sorted by value (seemingly sorted according to the alphanumerical order of the value data). 
-- | -- | -- | -- | --

**Aave v2 ARC**

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/aave-arc-ethereum&tab=protocol
-- | --
Date Reviewed | 9-Aug-22
Schema Version | 2.0.1
Subgraph Version | 1.2.6
Methodology Version | 1.0.0

Protocol |   |   |   |  
-- | -- | -- | -- | --
Section | Value | Status |   | Comments
financialsDailySnapshots | dailyLiquidateUSD | NOT OK |   | Usage is very low so shows at zero, double check if subgraph will report if there is an event.
financialsDailySnapshots | cumulativeLiquidateUSD | not ok |   | Double check if adding up liquidate numbers

Positions |   | not ok | Etherscan |  The data from etherscan shows that stETH tokens were enabled as collateral. However, the  'isCollateral' column in the subgraph data shows says FALSE. Should it be false? Double check The ranking in balance column is not sorted by value (seemingly sorted according to the alphanumerical order of the value data). 
[...]

PR #838, opened 2022-08-19
fix(#patch); aave v2 forks; fix lendingConfig issues

dmelotik:
Addressing QA #773 

issues:
- The `reserveFactor` is not being properly updated in the amm markets in some cases.

Fix: 
- make the `lendingPoolConfiguration` handlers/templates less verbose and hardcode it in instead of tracking the proxy upgrades as a template
- remove LendingPool template since it is more verbose and prone to error than just tracking the proxy listed here: https://docs.aave.com/developers/v/2.0/deployed-contracts/amm-market
- add `getOrCreateMarket()` since reserve factor is being set on markets before they are initialized in order to capture these changes

Testing: 
- https://okgraph.xyz/?q=dmelotik%2Faave-v2-ethereum-amm


jaimehgb (approved):
LGTM 🚀 

Just a few questions for my own learning purposes (non-merge-blocking):

What were the benefits of tracking proxy upgrades via templates? What do we potentially lose from ignoring them?

Does the contract API change in a relevant way in any of those upgrades?

dmelotik:
@jaimehgb 

> What were the benefits of tracking proxy upgrades via templates? What do we potentially lose from ignoring them?

I am not sure there are any benefits. This is what aave did in their official subgraph and when we initially made aave v2 we copied that, and i didn't pay much mind to it bc it worked. It is possible that this is what needed to be done in an old version in graph-cli. but I would say this is not good practice anymore, it is too verbose and makes a simple thing harder to understand.

> Does the contract API change in a relevant way in any of those upgrades?

Not from what I am aware of, we would probably have issues if that was the case. Because tracking the new proxy every time does not give you the new ABIs. When an ABI is changed in a proxy upgrade we would need that new ABI in the mappings in order to utilize the events/functions.

PR #839, opened 2022-08-19
fix(#patch); iron bank; skip bad ctoken

0xbe1:
Syncing deployment: https://okgraph.xyz/?q=0xbe1%2Firon-bank-subgraph-mainnet

See comments in code. This should fix both the totalBorrows as well as the TVL.

QA spec mentions a few other issues:
- latest events are months old: I don't think that's right, since events from subgraphs are very up-to-date, e.g. https://api.thegraph.com/subgraphs/id/QmR5xWsyobfRGoWUr5YJqm1mtuuHQ3egEgK1cnWodUgEoL/graphql?query=%7B%0A++borrows%28orderBy%3Atimestamp%2CorderDirection%3Adesc%29%7B%0A++++id%0A++++timestamp%0A++%7D%0A%7D
- totalPoolCount (40) more than ironbank dashboard (24): I believe ironbank simply picks certain pools to render :) we can hardcode the 24 pools in subgraphs and only index those markets, but I am not sure if we want to do that. Thoughts? @this-username-is-taken

### DIFF
--- a/subgraphs/compound-forks/protocols/iron-bank/src/mapping.ts
+++ b/subgraphs/compound-forks/protocols/iron-bank/src/mapping.ts
@@ -1,4 +1,4 @@
-import { Address, BigInt, log } from "@graphprotocol/graph-ts";
+import { Address, BigInt, dataSource, log } from "@graphprotocol/graph-ts";
 // import from the generated at root in order to reuse methods from root
 import {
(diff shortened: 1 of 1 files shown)


this-username-is-taken:
nah, we should have all pools.

PR #842, opened 2022-08-19
fix/refactor(#major); Balancer Forks; Fix bugs with rewards fields, vault mappings, and Balancer Emissions.

steegecs:
**Beethoven X Fantom is set to deploy on merging this PR. We should consider re-deploying all of Balancer V2 as well since the stakedOutputTokenAmount and rewardEmissions will be off.**

**Bug Fixes:**
- The calculation of the LiquidityPool stakedOutputTokenAmount field needed to be done in the updateFactoryRewards function. This way the staked token amount will be calculated properly for all gauges on all networks. Before it was only working for Balancer V2 on Ethereum.
- Needed to add an exception here as per Phoenix's request for registering tokens in vault mappings -> https://github.com/harsh9200/subgraphs/blob/Balancer-v2--Polygon-and-Arbitrium-support/subgraphs/balancer-forks/src/mappings/vaultMappings.ts#L47-L48
- Balancer emissions from the gauge controller were assumed to be static at 145,000 per week. With the introduction of veBAL, however, these emissions are set to decrease over time and updatable through governance. I added 2 fields to the RewardToken entity to track the rates. You can watch for updates in the MiningParametersUpdated event in this contract -> https://etherscan.io/address/0xf302f9f50958c5593770fdf4d4812309ff77414f#readContract. Learn more about this here -> https://forum.balancer.fi/t/introducing-vebal-tokenomics/2512#new-proposed-bal-inflation-schedule-4.
- Fixed reward token id to include the type in front of the address separated by a hyphen.
- Moved getPoolFromGauge() function out of utils so Beethoven X Fantom can deploy. This function requires loading in the gauge contract which Beethoven X Fantom does not use.

### DIFF
--- a/subgraphs/balancer-forks/src/mappings/balancerTokenAdminMappings.ts
+++ b/subgraphs/balancer-forks/src/mappings/balancerTokenAdminMappings.ts
@@ -1,0 +1,25 @@
+import { BigDecimal } from "@graphprotocol/graph-ts";
+import { MiningParametersUpdated } from "../../generated/BalancerTokenAdmin/BalancerTokenAdmin";
+import * as constants from "../common/constants";
+import { getOrCreateRewardToken } from "../common/initializers";
+import { getRewardsPerDay } from "../common/rewards";
+
+export function handleMiningParametersUpdated(
+  event: MiningParametersUpdated
+): void {
+  let protocolToken = getOrCreateRewardToken(
+    constants.PROTOCOL_TOKEN_ADDRESS,
+    constants.RewardTokenType.DEPOSIT,
+    event.block
+  );
+  protocolToken._inflationRate = BigDecimal.fromString(
+    event.params.rate.toString()
+  );
+  protocolToken._inflationPerDay = getRewardsPerDay(
+    event.block.timestamp,
+    event.block.number,
+    protocolToken._inflationRate!,
+    constants.INFLATION_INTERVAL
+  );
+  protocolToken.save();
+}
--- a/subgraphs/balancer-forks/src/modules/Rewards.ts
+++ b/subgraphs/balancer-forks/src/modules/Rewards.ts
@@ -16,6 +16,7 @@ export function getRewardsData(gaugeAddress: Address): RewardsInfoType {
   let rewardTokens: Address[] = [];
 
   let gaugeContract = LiquidityGaugeContract.bind(gaugeAddress);
+
   let rewardCount = utils.readValue<BigInt>(
     gaugeContract.try_reward_count(),
     constants.BIGINT_TEN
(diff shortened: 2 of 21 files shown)


subgraphs/balancer-forks/src/mappings/balancerTokenAdminMappings.ts
```diff
@@ -0,0 +1,25 @@
...
+  event: MiningParametersUpdated
+): void {
+  let protocolToken = getOrCreateRewardToken(
+    constants.PROTOCOL_TOKEN_ADDRESS,
+    constants.RewardTokenType.DEPOSIT,
+    event.block
+  );
+  protocolToken._inflationRate = BigDecimal.fromString(
+    event.params.rate.toString()
+  );
```
harsh9200, line 17: shouldn't we divide this number by `pow(10, 18)`?

subgraphs/balancer-forks/src/modules/Rewards.ts
```diff
@@ -49,21 +51,51 @@ export function updateControllerRewards(
...
+    return;
+  }
+
+  let protocolToken = getOrCreateRewardToken(
+    constants.PROTOCOL_TOKEN_ADDRESS,
+    constants.RewardTokenType.DEPOSIT,
+    block
+  );
+  // Daily BAL emissions is currently known as a static value, but it may be changed in the future.
+  // FOLLOW UP: How to keep track of the daily emission rate?
```
harsh9200, line 77: we can remove this now

subgraphs/balancer-forks/src/mappings/balancerTokenAdminMappings.ts
steegecs: It looks like you are setting the rewardTokenEmissionsAmount to the amount before decimal upscaling since it is in the full token about in the constants. The base for rewardTokenEmissionsAmount should be #tokens * 10^token.decimals. So I thought this was correct. 

Screen Shot 2022-08-20 at 11 59 55 PM
Screen Shot 2022-08-20 at 11 54 43 PM
Screen Shot 2022-08-20 at 11 55 15 PM

subgraphs/balancer-forks/src/modules/Rewards.ts
steegecs: Updated the comment!

subgraphs/balancer-forks/src/mappings/balancerTokenAdminMappings.ts
harsh9200: I thought the FE ([subgraphs.xyz](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/balancer-v2-ethereum&tab=pool&poolId=0x32296969ef14eb0c6d29669c550d4a0449130230)) doesn't handle the decimals manipulation for `rewardTokenEmissionAmoiunt`. So I normally divide the rewards emission per day by the rewards decimals `pow(10, 18)`.

harsh9200 (approved):
LGTM, Good Work!!

PR #846, opened 2022-08-21
Balancer Forks: Fix bugs/exception with stableFactoryPools

harsh9200:


subgraphs/balancer-forks/src/modules/Rewards.ts
```diff
@@ -116,7 +114,13 @@ export function updateFactoryRewards(
   );
   pool.save();
   //////////////////////////////////////////////////////////////////
+}
 
+export function updateFactoryRewards(
+  poolAddress: Address,
+  gaugeAddress: Address,
+  block: ethereum.Block
+): void {
```
steegecs, line 123: Won't this miss the stakedOutTokenAmount of the side chains though? Are you thinking there is not a way to calculate this for chains that are not Ethereum mainnet? My impression was that the updateControllerRewards() will only end up updating the stakedOutputTokenAmount for Ethereum mainnet if it is left there, because this code will only get executed on Ethereum mainnet.

subgraphs/balancer-forks/src/prices/common/constants.ts
```diff
@@ -46,6 +46,22 @@ export const WHITELIST_TOKENS_LIST: string[] = [
   "CRV",
 ];
 
+export const USD_PEGGED_TOKENS: string[] = [
```
steegecs, line 49: Why did you add these here?

steegecs: *All of the USD_PEGGED_TOKENS.

harsh9200: These are all the stable tokens on which the Price lib failed to fetch a correct price.

steegecs (commented):
Okay - Everything looks good to me!

steegecs (approved):

PR #849, opened 2022-08-22
fix(#minor); Uniswap V3; Minor fixes to pool entities and usage metrics calculation

steegecs:
**Context:**
1. The usage metrics we much lower than expected. This occurs, because the event.params.sender in the swap event is a proxy and not the actual sending account of the swap. 
2. The pool instantiation for Optimism regenesis included some fields that were unnecessary. 
3. Egregious TVL error on Uniswap V3. 

Issue #869 
4. The Pool name percentages needed to be updated to they are not all showing 0% at the end.

**Results:**
1. Count usage metrics on swaps by using event.params.from
2. Updated pool instantiation in regenesis to match the pool instantiation in the creators folder. Also did some refactoring so the pool instantiation is neater.
3. I am still trying to determine the source of the TVL error on Uniswap V3. It appears to be an issue with token overflow in a LUSD/wETH pool, and it is the only example I can find for all Uni V3 subgraphs. I am not yet sure how this overflow is occurring. For now, I am going to blacklist this pool and re-deploy, and see if it fixes the issue. Running tests and grafting now to determine the source of the error for a future fix if I do not find it soon.

Issue #869 
4. Updated pool naming function for pools.


jaimehgb (dismissed):

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -171,7 +166,8 @@ export function findUSDPricePerToken(
             nativeTokenValueLocked.gt(largestNativeTokenValue) &&
             nativeTokenValueLocked.gt(
               NetworkConfigs.getMinimumLiquidityThreshold()
-            )
+            ) &&
+            token0.lastPriceUSD != BIGDECIMAL_ZERO
```
jaimehgb, line 170: This doesn't change anything, right?
If this branch (or the one above) doesn't enter we return the default `priceSoFar` at line 182, which will be 0.

If lastPriceUSD is 0 we'll be returning 0 too by multiplying by 0.

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -171,7 +166,8 @@ export function findUSDPricePerToken(
...
           ) {
             largestNativeTokenValue = nativeTokenValueLocked;
```
jaimehgb, line 172: Not in the scope of this PR, feel free to ignore.

`largestNativeTokenValue` doesn't seem to have any use?

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: Correct! I have removed.

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: It does. It could use a better name though so I changed it. The priceSoFar only gets updated. When the whitelist token has a greater value than the previous pools.

jaimehgb: Ah right, I missed that `for` 🙈

jaimehgb (dismissed):
LGTM!

jaimehgb (approved):

Issue #571, opened 2022-07-13
Inverse Finance QA (Ethereum) Schema Version 1.3.0 Subgraph Version 1.2.2 Methodology Version 1.0.0


bye43:
Description | Value
-- | --
Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/inverse-finance-ethereum&tab=protocol
Date Reviewed | July 13 2022
Schema Version | 1.3.0
Subgraph Version | 1.2.2
Methodology Version | 1.0.0
Evidence Spreadsheet | https://docs.google.com/spreadsheets/d/19wfBowLv5-2k4xH5Ubyo8aQKIPuQFDPiShRXhB0bR1A/edit?usp=sharing

### Metrics To Review
**Protocol Metrics**
Section | Metric | Issue
-- | -- | --
financialsDailySnapshots | dailyProtocolSideRevenueUSD | Data scaling is off. Some dates are correctly matching Dune, but others are exponentially too large. For example, 7/09/2022 is 1,992,000,000,000,000,500
financialsDailySnapshots | cumulativeProtocolSideRevenueUSD | Daily is off so this will be too
financialsDailySnapshots | dailyTotalRevenueUSD | Data scaling is off. Some dates are correctly matching Dune, but others are exponentially too large. For example, 7/09/2022 is 1,992,000,000,000,002,800
financialsDailySnapshots | cumulativeTotalRevenueUSD | Daily is off so this will be too
financialsDailySnapshots | totalDepositBalanceUSD | The trend looks correct for this, but based on the Inverse Finance UI, there is about a -50% difference in values for certain dates. Link - https://www.inverse.finance/analytics
financialsDailySnapshots | dailyDepositUSD | Big Spike on 04/02/2022 that does not look correct
financialsDailySnapshots | cumulativeDepositUSD | Daily is off, hence cumulative will also be

**Pool Overview**
Pool | Metric | Issue
-- | -- | --
INV-DOLA-SLP | Reward Token | Should be INV rewards for this pool

**Pool Metrics**
Pool | Section | Metric
-- | -- | --
marketDailySnapshots | cumulativeSupplySideRevenueUSD | Daily looks correct, but the cumulative is about 50% off from dune query. Link - https://dune.com/queries/1021789
marketDailySnapshots | cumulativeProtocolSideRevenueUSD | Scaling is off again here too. Y - axis is going up to 2E23
[...]

PR #856, opened 2022-08-23
Fix (#patch); Inverse finance; address QA issues in #571

tnkrxyz:
- More details of the fixes/changes in #571
- [x] verify totalBorrowBalanceUSD and Reward APR (warnings on subgraph.messari.io)

### DIFF
--- a/subgraphs/inverse-finance/src/common/constants.ts
+++ b/subgraphs/inverse-finance/src/common/constants.ts
@@ -102,12 +102,12 @@ export const BLOCKS_PER_YEAR = BLOCKS_PER_DAY.times(DAYS_PER_YEAR);
 export const MANTISSA_DECIMALS = 18;
 
 //////////////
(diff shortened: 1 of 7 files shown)


subgraphs/inverse-finance/src/common/constants.ts
```diff
@@ -102,12 +102,12 @@ export const BLOCKS_PER_YEAR = BLOCKS_PER_DAY.times(DAYS_PER_YEAR);
 export const MANTISSA_DECIMALS = 18;
 
 //////////////
-export const FACTORY_ADDRESS = "0x4dCf7407AE5C07f8681e1659f626E114A7667339";
+export const FACTORY_ADDRESS = "0x4dCf7407AE5C07f8681e1659f626E114A7667339".toLowerCase();
```
this-username-is-taken, line 105: why not just use the lower case string?

tnkrxyz: Good point!

PR #877, opened 2022-08-25
fix(#patch): euler finance; fix market rate calculation & divide by 0

jaimehgb:
It seems we were calculating the `depositToken<>outputToken` exchangeRate the other way around.

If `exchangeRate` is the amount of input tokens a full share of output tokens is worth, then it should be `total deposited input tokens / output token supply`. We were calculating the inverse instead. 

This was causing syncing errors becuase of eventually dividing by 0, when a pool is left empty.

I checked other protocols to see if it was our own particular way of calculating it, but no, it is wrong 😅 
Maple finance, for example, [does it right.](https://github.com/messari/subgraphs/blob/master/subgraphs/maple-finance/src/common/mappingHelpers/update/intervalUpdate.ts#L84)

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -455,7 +455,7 @@
       "euler": {
         "ethereum": {
           "template": "euler.template.yaml",
-          "deploy-on-merge": false
+          "deploy-on-merge": true
         }
       }
     },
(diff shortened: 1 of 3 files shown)


dmelotik (changes requested):
lfgtm, only thing is to update the version. And if you don't want to manually deploy I would update the deployment to do that. Good catch!

this-username-is-taken (approved):
Correct! Nice catch. As Dylan said you should update the subgraph version. You can deploy it to prod after it's merged as the current one is broken

PR #883, opened 2022-08-26
[GOV] OZ Governor: Add block and txnHash for missing proposal states

danielkhoo:
## Description
Add some additional block, timestamp and txnHash fields to the schema for building a proposal's state history downstream. 

## Related PRs
- Downstream change in Governor-Service https://github.com/messari/governor-service/pull/92

## Subgraph Link
https://thegraph.com/hosted-service/subgraph/messari/ens-governance

## Screenshots (if appropriate):
image

### DIFF
--- a/subgraphs/openzeppelin-governor/protocols/unlock-governance/src/unlock-protocol-governor.ts
+++ b/subgraphs/openzeppelin-governor/protocols/unlock-governance/src/unlock-protocol-governor.ts
@@ -64,7 +64,7 @@ export function handleProposalExecuted(event: ProposalExecuted): void {
 
 // ProposalQueued(proposalId, eta)
 export function handleProposalQueued(event: ProposalQueued): void {
-  _handleProposalQueued(event.params.proposalId, event.params.eta);
+  _handleProposalQueued(event.params.proposalId, event.params.eta, event);
 }
 
 // QuorumUpdated(oldQuorum, newQuorum)
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -996,7 +996,7 @@
           "template": "angle-governance.template.yaml",
           "danielkhoo": "danielkhoo/angle-governance",
           "messari": "messari/angle-governance",
-          "deploy-on-merge": false
+          "deploy-on-merge": true
         }
       },
       "code4rena-governance": {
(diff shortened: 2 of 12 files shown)


subgraphs/openzeppelin-governor/protocols/unlock-governance/src/unlock-protocol-governor.ts
```diff
@@ -64,7 +64,7 @@ export function handleProposalExecuted(event: ProposalExecuted): void {
 
 // ProposalQueued(proposalId, eta)
 export function handleProposalQueued(event: ProposalQueued): void {
-  _handleProposalQueued(event.params.proposalId, event.params.eta);
+  _handleProposalQueued(event.params.proposalId, event.params.eta, event);
```
mmnavarr, line 67: Feels weird to pass 3 args -- 2 of which are fields from event and then pass the entire event as well

Should either:
1) pass only the event fields we care about
or
2) pass the entire event and let the fn take what it needs (probably more forward compatible)

danielkhoo: i agree about the weirdness, but this method allows us to have typed args for the generalised handler. `event.params` are unique to the contract while the `_handleProposalQueued` function uses a generic `ethereum.event`.

The alternative is to fully list out each arg of the event args like txnHash, block number, and timestamp but we're already at 10+ args for some functions.

mmnavarr: Maybe its worth passingin event.transaction and event.block then? That way the handlers get only what they need

The handlers then receive param specific items (proposalId + eta), and transaction + block which feels cleaner from a data perspective

Not too heavily opinionated on it so if you disagree lets roll with what you have

danielkhoo: I'm okay to make the change, but could we do it in a separate future PR?

mmnavarr (approved):

subgraphs/openzeppelin-governor/protocols/unlock-governance/src/unlock-protocol-governor.ts
mmnavarr: Lets keep it as is then

PR #897, opened 2022-08-29
fix(#patch): aave-v3; sort reward tokens

jaimehgb:
AAVE-V3 reward tokens order depends on when they are added. So it might happen that the emissions and token IDs don't match when querying graphql.

This PR makes it so reward tokens are sorted every time they are updated.


this-username-is-taken (approved):
LGTM!

PR #898, opened 2022-08-29
fix(#patch); abracadabra; div by 0 error

dmelotik:
Fixing divide by 0 error and grafting each abra subgraph

updating to version `1.2.7` testing: https://okgraph.xyz/?q=dmelotik%2Fabracadabra-fantom

cc @this-username-is-taken is it okay to graft these on `messari/abracadabra-money-...` since it is not quite prod ready?


this-username-is-taken:
yep, works for me

PR #905, opened 2022-08-30
Add Radiant Capital methodology

SerenityFund:


subgraphs/radiant-capital/README.md
```diff
@@ -0,0 +1,171 @@
...
+#### Borrowing
+- Borrow after depositing assets as collaterals
+- Repay borrowed funds
+- Claim RNDT rewards from the borrowed assets
+
+#### Liquidation
+- Liquidate undercollateralized positions/loans
+
+### 1-Click Looping
+- Automate and repeat the process of borrow and deposit cycle multiple times by looping to earn greater yield
```
jaimehgb, line 71: Does this count as one user action? or one per underlying?
If the user loops 5 times, will it count 5 deposits + 5 borrows + 10 transactions? or 5 deposits + 5 borrows + 1 transaction?

subgraphs/radiant-capital/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Users who lock their RDNT earn 50% of the platform fees too, generated from borrowing interest. They also receive penalty fees from users from exit early from their vested tokens. 
+Rewards for locked RDNT = Platform fees + Penalty fees
+
+Note: 20% of all RDNT emissions are allocated to Pool2 (RDNT/WETH) rewards.
+Reference: https://docs.radiant.capital/docs/project-info/vesting-and-locking/pool2
+
+## Protocol owned liquidity
+3% of RDNT allocation is for Treasury and Liquidity.
+Reference: https://docs.radiant.capital/docs/project-info/rdnt-token
```
jaimehgb, line 169: To account for protocol owned liquidity, do we only count LPs owned by the treasury? or both LPs and liquid RDNT kept at the treasury address?

subgraphs/radiant-capital/README.md
```diff
@@ -0,0 +1,171 @@
...
+Note: At the time of writing, the platform only supports assets or tokens such as BTC, ETH, USDC, USDT, and DAI.
+
+And,
+
+> TVL of Pool2 = $\sum$  value of LP tokens of RDNT/WETH from SushiSwap staked in Radiant platform
+
+Note: This refers to RDNT/WETH liquidity pool on SushiSwap and staking the LP token in Radiant to earn more RDNT.
+
+### Total Revenue
+> Total Revenue = $\sum$  value of interest payments incurred by borrowers (platform fees)
```
jaimehgb, line 104: Liquidation fees should be left out of total revenue?

subgraphs/radiant-capital/README.md
SerenityFund: I suggest we judge basing on if there's a separate function for this in the smart contract. If not a smart contract function but just a UI feature, this is not different from a user doing this loop manually.

subgraphs/radiant-capital/README.md
SerenityFund: Hi I think we do not count RDNT tokens as protocol owned liquidity. Treasury is not protocol owned liquidity. Will update this part to make it clear.

subgraphs/radiant-capital/README.md
SerenityFund: Liquidation is not revenue. Revenue is only interest paid.

subgraphs/radiant-capital/README.md
jaimehgb: Hmm, it is an additional contract, separate from the main aave forked ones. A single call to this contract results in 5 + 5 calls to the lending pool. It all happens in a single EVM transaction.

If we observe the main contracts directly it will seem as individual actions. But from the UI it will be a single transaction or action.

subgraphs/radiant-capital/README.md
this-username-is-taken: @ishraq8 just wanted to confirm with you here, do we include liquidation as revenue?

ishraq8: We don not include liquidations as revenue. It makes it more complicated. Radiant looks like an AAVE/COMP fork on arbitrum, we don't include liquidations as revenue on those protocols either. The revenue from liquidations go to liquidators using a `liquidationCall()` function ([docs](https://docs.radiant.capital/docs/project-info/borrow/liquidations). It is not distributed to suppliers.

subgraphs/radiant-capital/README.md
SerenityFund: Ok noted. I guess it's a judgement call for Messari protocol specialist then.

jaimehgb: @ishraq8 what do you think?

jaimehgb (approved):

subgraphs/radiant-capital/README.md
ishraq8: Here is my logic. In AAVE we consider manual looping as 5+5, in this it is automated through a separate contract (5+5). I am thinking keeping it 5+5 to keep the counting consistent between AAVE and forks like this. It also makes it easier to compare to AAVE. We may want to consider it one transaction on the front end depending how we digest this in the future.

subgraphs/radiant-capital/README.md
this-username-is-taken: Sounds good. Thanks @ishraq8!

PR #907, opened 2022-08-31
Feat(#major); Monitor Data Validation; Adding pool level validation

MichaelC1999:
PR for the pool level monitoring service


dmelotik:
@MichaelC1999 can I motion for some text formatting in the monitor? 

for example `euler-ethereum` and *Issue* or **Warning**

MichaelC1999:
> @MichaelC1999 can I motion for some text formatting in the monitor?
> 
> for example `euler-ethereum` and _Issue_ or **Warning**

The messaging is just temporary, the functionality is the focus atm. When we start revising the messaging we'll implement formatting like what youre recommending.

monitor/Deployments_7-30-2022.json
```diff
@@ -0,0 +1,6478 @@
+{
+	"apeswap-bsc": {
```
this-username-is-taken, line 2: what's this file for? I don't think you shuold include it in the PR

Issue #866, opened 2022-08-24
Banker Joe Avalanche QA


this-username-is-taken:
https://docs.google.com/spreadsheets/d/1PcbldMq3p35GH5lfuD95SVzoAdaARHXW/edit#gid=129452289

dmelotik:
What to look at:
- ✅ protocol totalBorrowBalanceUSD, off by $10k at one point, think that's okay
- ✅ cumulativeUniqueLiquidatees fixed in previous change
- isActive needs a second look
- canBorrowFrom needs a second look
- look at revenues
- ✅ jWBTC exchangeRate, outputTokenSupply, and outputTokenPrice needs to be looked at
    - says the outputTokenSupply / price are flipped: looks okay to me
    - WBTC may be confusing due to the large price so it looks abnormal compared to other tokens
- ✅ rates flagged for wBTC but i don't see anything wrong

Need to add rewards, even though no rewards have ever been accrued to make future proof.

this-username-is-taken:
@ishraq8 this is ready for another round 🚀

ishraq8:
This looks good. issues are fixed. There was some confusion regarding `outputtokenprice` in the original QA. `outputtoken` will not follow the price of Bitcoin (could be worth zero with no secondary market. `inputtoken` matches perfectly. As always, great work @dmelotik.

PR #908, opened 2022-08-31
fix(#patch); banker joe; address QA

dmelotik:
Fixing #866 

Changes:
- update to `1.1.4`
- make a contract call check for isActive / canBorrow from every day (each snapshot) to ensure proper data

Testing: https://okgraph.xyz/?q=dmelotik%2Fbanker-joe-avalanche


this-username-is-taken (approved):
LGTM!

Issue #916, opened 2022-09-01
VVS Finance Chronos (Schema Version	1.3.0, Subgraph Version	1.1.5, Methodology Verstion 	1.0.0) 


ishraq8:
Summary: TVL is off. seems like WCRO pools not reporting what we see on front end leading to a 100M discrepency.

Subgraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://graph.cronoslabs.com/subgraphs/name/messari/vvs-finance&tab=protocol
-- | --
Date Reviewed | 18th Aug 2022
Schema Version | 1.3.0
Subgraph Version | 1.1.5
Methodology Verstion | 1.0.0

Protocol |   |   |  
-- | -- | -- | --
Section | Value | Status | Comments
financialsDailySnapshots | totalValueLockedUSD | NOT OK | General shape of graph seems good, but the numbers do not match up. Appears WCRO pools are either not reporting or showing uop as half
financialsDailySnapshots | dailyVolumeUSD | NOT OK | General shape of graph seems good, but the numbers do not match up/
financialsDailySnapshots | cumulativeVolumeUSD | NOT OK |  

Pool overview |   |   |  
-- | -- | -- | --
Rewardtokens | Rewardtokens | Not ok | Front end

Pool |   |   |   |  
-- | -- | -- | -- | --
Wcro/WETH |   |   |   |  
liquiditypoolDailySnapshots | totalValueLockedUSD | not ok | front end | should be 16 mil showing 8 mil
Wcro/USDC |   |   |   |  
liquiditypoolDailySnapshots | totalValueLockedUSD | not ok | front end | should be 14 mil showing 7 mil
Wcro/WBTC |   |   |   |  
liquiditypoolDailySnapshots | totalValueLockedUSD | not ok | front end | Non existant on subgraph
Wcro/ATOM |   |   |   |  
liquiditypoolDailySnapshots | totalValueLockedUSD | not ok | front end | should be 3 mil showing 1.7M
Wcro/Tonic |   |   |   |  
liquiditypoolDailySnapshots | totalValueLockedUSD | not ok | front end | should be 3 mil showing 1.5M
  |   |   |   |

this-username-is-taken:
Quick note here we are pending for a response from VVS's team on rewards. I think we should skip rewards for now

this-username-is-taken:
see https://github.com/messari/subgraphs/issues/935

this-username-is-taken:
[...]

PR #925, opened 2022-09-01
fix (#minor); VVS Finance; Added lower case to token addresses. 

steegecs:
**Context:**
Response to Issue #916. wCRO was giving 0 values for the price. This was because the addresses were read from strings with capital letters in them.


this-username-is-taken:
why not just replace them with lowercase strings: https://convertcase.net/

steegecs:
> why not just replace them with lowercase strings: https://convertcase.net/

Updated

PR #928, opened 2022-09-02
fix (#patch); makerdao; fixed totalBorrowBalanceUSD after liquidation

tnkrxyz:
See #500 for more details. Will be ready for review when the deployed subgraph (https://thegraph.com/hosted-service/subgraph/tnkrxyz/makerdao-ethereum?version=pending) finishes syncing and verified fixed.

### DIFF
--- a/subgraphs/makerdao/protocols/makerdao/config/templates/makerdao.template.yaml
+++ b/subgraphs/makerdao/protocols/makerdao/config/templates/makerdao.template.yaml
@@ -1,11 +1,6 @@
 specVersion: 0.0.4
 schema:
   file: ./schema.graphql
-features:
-  - grafting
-graft:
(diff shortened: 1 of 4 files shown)


tnkrxyz:
The deployed subgraph addressed the remaining issue in totalBorrowBalanceUSD, for 8/19/2022 (and other days), caused by not removing borrow balance after a liquidation: https://thegraph.com/hosted-service/subgraph/tnkrxyz/makerdao-ethereum

tnkrxyz:
I also removed the grafting, so this subgraph will sync from the startBlock. It will take longer, but can make sure it work correctly from scratch since we have pushed some many patches.

tnkrxyz:
Ready for review/merge @this-username-is-taken

PR #944, opened 2022-09-06
feat(#major); rocket pool; adding 🚀 pool subgraph

zdql:
Currently, largest error with rocket-pool subgraph is the price oracle failing to find accurate prices for rETH. Once rETH pricing is fixed, revenue is essentially complete. 

(This excludes RPL, which depends on separate issues but will be added shortly as the roadmap to solution is much clearer)


subgraphs/rocket-pool/src/entities/token.ts
```diff
@@ -0,0 +1,121 @@
...
+  if (tokenAddress == Address.fromString(RETH_ADDRESS)) {
+    rethpriceUSD = getRETHpriceUSD(
+      Address.fromString(RETH_ADDRESS),
+      blockNumber
+    );
+    log.warning("[getOrCreateToken]rethPriceUSD call result: {} ", [
+      rethpriceUSD.toString(),
+    ]);
+  }
+  let price = getUsdPricePerToken(tokenAddress);
```
jaimehgb, line 49: Since you have available the block number here, you can check the last time the price was updated for this token so you don't fetch it multiple times on the same event or block.

subgraphs/rocket-pool/src/entityUpdates/financialMetrics.ts
```diff
@@ -0,0 +1,422 @@
...
+
+  log.error("[updateProtocolAndPoolTvl] reward tokens check: {}", [
+    rewardTokens[0],
+  ]);
+  pool.rewardTokens = rewardTokens;
+  pool.save();
+  // Pool
+  let inputTokenBalances: BigInt[] = [];
+  inputTokenBalances.push(pool.inputTokenBalances[0].plus(amount));
+  inputTokenBalances.push(pool.inputTokenBalances[1].plus(rewardAmount));
```
jaimehgb, line 88: These amounts will be referring to the wrong token, as per what was said on one of the previous commments

subgraphs/rocket-pool/src/mappings/RocketPool.ts
```diff
@@ -0,0 +1,251 @@
...
+    event.block,
+    BIGINT_NEGATIVE_ONE.times(event.params.amount),
+    BIGINT_ZERO
+  );
+  updateSnapshotsTvl(event.block);
+}
+
+/** Can potentially use minipool created/destroyed to track this instead */
+
+/** Staking rewards is a function of amt of reth/eth ratio so can be calculated for any minipool based on amt of staked eth, and then slashing is subtracted from it ? */
```
jaimehgb, line 95: Is this a question? :P

subgraphs/rocket-pool/protocols/rocketpool/config/templates/rocketpool.template.yaml
```diff
@@ -0,0 +1,427 @@
...
+  #   mapping:
+  #     kind: ethereum/events
+  #     apiVersion: 0.0.6
+  #     language: wasm/assemblyscript
+  #     file: ./src/mappings/RocketPool.ts
+  #     entities:
+  #       - Token
+  #       - Pool
+  #       - Minipool
+  #       - PoolDailySnapshot
```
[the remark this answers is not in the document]
BenReilly19: keeping temporarily as may need to use those events. Will be either included /removed before finalizing PR.

subgraphs/rocket-pool/src/entities/pool.ts
```diff
@@ -0,0 +1,65 @@
...
+    pool.symbol = "RETH";
+    pool.protocol = protocol.id;
+    pool.createdTimestamp = blockTimestamp;
+    pool.createdBlockNumber = blockNumber;
+
+    // Tokens
+    pool.inputTokens = [
+      getOrCreateToken(Address.fromString(ETH_ADDRESS), blockNumber).id,
+      getOrCreateToken(Address.fromString(RPL_ADDRESS), blockNumber).id,
+    ];
```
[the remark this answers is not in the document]
BenReilly19: Modified and fixed

subgraphs/rocket-pool/src/entityUpdates/financialMetrics.ts
BenReilly19: swapped

subgraphs/rocket-pool/src/mappings/RocketPool.ts
BenReilly19: Just my manic scratchwork. Will delete XD

subgraphs/rocket-pool/networks.json
```diff
@@ -0,0 +1,7 @@
+{
+  "mainnet": {
+    "rocketpool": {
+      "address": "0x4D05E3d48a938db4b7a9A59A802D5b45011BDe58"
+    }
+  }
+}
```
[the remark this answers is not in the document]
BenReilly19: Actually not super familiar with deployment frameworks / what this is used for, so I'd defer to you for guidance here. Whatever makes the most sense. @jaimehgb

subgraphs/rocket-pool/src/entities/token.ts
```diff
@@ -0,0 +1,121 @@
...
+  let rethpriceUSD: BigDecimal | null = null;
+  if (tokenAddress == Address.fromString(RETH_ADDRESS)) {
+    rethpriceUSD = getRETHpriceUSD(
+      Address.fromString(RETH_ADDRESS),
+      blockNumber
+    );
+    log.warning("[getOrCreateToken]rethPriceUSD call result: {} ", [
+      rethpriceUSD.toString(),
+    ]);
+  }
```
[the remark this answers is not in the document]
BenReilly19: Yep. Early save and return added here.

subgraphs/rocket-pool/src/entities/token.ts
BenReilly19: adding new return earlier makes this call necessary for non-RETH tokens.

subgraphs/rocket-pool/networks.json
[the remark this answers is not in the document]
jaimehgb: The way it is done in most other subgraphs is by having the yaml being a template, and then filling the contract addresses and start blocks from a config.json file that is different for each network.

Take a look at badgerdao's subgraph for example (only has 1 deployment): 
- `/config/deployments/badgerdao-ethereum/configurations.json` -> values used to fill the template
- `/config/deployments/templates/badgerdao.template.yaml` -> template from which we build the yaml
- `package.json - "prepare:yaml" script` -> creates the yaml from the two previous files

I imagine rocketpool won't be deploying its contracts to other networks, so the "only" reason to templatize the yaml would be to keep the consistency with the rest of subgraphs. I would do it :D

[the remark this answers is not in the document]
jaimehgb: Also, the folder structure has varied slightly recently. So instead of `/config/networks/ethereum` it will be `/config/deployments/rocket-pool-ethereum/configurations.json`.

You can copy the `prepare:yaml` script from the same badgerdao too, should work as is

subgraphs/rocket-pool/src/entities/token.ts
BenReilly19: Have just modified this code to return early if same blockNumber. Smart workaround! Thx

dmelotik:
@BenReilly19 this could be renamed to something like `feat(#major); rocket pool; adding rocket pool subgraph`

BenReilly19:
Current issue - subgraph fails on handleBalanceUpdate(event: BalancesUpdated), which is integral to revenue calculation logic. Not sure if my approach is inherently flawed but believe code should function properly. @jaimehgb

jaimehgb:
> Current issue - subgraph fails on handleBalanceUpdate(event: BalancesUpdated), which is integral to revenue calculation logic. Not sure if my approach is inherently flawed but believe code should function properly. @jaimehgb

Do you have a link to your deployment? So I can see the exact issue

BenReilly19:
https://thegraph.com/hosted-service/subgraph/benreilly19/rocketpool?version=pending

jaimehgb:
> https://thegraph.com/hosted-service/subgraph/benreilly19/rocketpool?version=pending @jaimehgb

@BenReilly19 interesting error you got there 😅 
Take a look to see if any of these might apply to you:
img

https://discord.com/channels/953684103012683796/953685205531631637/1026978150174957738

PR #955, opened 2022-09-07
fix(#patch); erc721-metadata; fix JSON decoding for invalid/semi-valid NFT metadata

mrbrianhobo:
subgraph status (grafting from block `13003873`): https://okgraph.xyz/?q=messari%2Ferc721-metadata

- fixes stalled deployment at block 13003874
- add better warning logs for invalid JSON metadata
- updated JSON object check in `valueToObject` to filter out invalid ERC721 metadata but process semi-valid metadata like example described below
- faulty txn: https://etherscan.io/tx/0xfc97c1f1a72fd77b39b1e5cb429311c13fde43d28310d3b8aab2d768457cddbb
- faulty token uri: https://cloudflare-ipfs.com/ipfs/QmPYJRwT2JZvJm34RbQQfz3wJ6QcuhqdUfX6Ktsc1eewAa/1334.json
  - reason: original metadata JSON object wrapped in a JSON array (still valid JSON but triggers an error in subgraph)


subgraphs/erc721-metadata/src/mappings/token.ts
```diff
@@ -32,29 +31,29 @@ export function normalize(strValue: string): string {
 }
 
 export function createToken(
+  event: Transfer,
```
dmelotik, line 34: It seems like it is not necessary to make this event specifically type `Transfer`. To be more generic you could use `ethereum.Event` for the type so it doesn't have to come from a transfer.

dmelotik (commented):
lfgtm 🚀 🌔 

2 things though:
- increment the version
- My comment about the event is a suggestion, so up to you

subgraphs/erc721-metadata/src/mappings/token.ts
mrbrianhobo: makes sense to me! will update that 👍 
also will increment the version in a separate PR (these metadata/holders subgraphs have no versioning mechanism currently)

mrbrianhobo:
@dmelotik thanks for the review! will update the `ethereum.Event` suggestion. also will increment the version in a separate PR (these metadata/holders subgraphs have no versioning mechanism currently)

this-username-is-taken (approved):
LGTM!

PR #959, opened 2022-09-08
docs (#minor); Uniswap Forks; Added README for each protocol. 

steegecs:
**Context:**
We wanted to add documentation for each protocol in the Uniswap Forks. I have added documents describing the methodology within each protocol specific folder.

### DIFF
--- a/subgraphs/uniswap-forks/README.md
+++ b/subgraphs/uniswap-forks/README.md
@@ -2,41 +2,22 @@
 
 ## Protocols:
 
-- Apeswap
-  - BSC, Matic
-    - The two chains use the code from uniswap V2. Additionally, there is a Masterchef and MasterchefV2 contract. BSC only uses the Masterchef contract, while MasterchefV2 is used on the Matic chain. The masterchef contracts are slightly different for Apeswap vs. Sushiswap in the naming of certain functions. It looks like Apeswap forked from Pancakeswap. This is why they each have a separate folder for Masterchef.
-- Quickswap
-  - Matic
-    - Uses the same codebase as Uniswap-V2 for the factory and pools stored in the uniswap-forks/src folder
-- Solarbeam
-  - Moonriver
-    - Uses the uniswap-v2 codebase as Uniswap V2 for factory and pools. Additionally, it uses a masterchef contract that is templated from Sushiswap's implementation.
-- Spiritswap
-  - Moonbeam
-    - Uses the uniswap-v2 codebase as Uniswap V2 for factory and pools. Additionally, it uses a masterchef contract that is templated from Sushiswap's implementation.
-- Sushiswap
-  - Mainnet
-    - The only code that differs between uniswap is the use of the masterchef contract. This contract allows users to deposit LP tokens to earn rewards. Mainnet usees both the Masterchef and MasterchefV2 contract for these purposes.
-  - Arbitrum, Celo, Fantom, Fuse, Matic, Moonbeam, Moonriver, Xdai
-    - Similarly, these chains inherit the same code from uniswap. However, these chains only use the MiniChef contract for staking LP and earning rewards. It works very similar to the Masterchef contracts.
-  - BSC, Avalanche
-    - These two sushiswap chains deploy in just the same way as Uniswap V2. Just with different configurations.
-- Trader Joe
-  - Avalanche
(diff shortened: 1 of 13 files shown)


subgraphs/uniswap-forks/protocols/apeswap/config/networks/bsc/bsc.ts
```diff
@@ -63,7 +63,7 @@ export class ApeswapBscConfigurations implements Configurations {
     return FeeSwitch.ON;
   }
   getRewardIntervalType(): string {
-    return RewardIntervalType.TIMESTAMP;
+    return RewardIntervalType.BLOCK;
```
this-username-is-taken, line 66: why are we changing this? I thought these ones have been QAed

subgraphs/uniswap-forks/protocols/spiritswap/config/networks/fantom/fantom.ts
```diff
@@ -51,7 +51,7 @@ export class SpiritSwapFantomConfigurations implements Configurations {
     );
   }
   getTradeFee(): BigDecimal {
-    return BigDecimal.fromString("0.3");
+    return BigDecimal.fromString("0.18");
```
this-username-is-taken, line 54: do we need to redeploy this change?

subgraphs/uniswap-forks/protocols/vvs-finance/config/networks/cronos/cronos.ts
```diff
@@ -64,7 +64,7 @@ export class VSSFinanceCronosConfigurations implements Configurations {
     return FeeSwitch.ON;
   }
   getRewardIntervalType(): string {
-    return RewardIntervalType.NONE;
+    return RewardIntervalType.BLOCK;
```
this-username-is-taken, line 67: Same here

steegecs: It should have been BLOCK, but it does not need need to be re-deployed. There should be another option in the reward calculator. Currently, it does something like.... `If TIMESTAMP {handle like rewards are by timestamp};  else {handle like rewards are by block}. Therefore it will not change anything.

subgraphs/uniswap-forks/protocols/spiritswap/config/networks/fantom/fantom.ts
steegecs: No this was my bad. I meant to revert this change. When I was sifting through the documents, I thought it might be wrong. I need to look into this more tomorrow. In the docs it talked about this change being made with SpiritswapV2.

subgraphs/uniswap-forks/protocols/apeswap/config/networks/bsc/bsc.ts
steegecs: I wanted to review this with the protocol specialists. It was my bad for including this in this PR. When I was looking over the documents, it looked like the intervals needed to be flipped for BSC and Polygon. I was unsure though since it passed QA.

steegecs: I am following up on this.

steegecs: BANANA token is not even being priced on BSC and it looks like the pricing on Polygon is way off. The current price from the subgraph is 10 cents while the lowest price I see on coingecko is not less than 60 cents. If I understand how APR is calculated, I think this should lead to issues as well.

subgraphs/uniswap-forks/protocols/vvs-finance/config/networks/cronos/cronos.ts
steegecs: SpritiswapV2 has been out for about a month now with a different set of contracts. So this change is irrelevant - though, we might want to consider indexing this one!

subgraphs/uniswap-forks/protocols/spiritswap/config/networks/fantom/fantom.ts
steegecs: SpritiswapV2 has been out for about a month now with a different set of contracts. So this change is irrelevant - though, we might want to consider indexing this one!

subgraphs/uniswap-forks/protocols/apeswap/config/networks/bsc/bsc.ts
steegecs: Fixing in PR #964.

PR #962, opened 2022-09-09
Lending Schema: Include MarketOracle Entity

Data-Nexus:
Initial Proposal to include a MarketOracle entity. This can be used to determine the conversion rate of a market's tokens as per how the protocol also determines the conversion.

Also, I added (immutable: true) to event level data since this should not be updated after `.save()` which should save some db space (I believe the block_range column in the table is omitted when immutable: true).

### DIFF
--- a/schema-lending.graphql
+++ b/schema-lending.graphql
@@ -1,5 +1,5 @@
 # Subgraph Schema: Lending Protocol
-# Version: 2.0.1
+# Version: 2.1.0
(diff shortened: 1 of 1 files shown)


Data-Nexus:
As I'm going through it seems like having an Oracle and TokenOracle entity is a better route than MarketOracle. Will report updates once I'm all the way through.

dmelotik:
Hey @Data-Nexus I am closing this given we addressed it in our latest upgrade https://github.com/messari/subgraphs/pull/1265

PR #965, opened 2022-09-09
refactor(#minor); Uniswap Forks; Refactor configurations

steegecs:
**Context:**
Many of the token addresses in Uniswap Forks were being converted to lower case using methods. Instead of using methods, I went through the configurations and manually converted token addresses to lower case.


this-username-is-taken:
This is nice. Two minor feedback:

1. I don't think we should turn token symbols to lower case (e.g. wGLMR is better than wglmr)
2. Don't forget to update versions

PR #972, opened 2022-09-11
Updated Methodology for Notional after discussion

SerenityFund:
This is updated after discussion with the subgraph writer. For records and review purposes, as there might be another PR pending from the writer as well.


this-username-is-taken:
Would you add a version to this (as well as the rest of methodology docs)? For example: https://github.com/messari/subgraphs/tree/master/subgraphs/lido

PR #974, opened 2022-09-12
Add Ribbon Finance Methodology

SerenityFund:


ishraq8 (approved):
Protocol is fairly straightforward and looks like a Dopex fork. Methodology is consistent with subgraph standard and can be added.

PR #976, opened 2022-09-12
feat(): Added radiant-capital, aave-v2 fork on arbitrum

jaimehgb:
Based this one off of Geist fork. I only made the following changes to it:
- Arbitrum contract calls are very slow, so rewards are calculated twice per day.
- Reward token price USD is fetched from SushiSwap RDNT/WETH, and WETH price from the protocol oracle.
- rToken (aToken) `getAssetPrice()` seems to return prices with 8 decimals, instead of 18.
- Individually fetching lending and borrowing rewards. From our code, Geist seems to always give 3x the lender reward to the borrower. In radiant this is not the case, and the lender/borrower reward ratio is arbitrary on each pool.

Syncing here: https://okgraph.xyz/?q=QmcsfyCzhV5gfcWUzi4MJGbsJyxF4MU5FqoBhY1hxtuDKu


subgraphs/aave-v2-forks/protocols/radiant/src/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+  "0x24704aff49645d32655a76df6d407e02d146dafc"; // RDNT/WETH
+export const RWETH_ADDRESS = "0x15b53d277af860f51c3e6843f8075007026bbb3a";
+
+/////////////////////////////
+///// Protocol Specific /////
+/////////////////////////////
+
+export namespace Protocol {
+  export const NAME = "Radiant Captital";
+  export const SLUG = "radiant-capital";
```
this-username-is-taken, line 19: we should have the slug consistent across the folder name and the deployment, so should be `radiant` instead

this-username-is-taken (approved):
LGTM!

this-username-is-taken:
you should update `deployment.json` as well?

subgraphs/aave-v2-forks/protocols/radiant/src/constants.ts
```diff
@@ -16,7 +16,7 @@ export const RWETH_ADDRESS = "0x15b53d277af860f51c3e6843f8075007026bbb3a";
 
 export namespace Protocol {
   export const NAME = "Radiant Captital";
-  export const SLUG = "radiant-capital";
+  export const SLUG = "radiant";
```
dmelotik, line 19: why not `radiant-capital`? https://www.coingecko.com/en/coins/radiant-capital

subgraphs/aave-v2-forks/schema.graphql
```diff
@@ -545,6 +545,9 @@ type Market @entity {
 
   " Aditional tokens that are given as reward for position in a protocol, usually in liquidity mining programs. e.g. SUSHI in the Onsen program, MATIC for Aave Polygon "
   rewardTokens: [RewardToken!]
+
+  " Aux: Timestamp of the last time rewards were calculated for this market. "
+  _lastRewardsUpdated: BigInt
```
dmelotik, line 550: this is the standard I have been using in this directory: https://github.com/messari/subgraphs/blob/master/subgraphs/aave-v2-forks/schema.graphql#L317

subgraphs/aave-v2-forks/protocols/radiant/src/rewards.ts
```diff
@@ -0,0 +1,134 @@
...
+  let tryDepPoolInfo = incentiveController.try_poolInfo(
+    Address.fromString(market.outputToken!)
+  );
+  let tryBorPoolInfo = incentiveController.try_poolInfo(
+    Address.fromString(market.vToken!)
+  );
+  let tryTotalAllocPoint = incentiveController.try_totalAllocPoint();
+  let tryTotalRewardsPerSecond = incentiveController.try_rewardsPerSecond();
+
+  if (
```
dmelotik, line 52: I think these should be ORs. If not all are reverted it won't skip the function and could try to read value on a reverted contract call.

dmelotik (commented):
Nice job @jaimehgb ! 🚀 

One small general thing:
- The config directory will need to be updated according to steegecs udpate here: https://github.com/messari/subgraphs/commit/83b8bbaaf68e5c58be956acb34315e1d4f6fc677

subgraphs/aave-v2-forks/protocols/radiant/src/constants.ts
jaimehgb: The hosted service deployment was just `radiant-arbitrum`, so dropped the "capital" to make it consistent with that 😅 

It only takes a few days to sync, so can probably deploy it into a new one with the correct name.

subgraphs/aave-v2-forks/schema.graphql
jaimehgb: Oh, sounds good. Changed!

subgraphs/aave-v2-forks/protocols/radiant/src/rewards.ts
jaimehgb: Oh, good catch! Changing that 🚀

subgraphs/aave-v2-forks/schema.graphql
```diff
@@ -677,6 +674,9 @@ type Market @entity {
 
   " The state of the protocol prior to being paused (isActive / canUseAsCollateral / canBorrowFrom) "
   prePauseState: [Boolean!]!
+
+  " Aux: Timestamp of the last time rewards were calculated for this market. "
+  _lastRewardsUpdated: BigInt
```
dmelotik, line 679: Sorry, I also meant the no underscore. What are your thoughts?

subgraphs/aave-v2-forks/schema.graphql
dmelotik: I also meant to remove the underscore, what do you think?

jaimehgb: Hmm 🤔 
I added the underscore as that's how I've seen it done in other subgraphs. I think the underscore is convenient since you might not always be looking at where exactly the variable is defined in the schema. For example when you are querying the subgraph from the graphql web client, or trusting some autocomplete in your IDE.

I'll remove the underscore for now because having mixed names (some with and some without underscore) on the same subgraph might be even more confusing 😅  But I think we should use underscores or some other identifier in general, it doens't hurt and adds clarity.

dmelotik: I agree, not sure why I did that when building the aave forks. I will make a note for me to fix that after this is merged

dmelotik (approved):
approved 🥇

PR #978, opened 2022-09-12
fix (#minor); Uniswap V3 Celo; Updated reference token and reduced minimum liquidity threshold. 

steegecs:
**Conext:** 
There were some minor pricing issues in the Uniswap V3 Celo. I am fixing this by making some small changes to the configurations with the reference token and minimum liquidity threshold. This is in response to a data issue brought to my attention from @bye43.

- https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/uniswap-v3-celo&tab=pool&poolId=0xd88d5f9e6c10e6febc9296a454f6c2589b1e8fae

### DIFF
--- a/deployment/deployment.dev.json
+++ b/deployment/deployment.dev.json
@@ -3178,7 +3178,7 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.1.6",
+          "subgraph": "1.1.7",
           "methodology": "1.0.0"
         },
         "deployment-ids": {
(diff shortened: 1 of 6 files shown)


jaimehgb (dismissed):
Looks ok to me.

How low can we go with minimum liquidity thresholds for pricing? Should that be part of the `methodology` at some point? If prices are not set because liquidity doesn't hit the threshold, isn't that intended?

steegecs:
> Looks ok to me.
> 
> How low can we go with minimum liquidity thresholds for pricing? Should that be part of the `methodology` at some point? If prices are not set because liquidity doesn't hit the threshold, isn't that intended?

Yeah that makes sense to add it to the methodology. In fact, if we are displaying our methodologies, It'd probably be a good idea to do that sooner than later. Unfortunately, the method is not too standard. Specifically with Uniswap-v3, it seems that when there is more volume, I need to raise the threshold to avoid outliers.

jaimehgb (dismissed):

jaimehgb (approved):

Issue #952, opened 2022-09-07
Ubeswap Celo QA


SerenityFund:
## Review Info

| Description | Info | 
| :-------- | :--- | 
|Sugraph Reviewed | https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/ubeswap-celo&tab=protocol |
| Schema Version  |  1.3.0  | 
| Subgraph Version |  1.1.5  | 
| Methodology Version  |	1.0.0 |
| Evidence Spreadsheet |  https://docs.google.com/spreadsheets/d/1QNiYnJ00Q6Sow1qok81XeZvymN2jvZsC/edit?usp=sharing&ouid=104027705798814439445&rtpof=true&sd=true |

### Pool Overview TVL
 - Only two out of 30 pools have shown large discrepancy of around 50% in its TVL data. These pools are the 13th pool (MOO/mCELO) and 26th pool (cETH/mCUSD). 
 - cGLD ticker must be changed to CELO ticker. They are the same but it's confusing.
 - Refer to the sheet "Pool_Overview" for details

##  Pool Section Error
- Six out of the first 10 pools are displaying issues in the 'Pool' Section. It says failed to fetch.
-  The pools are 2nd pool (cGLD/mCUSD), 3rd pool (WETH/mCUSD), 5th pool (cGLD/mCUSD), 7th pool  (sCELO/cGLD), 9th pool (mCUSD/WBTC), and 10th pool (cGLD/mCEUR).
- Refer to the sheet "Pool_vault schema" for details

## Pool_snapshots
- The reward APRs of the first (mCUSD, Mceur) & fourth  (UBE, cGLD) pools are outdated and they both have the same percentage data which is 7.73%. 
- Please refer to the sheet 'Reward-APR-Support' for more details

### Editorials  There are some differences we have noticed and are not necessarily data errors.

- Protocol TVL
TVL data is in line with data from Defi Llama in the past five days. However, as compared against official website, there has been 3% discrepancy in the past four days before the latest day. Additionally there have been two period wherein the discrepancy was ranging from 8% to 11%. This could be due to some individual pools having big discrepancies.

Please refer to the attached worksheet for details.

steegecs:
Are we sure that we want to change the cGLD ticker to CELO? This appears to be the proper token symbol and is the same one used by Coinbase. And if we do want to change it, I propose we update but re-deploy when there is a new change that required re-deployment.

steegecs:
[...]

PR #980, opened 2022-09-12
fix (#minor); Ubeswap; Fix in reponse to issue #952

steegecs:
**Context:** 
Response to issue #952 

Screen Shot 2022-09-12 at 1 02 06 PM

This fix is supposed to address this issue by extending the token whitelist. This should price in more pools and allow the TVL in our subgraph to catchup to defillama.


deployment/deployment.dev.json
```diff
@@ -0,0 +1,3634 @@
+{
```
mrbrianhobo, line 1: is this file supposed to be included?

PR #981, opened 2022-09-12
fix (#minor); Solarbeam; Added xcksm to whitelist

steegecs:
**Context:**
Response to issue #547.

Outputting 0 TVL
- Solarbeam Wrapped liquid staked KSM/xcKSM - 0x5568872bc43bae3757f697c0e1b241b62eddcc17

Needed to add xcKSM to the token whitelist.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2813,7 +2813,7 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.1.4",
+          "subgraph": "1.1.5",
           "methodology": "1.0.0"
         },
         "deployment-ids": {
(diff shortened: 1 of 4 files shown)


deployment/deployment.dev.json
```diff
@@ -0,0 +1,3634 @@
+{
```
mrbrianhobo, line 1: is this file supposed to be apart of this change?

mrbrianhobo (approved):
looks good to me!

deployment/deployment.dev.json
steegecs: Yeah - I am not sure how that ended up there. For some reason I added the deployment.dev.json back in, but we do not want it.

mrbrianhobo (approved):
LGTM

PR #989, opened 2022-09-13
fix(#patch): uniswap-forks; fixed breaking change preventing grafts

jaimehgb:
I recently introduced a breaking change by adding a required field to MasterChef entity in the schema. This field was not being set before and makes grafting from a previous version impossible:

Screenshot 2022-09-13 at 17 10 41

I've made the field optional, and when fetching the entity, if it is not set, it will be set and saved. This way we can still ensure it will always be present while making the schema compatible.

I've also updated the trader-joe version (the one I was trying to graft :D )

### DIFF
--- a/subgraphs/uniswap-forks/src/common/masterchef/helpers.ts
+++ b/subgraphs/uniswap-forks/src/common/masterchef/helpers.ts
@@ -42,17 +42,26 @@ export function getOrCreateMasterChef(
 ): _MasterChef {
   let masterChef = _MasterChef.load(masterChefType);
 
-  if (!masterChef) {
-    masterChef = new _MasterChef(masterChefType);
+  if (masterChef) {
+    if (masterChef.address) {
+      return masterChef;
+    }
+
     masterChef.address = event.address.toHexString();
-    masterChef.totalAllocPoint = BIGINT_ZERO;
-    masterChef.rewardTokenInterval = NetworkConfigs.getRewardIntervalType();
-    masterChef.rewardTokenRate = NetworkConfigs.getRewardTokenRate();
-    log.warning("MasterChef Type: " + masterChefType, []);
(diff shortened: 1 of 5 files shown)


subgraphs/uniswap-forks/src/common/masterchef/helpers.ts
```diff
@@ -42,17 +42,26 @@ export function getOrCreateMasterChef(
 ): _MasterChef {
   let masterChef = _MasterChef.load(masterChefType);
 
-  if (!masterChef) {
-    masterChef = new _MasterChef(masterChefType);
+  if (masterChef) {
+    if (masterChef.address) {
```
steegecs, line 46: Why do it like this and not just `if (masterchef)`

jaimehgb: `if(masterchef)` doesn't guarantee that the `address` field is set.

The idea is:
if masterchef exists and the address is set, return as is
if it exists but the address is not set, set it and save, then return
if it doesn't exist, create

steegecs: So is there an instance where you would be creating a masterchef entity but not add the address? If not, then isn't it safe to assume that the address is there?

jaimehgb: For subgraphs that are already deployed, that already have the masterchef instance created but it happened before the address was added.

Shouldn't be a problem for subgraphs that sync from scratch, but it will for the ones that are grafted from the old schema

steegecs: Oh I get. Thanks for explaining!

steegecs (approved):

PR #990, opened 2022-09-13
fix(#patch); rari fuse; FLOAT exploit

dmelotik:
$FLOAT price oracle was exploited at this transaction https://etherscan.io/tx/0x40db7bd89d2a3f7df2793ba4f5be9a2ca93463d6bb6af024e5cd1b73ff827248 and to fix this we use our own price oracle to get the price at that block. Increment the version to `1.1.4`

Testing: https://okgraph.xyz/?q=dmelotik%2Ffuse-mainnet


0xbe1 (approved):
lgtm

Issue #914, opened 2022-09-01
Rari Fuse Mainnet QA


SerenityFund:
## Review Info

| Description | Info | 
| :-------- | :--- | 
| Subgraph Reviewed |  https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/rari-fuse-ethereum&tab=protocol  |
| Schema Version  |  2.0.1  | 
| Subgraph Version |  1.1.3  | 
| Methodology Version  |	1.0.0 |
| Evidence Spreadsheet |  https://docs.google.com/spreadsheets/d/1Qd_4aTM2ShSjIQ7SAnS4PYMSxlj1eEVe/edit?usp=sharing&ouid=104027705798814439445&rtpof=true&sd=true  |

###  Protocol TVL
As compared to official website data, both subgraph and Defi Llama data show a large discrepancy of around 50%. This maybe due to the inactive pools that are still being tracked by the subgraph and Defi Llama.

| Subgraph | Defi Llama | Absolute Difference | % Difference | Official Website | Absolute Difference | % Difference | 
| :---- | :--- | :---  | :---  | :---  | :---  | :---  | 
| 196,250,765 | 192,916,184 | 3,334,581 | 1.73% | 127,067,102 | 69,183,662 | 54.45% |

### Pool Overview TVL

 - 6 out of top 20 pools in the subgraph data are not shown in the official website.
 - Out of 14 pools that are shown in the official website, only two pools have large discrepancies compared to subgraph data ( The 11th and 20th pools)
- Only one out of the six missing pools have accurate TVL subgraph data as compared against to DeBank data.
- Please refer to the sheet "Pool_Overview" for details

### TotalBorrowBalance
We have computed the revenue implied by the data from official website on 30 August 2022 and this shows large discrepancy with subgraph data. Refer to the sheet "Support-BorrowBalance". This is maybe due to the reason that some of the pools that are being tracked by the subgraph are not displayed on the official website.

|Subgraph	|Official Website|	Absolute Difference 	|% Difference |
| :---- | :--- | :---  | :---  |
|211,624,071	| 91,671,114	|119,952,957	|130.85%|

### Rates

The interest rates in subgraph is slightly off compared to the official website. E.g. for Ferrari Pool, 

|  |Subgraph	|Official Website|
| :---- | :---- | :--- |
| Borrowing APR | 2.86%	| 2.96% |

[...]

PR #995, opened 2022-09-14
fix(#patch); rari fuse; FLOAT exploit fix

dmelotik:
fix FLOAT exploit crazy high dailyDeposit, and noted the event in the readme.

Linking #914


dmelotik:
dupe, nvm

PR #1000, opened 2022-09-14
patch(#minor); Uniswap V3; Updated usage metrics for swaps, and removed unnecessary log

steegecs:
**Context:**
Usage metrics are pretty inconsistent with defillama. This is because the usage on the swaps are being recorded incorrectly. This PR fixes that by updating the user for usage metrics. Also, there is the remove of an unnecessary log in this PR.

### DIFF
--- a/subgraphs/uniswap-v3/src/mappings/core.ts
+++ b/subgraphs/uniswap-v3/src/mappings/core.ts
@@ -1,4 +1,5 @@
 // import { log } from '@graphprotocol/graph-ts'
+import { log } from "@graphprotocol/graph-ts";
 import {
   Burn as BurnEvent,
   Initialize,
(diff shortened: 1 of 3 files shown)


subgraphs/uniswap-v3/src/mappings/core.ts
```diff
@@ -1,4 +1,5 @@
 // import { log } from '@graphprotocol/graph-ts'
+import { log } from "@graphprotocol/graph-ts";
```
harsh9200, line 2: nit: remove the commented part

steegecs: Removed.

harsh9200 (approved):

PR #1004, opened 2022-09-14
patch(#minor); Uniswap V3; Updated protocol getter to update versions on get.

steegecs:
Context:
Relate to PR #1003 

The protocol entity gets created at the beginning of indexing, and when you graft on top, even though you change the version, the protocol entity still displays the older version. This will happen unless the versions get reset in the protocol getters. This fix sets the version after each get.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2410,7 +2410,7 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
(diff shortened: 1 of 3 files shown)


dmelotik:
lgtm, but you should also add this in the docs somewhere. This is a good practice for us!

dmelotik (approved):

PR #1008, opened 2022-09-15
fix(#patch); compound forks; add transfer handler

dmelotik:
Adding the transfer handler allows us to properly track all positions in the protocol and not miss any.

incremented patch version of every compound-fork

This is not going to be a change for protocol metrics for mainnet. However, this will affect jacob anderson's whale watchers data app for mainnet.

### DIFF
--- a/subgraphs/compound-forks/src/mapping.ts
+++ b/subgraphs/compound-forks/src/mapping.ts
@@ -116,7 +116,7 @@ export function _handleNewCollateralFactor(
   newCollateralFactorMantissa: BigInt
 ): void {
   let market = Market.load(marketID);
-  if (market == null) {
+  if (!market) {
     log.warning("[handleNewCollateralFactor] Market not found: {}", [marketID]);
     return;
   }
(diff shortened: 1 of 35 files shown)


subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1056,6 +1062,93 @@ export function _handleNewReserveFactor(
...
 
+export function _handleTransfer(
+  event: ethereum.Event,
+  marketID: string,
+  to: Address,
+  from: Address,
+  comptrollerAddr: Address
+): void {
+  let protocol = LendingProtocol.load(comptrollerAddr.toHexString());
+  if (protocol == null) {
```
0xbe1, line 1073: nit: be consistent with the way we null check. either `!xxx` or `xxx == null`

here we `== null` to null check, but later on `fromAccount` and `toAccount` we do `!xxx` :P

0xbe1 (approved):
lgtm. very clean. look forward to seeing it syncs

PR #1012, opened 2022-09-15
refactor (#minor); Updated the directory structure for all subgraphs and configured deployment scripts to be compliant

steegecs:
**Context:**
The purpose of this PR is to configure our directories to be more consistent with the new deployment.json. This required a small refactor to the deployment scripts and updates to the package.json files. 

- [x] Update directory structure of all subgraphs
- [x] Updated deployment scripts to be compliant with new structure
- [x] Updated all package.json to contain compliant scripts.


jaimehgb (approved):
This is an example of a giant PR but that is easy to review 😅

PR #1013, opened 2022-09-16
fix(#patch); cream eth; deprecate snapshots

dmelotik:
deprecate snapshots as well to ensure proper viewing on protocol metrics FE

Grafting: https://okgraph.xyz/?q=dmelotik%2Fcream-finance-ethereum


0xbe1 (approved):
LGTM

PR #1015, opened 2022-09-16
fix(#patch); euler; fix CTX price mismatch

dmelotik:
We saw a price issue in CTX (Cryptex) on block `15358330` which was making dailyBorrowUSD artificially high.

Fix: take the price and hardcode from coingecko at that exact block

### DIFF
--- a/subgraphs/euler-finance/src/mappings/handlers.ts
+++ b/subgraphs/euler-finance/src/mappings/handlers.ts
@@ -144,8 +144,8 @@ function updateProtocolAndMarkets(block: ethereum.Block): void {
   let protocolUtility = getOrCreateProtocolUtility(blockNumber);
   let markets = protocolUtility.markets;
 
-  if (protocolUtility.lastBlockNumber >= blockNumber ) {
-    return; // Do this update every block
+  if (protocolUtility.lastBlockNumber <= blockNumber - 120 ) {
+    return; // Do this update every 120 blocks
   }
 
   let eulerViewQueryResult = queryEulerGeneralView(markets, block);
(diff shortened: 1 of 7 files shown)


subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -144,8 +144,8 @@ function updateProtocolAndMarkets(block: ethereum.Block): void {
   let protocolUtility = getOrCreateProtocolUtility(blockNumber);
   let markets = protocolUtility.markets;
 
-  if (protocolUtility.lastBlockNumber >= blockNumber ) {
-    return; // Do this update every block
+  if (protocolUtility.lastBlockNumber >= blockNumber - 120 ) {
+    return; // Do this update every 120 blocks
```
mrbrianhobo, line 148: how come every 120 blocks? maybe include some reasoning/methodology so it doesn't seem like its an arbitrary number pulled out of nowhere

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -89,7 +90,17 @@ export function createBorrow(event: Borrow): BigDecimal {
   borrow.to = marketId;
   borrow.amount = event.params.amount;
   const marketUtility = getOrCreateMarketUtility(marketId);
-  borrow.amountUSD = amountToUsd(borrow.amount, marketUtility.twap, marketUtility.twapPrice);
+  
+  // catch CRYPTEX outlier price at block 15358330
+  // see transaction: https://etherscan.io/tx/0x77885d38a6c496fdc39675f57185ab8bb11e8d1f14eb9f4a536fc1c4d24d84d2
+  if (marketId.toLowerCase() == CRYPTEX_MARKET_ID.toLowerCase() && event.block.number.equals(BigInt.fromI32(15358330))) {
+    // this is the price of CTX on August 17, 2022 at 11AM UTC-0
```
mrbrianhobo, line 97: maybe leave a reference/source where you got this price

mrbrianhobo (approved):
looks good but probably best to leave a reference/reason for hard coded values 😎

mrbrianhobo (approved):
nice! 🚀

PR #1020, opened 2022-09-16
fix(#patch); Balancer v2; scaling factor

harsh9200:
**Context**:
Many of the pools in Balancer v2(Mainnet) have dynamic weights and we update these pools weights based on the scaling factors of the pool but In some pools the scaling factor also includes the factor corresponding to the output token which leads to wrong weights. 

**Fixes**:
- Use `getScalingFactor(Address)` instead of `getScalingFactors()` for updating pool weights.
- Updated Price Module.
- Harddcoded the following missing/wrong priced stable coins:
```
  "0x5f98805a4e8be255a32880fdec7f6728c6568ba0", // LUSD Token
  "0x8f4063446f5011bc1c9f79a819efe87776f23704", // bb-f-DAI
  "0xb0f75e97a114a4eb4a425edc48990e6760726709", // bb-f-LUSD
  "0xc8c79fcd0e859e7ec81118e91ce8e4379a481ee6", // bb-f-fei
  "0xd997f35c9b1281b82c8928039d14cddab5e13c20", // bb-f-USD
  "0x2bbf681cc4eb09218bee85ea2a5d3d13fa40fc0c",, // bb-a-USDT
  "0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb2", // bb-a-USD
  "0x804cdb9116a10bb78768d3252355a1b18067bf8f", // bb-a-DAI
  "0x9210f1204b5a24742eba12f710636d76240df3d0", // bb-a-USDC
  "0x4fd63966879300cafafbb35d157dc5229278ed23", // bb-a-USD
  "0x652d486b80c461c397b0d95612a404da936f3db3", // bb-a-USDC
  "0xa3823e50f20982656557a4a6a9c06ba5467ae908", // bb-a-DAI
  "0xe6bcc79f328eec93d4ec8f7ed35534d9ab549faa" // bb-a-USDT
```


steegecs (commented):
Have the oracle changes already been validated? 

Also, it seems like these are three distinct changes/improvements you are making. Do you think these changes  depend on each other? If not, in the future, they may be better off in separate PRs.

harsh9200:
> Have the oracle changes already been validated?
yes, the price lib changes have been validated.

> Also, it seems like these are three distinct changes/improvements you are making. Do you think these changes depend on each other? If not, in the future, they may be better off in separate PRs.
I will keep that in mind.

steegecs:
Can you rebase one more time and push? Updated the workflow runner, and I think it will make the workflow action pass.

steegecs (commented):
These updates to the swap fee are also better put in a new PR. It's hard to keep track of everything that is changing and for what reason. 

Why are you multiplying the protocol fees by the swap fee here? Aren't protocol fees from swaps usually calculated something like {swapFee.minus(supplySideFee).times(BIGDECIMAL_HUNDRED)}?

harsh9200:
In the last commit, only the `protocolFees` is changed as it should be calculated against the total trading fees (aka `swapFees`), the revenue logic is still the same as before. [Ref](https://cdn.discordapp.com/attachments/964582481099632642/1021455553706332341/unknown.png)

this-username-is-taken (approved):
Great work!

PR #1022, opened 2022-09-18
feat (#minor); Eslint; Add eslint dependencies, lint-staged, and lint configs for subgraphs.

steegecs:
**Context:**
Below is a list of rules for eslint curate by @jaimehgb and I. The rules we chose were curated from recommended rules from the TS eslint plugin. Prettier options are turned off. All of the rules listed below may not exhaust all rules applied, but they are all the rules that caused issues when lint checking the whole subgraph repository. There were a couple of rules turned off from the TS config

- Rule: @typescript-eslint/no-non-null-assertion - Disallowing the non-null assertion -> '!"
- Rule: prefer-const - Make vars const if not mutated.
- Rule: @typescript-eslint/no-unused-vars - Disallow unused variables
- Rule: no-constant-condition - Disallow the use of constants in a condition. Example of this is --> while (true)  instead of while (is_true)
- Rule: no-fallthrough - Forces fall throughs on switch to be explicit
- Rule: @typescript-eslint/ban-ts-comment - Do not use "@ts-ignore" because it alters compilation errors
- Rule: @typescript-eslint/no-empty-function - Disallow empty functions. A function will not be considered a problem if it contains a comment.
Not even a rule just expected - Parsing error - It did not like me mustaching on a typescript file
- Rule: no-self-assign - Prevent self assignment of variables - i.e. let obj = obj.
- Rule: @typescript-eslint/ban-types - It was banning BigInt, because its not a native typescript type, but its a graph-ts library type so I unbanned it. Also, it does not like using the String type -> Instead just use string literals.
- Rule: @typescript-eslint/no-loss-of-precision - Catch variable overflows that could lead to precision errors.


package.json
```diff
@@ -4,10 +4,25 @@
...
+    "pretty-quick": "^3.1.3",
+    "lint-staged": "^13.0.3",
+    "prettier": "^2.7.1"
+  },
+  "lint-staged": {
+    "*.ts": "eslint --cache --fix",
+    "*.{ts,css,md}": "prettier --write"
   },
   "scripts": {
+    "prettier": "prettier --loglevel silent --write \"**/*.{md,jsx,tsx,ts,json,html,css,js,yml}\"",
```
dmelotik, line 22: should the last type read `yaml`?

dmelotik:
@steegecs this is cool! I have some questions.

> Rule: @typescript-eslint/no-unused-vars - Disallow unused variables
does this include unused imports and unused functions?

Looks like this also adds a prettier command for the entire repo. Should we prevent this from going into `./apps` and `./dashboard`?

After this is merged we can also remove any custom prettier configs for individual subgraphs.

package.json
steegecs: Hmm, I actually think I should remove this since we have not been using prettier on yaml. But, if we wanted it, we would probably want to add `.yml` and `.yaml` since they are used interchangeably.

steegecs:
> > Rule: @typescript-eslint/no-unused-vars - Disallow unused variables
> 
> does this include unused imports and unused functions?

This is just unused variables

steegecs:
> Looks like this also adds a prettier command for the entire repo. Should we prevent this from going into `./apps` and `./dashboard`?

I have re-factored and should be set up so it only applies to our subgraphs repo.

steegecs:
> Also does the linter check other directories other than `./subgraphs`?

Yes - just the subgraphs.

steegecs:
> After this is merged we can also remove any custom prettier configs for individual subgraphs.

Yes! We can do that. And also add a prettier config at the head of repo or in the subgraphs directory.

jaimehgb (approved):
Ship it 🚀

jaimehgb (approved):

PR #1029, opened 2022-09-19
fix(): geist-finance folder new structure

jaimehgb:
Geist finance was still with the old folder structure, so changes to aave-v2-forks subgraph fail to build


steegecs (approved):
Looks good. Thanks!

PR #1032, opened 2022-09-19
docs(); truefi; add truefi to deployment files

dmelotik:
added truefi to deployment files, as we need it for protocol metrics since it is QA'd


steegecs (approved):
Looks good!

PR #1035, opened 2022-09-19
feat(#minor): validate subgraph folder structure CI action

jaimehgb:
The build and deploy scripts expect a certain directory structure. In case some subgraph doesn't follow that structure, those scripts can either fail, or depending on how it is structured, succeed after omitting the build.

What do you guys think about enforcing that the structure is followed? @steegecs @dmelotik @this-username-is-taken @0xbe1 @harsh9200  
It might be a bit aggressive, but it makes it clear how we expect subgraphs' code to be organized, will allow us to catch wrong structures faster if we change them again and miss any, and the deployment scripts can expect files and paths to be a certain way.

This PR adds (`folderslint`)[https://github.com/denisraslov/folderslint] (the only decent tool I've found that does this easily). It defines a list of acceptable paths for subgraphs and their files. I would have liked to be able to not only whitelist paths but also require them, but `folderslint` is very simple.

Also fixes some subgraphs with the legacy structure.

Also, took the freedom to rename the `Deployment` CI job to `Build`, as I think it is not deploying anything. But let me know if needs to be changed back :P


steegecs:
> The build and deploy scripts expect a certain directory structure. In case some subgraph doesn't follow that structure, those scripts can either fail, or depending on how it is structured, succeed after omitting the build.
> 
> What do you guys think about enforcing that the structure is followed? @steegecs @dmelotik @this-username-is-taken @0xbe1 @harsh9200 It might be a bit aggressive, but it makes it clear how we expect subgraphs' code to be organized, will allow us to catch wrong structures faster if we change them again and miss any, and the deployment scripts can expect files and paths to be a certain way.
> 
> This PR adds (`folderslint`)[https://github.com/denisraslov/folderslint] (the only decent tool I've found that does this easily). It defines a list of acceptable paths for subgraphs and their files. I would have liked to be able to not only whitelist paths but also require them, but `folderslint` is very simple.
> 
> Also fixes some subgraphs with the legacy structure.
> 
> Also, took the freedom to rename the `Deployment` CI job to `Build`, as I think it is not deploying anything. But let me know if needs to be changed back :P

steegecs:
Yeah

subgraphs/.folderslintrc
```diff
@@ -0,0 +1,28 @@
...
+
+    "### generic subgraph structure ###",
+    "*/abis/**",
+    "*/protocols/*/config/deployments/*",
+    "*/protocols/*/config/templates/*",
+    "*/src/**",
+
+    "### uniswap forks structure ###",
+    "*/configurations/**",
+    "*/protocols/*/config/deployments/**",
```
steegecs, line 17: Will lint fail here without the extra asterisk?

jaimehgb: It doesn't! Removed :D

jaimehgb:
> What is the purpose of including the outliers here, though. For keeping track of folder dissimilarities? Will the lint fail if you remove them?

Yes. If a folder is found that doesn't match any of those patterns it errors. Since I don't know what they exactly contain and don't want to generalize that structure (at least just yet) I decided to whitelist them individually. They alll mainly seem to contain tooling and stuff. We could normalize the folder name where to contain that.

> Also, is it necessary to have generated, build, and node modules? Do you think it's better to require these rather than just omit them since they don't need to be pushed to the repo?

Hmm, good point. I added them because they broke when running them locally. But it is true that at the CI pipelines they won't be present. It would be ideal if this tool ignored the errors from files in the `.gitignore` or something. Answering the question, if people are going to be running `folderslint` locally before pushing it shouldn't fail for those files. So I think I would leave them whitelisted. Some alternatives could be:
- Parsing the error outputted and ignore it if files are in the gitignore
- Submitting a PR to fodlerslint proposing it or forking it
- ?

steegecs:
> Hmm, good point. I added them because they broke when running them locally. But it is true that at the CI pipelines they won't be present. It would be ideal if this tool ignored the errors from files in the `.gitignore` or something. Answering the question, if people are going to be running `folderslint` locally before pushing it shouldn't fail for those files. So I think I would leave them whitelisted. Some alternatives could be:
> 
> * Parsing the error outputted and ignore it if files are in the gitignore
> * Submitting a PR to fodlerslint proposing it or forking it
> * ?

Having them whitelisted is fine with me then! Probably the easiest way to handle this, and it should not cause any issues.

subgraphs/.folderslintrc
```diff
@@ -0,0 +1,31 @@
...
+    "### generic subgraph structure ###",
+    "*/abis/**",
+    "*/protocols/*/config/deployments/*",
+    "*/protocols/*/config/templates/",
+    "*/protocols/*/src/**",
+    "*/src/**",
+
+    "### uniswap forks structure ###",
+    "uniswap-forks/configurations/*",
+    "uniswap-forks/protocols/*/config/deployments/*",
```
steegecs, line 18: Can't you remove these then too since they are redundant?

jaimehgb: Good catch 💪

steegecs (approved):

this-username-is-taken:
Thanks @jaimehgb! This is great. Should we also add some documentation about folder structure? Maybe to here: https://github.com/messari/subgraphs/blob/master/docs/Contributing.md

this-username-is-taken (approved):

jaimehgb:
Merging this 🚀 
Hope it goes well 🤞 Will be monitoring it closely, but let me know if it breaks for any of you guys on your PRs

PR #1039, opened 2022-09-20
docs(); Updated README to include comments on best practices for protocol version updates. 

steegecs:
Context:
Relate to PR https://github.com/messari/subgraphs/pull/1003 and https://github.com/messari/subgraphs/pull/1004

The protocol entity gets created at the beginning of indexing, and when you graft on top, even though you change the version, the protocol entity still displays the older version. This will happen unless the versions get reset in the protocol getters. This fix sets the version after each get.


dmelotik (approved):
Good note! lgtm 🚀

PR #1046, opened 2022-09-23
feat(#major); cli tool; add deployment cli tool

dmelotik:
What is this for?
- when deploying from the command line it can be tedious to figure out exactly which commands to call and which flags to include.

This cli tool makes it really user friendly to deploy. The tool simply creates the cli args needed to deploy based on a couple inputs and selections from the user.

TODO: 
- [ ] make it simple by only choosing to deploy one protocol at a time
- [ ] add video of using it to docs


dmelotik:
accomplished more seamlessly in #1124

PR #1060, opened 2022-09-26
fix(#patch): fixed trader joe build

jaimehgb:
The build was broken for trader joe. It worked for me the last time I pushed it because I had some leftovers on my `generated` folder.


steegecs (approved):
LGTM

PR #1049, opened 2022-09-26
fix (#patch); dforce-arbitrum; fixed interest rates and rewardAPY

tnkrxyz:
This PR fixes #939 by changing two constants ARBITRUM_BLOCKS_PER_YEAR and ARBITRUM_BLOCKS_PER_DAY to be equal to ETHEREUM_BLOCKS_PER_YEAR and ETHEREUM_BLOCKS_PER_YEAR. More details in https://github.com/messari/subgraphs/issues/939#issuecomment-1257291939

Issue #868, opened 2022-08-24
dForce BSC QA


this-username-is-taken:
https://docs.google.com/spreadsheets/d/1iom5pm2nouStheavD9zgvbqK5JZt1RFb/edit#gid=507208116

tnkrxyz:
## Protocol

Section | Value | Status | Comments
-- | -- | -- | --
financialsDailySnapshots | totalValueLockedUSD | NOT OK | Different values in further dates as shown in DefiLlama and Footprint Network
financialDailySnapshots | mintedTokenSupplies | NOT OK | Cant be zero => seen on BSC addresses that mint occured, so as in dApp analytics | 

## Pool Overview 
| Pool | Field | Status | Comment |
| ---- | ----- | ------- | ---------- |
| All pool | Reward Token | not OK | Reward Token output error on subgraph (Code Error) | 
| BNB | inputToken | not OK | address 0x0000000000000000000000000000000000000000 conflicts with ETH |

tnkrxyz:
Fixed with merged #1062

PR #1062, opened 2022-09-27
fix (#minor); dforce; fixed QA issues reported in #868

tnkrxyz:
Fixed QA issues reported in #868
1. Fixed underlying token conflict: ZERO_ADDRESS is ETH on mainnet, BNB on BSC
2. Fixed protocol.mintedTokens and protocol.mintedTokenSupplies
3. Fixed financialDailySnapshot.mintedTokenSupplies (compound-forks/src/mapping.ts)
4. ARBITRUM_BLOCKS_PER_YEAR and ARBITRUM_BLOCKS_PER_DAY (also in #1049; will rebase this PR if #1049 is merged first)
5. bumped subgraph version to 1.2.0

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1255,7 +1255,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1277,7 +1277,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1299,7 +1299,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1321,7 +1321,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1343,7 +1343,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1365,7 +1365,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
           "methodology": "1.0.0"
         },
         "files": {
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2611,8 +2611,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.1",
-          "methodology": "1.1.0"
+          "subgraph": "1.2.0",
+          "methodology": "1.0.1"
         },
         "files": {
           "template": "makerdao.template.yaml"
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2611,8 +2611,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.0",
-          "methodology": "1.0.1"
+          "subgraph": "1.2.1",
+          "methodology": "1.1.0"
         },
         "files": {
           "template": "makerdao.template.yaml"
--- a/subgraphs/compound-forks/protocols/dforce/src/mapping.ts
+++ b/subgraphs/compound-forks/protocols/dforce/src/mapping.ts
@@ -157,9 +157,16 @@ export function handleMarketAdded(event: MarketAdded): void {
       underlyingSymbol = "DF";
       underlyingDecimals = 18;
     } else if (underlyingTokenAddr.toHexString() == ZERO_ADDRESS) {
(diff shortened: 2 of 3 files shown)


tnkrxyz:
See also #1106

tnkrxyz:
@dmelotik , @0xbe1 , or @this-username-is-taken , can one of you help review? thx!

deployment/deployment.json
```diff
@@ -1186,7 +1186,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.2.0",
```
dmelotik, line 1189: A patch should only update the last number: https://semver.org/#summary

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
```diff
@@ -603,33 +610,46 @@ export function handleStablecoinTransfer(event: StablecoinTransfer): void {
...
-    if (tokenIndex > 0) {
+    let tokenIndex = mintedTokens.indexOf(tokenId);
+    if (tokenIndex >= 0) {
       // token already in protocol.mintedTokens
-      protocol.mintedTokenSupplies![tokenIndex] = supply;
+      mintedTokenSupplies![tokenIndex] = supply;
+      protocol.mintedTokenSupplies = mintedTokenSupplies;
     } else {
-      if (tokenId < protocol.mintedTokens![0]) {
+      if (tokenId < mintedTokens[0]) {
```
dmelotik, line 637: Be careful with ordering here and checking that way (https://github.com/messari/subgraphs/blob/master/docs/Mapping.md#array-sorting-when-querying)

dmelotik (commented):
Nice work! A couple things, I now think we are enforcing rebasing onto the latest master, so you will have to do that.

I see dForce does have minted tokens, are we including this in the TVL on our subgraph?

And @ishraq8 for a second opinion. dForce minted tokens seem to be apart of their synthetics offerings, should this be included in this subgraph? To me it seems out of scope.

ishraq8:
@dmelotik The synthetics in dforce seems to work somewhat similarly to synthetix (CDP and then goes to trade synthetic assets). The dforce synthetics while it can use the similar depositm, the mechanics are separate from the borrowing and may probably need it's own subgraph in the future like a CDP. Their synthetics market is ~$400K

dmelotik:
@this-username-is-taken for visibility, do we remove this from the dforce subgraph then?

deployment/deployment.json
tnkrxyz: Maybe I should reword the title of the PR instead. I feel the changes are substantial enough to warrant a minor version increase.

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
tnkrxyz: The purpose of the whole chunk of code (L625-649) is to make sure the order of mintedToken and mintedTokenSupplies arrays are consistent.

tnkrxyz:
> Nice work! A couple things, I now think we are enforcing rebasing onto the latest master, so you will have to do that.
I believe I did that; I will rebase again before merging.

> 
> I see dForce does have minted tokens, are we including this in the TVL on our subgraph?
They are not included. Since (most) minted tokens are borrowed against collateral, including them would double count? 

> 
> And @ishraq8 for a second opinion. dForce minted tokens seem to be apart of their synthetics offerings, should this be included in this subgraph? To me it seems out of scope.

@this-username-is-taken and I talked a while back when I first started working on the dforce subgraph and he thought synthetics should not be included. Welcome second/third opinion here.

ishraq8:
Thanks @tnkrxyz! Looks like there is consensus that the synthetics component should be handled separately.

dmelotik:
> I believe I did that; I will rebase again before merging.

I was looking at this, but we will see if it is an issue when we are actually ready to merge: 
Screen Shot 2022-10-04 at 10 13 31 AM

I guess my question is should we have mintedTokens in the dForce subgraph? I am kinda confused there. I think we could also merge this PR then in a separate PR remove the mintedTokens mappings. Up to you really.

deployment/deployment.json
dmelotik: I misread the title, I thought it said #patch . I guess the way I think of it is that since nothing was added to the schema there were no functionality changes. But honestly I am not totally sure

tnkrxyz:
> I was looking at this, but we will see if it is an issue when we are actually ready to merge: Screen Shot 2022-10-04 at 10 13 31 AM

Rebased again.

> I guess my question is should we have mintedTokens in the dForce subgraph? I am kinda confused there. I think we could also merge this PR then in a separate PR remove the mintedTokens mappings. Up to you really.

My understanding is that mintedTokens and mintedTokenSupplies are specific for protocols like dforce, inverse-finance, and markedao, that mints stable coins, so they should be included here. Ping @this-username-is-taken for confirmation that is still the case.

this-username-is-taken:
That's right, we should include minted tokens (similar to makerdao, abra, etc) so there is a way to track them

deployment/deployment.json
```diff
@@ -2555,8 +2555,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.0",
-          "methodology": "1.0.1"
+          "subgraph": "1.2.1",
```
dmelotik, line 2558: remove this update

dmelotik:
Small update and I think we are ready to merge!

Note, same thing with the new linting checker. Ignore this for now and we will make a separate PR to tackle that since there is a lot in it.

tnkrxyz:
> Small update and I think we are ready to merge!
> 
> Note, same thing with the new linting checker. Ignore this for now and we will make a separate PR to tackle that since there is a lot in it.

Since #1157 is now merged, I'm rebasing to get the fix.

dmelotik (approved):

PR #1089, opened 2022-09-28
fix(#patch); truefi; decimals for input token balance

dmelotik:
- Fixed inputTokenBalance to just use the value from `value()` contract call
- used the new value and outputTokenSupply to get exchangeRate "properly"
- used tvl for totalDepositBalanceUSD
- updated to 1.0.2

testing: https://okgraph.xyz/?q=dmelotik%2Ftruefi-ethereum


this-username-is-taken (approved):
LGTM!

PR #1092, opened 2022-09-28
fix (#minor); Uniswap V3; Incorporate Collect events into indexing process to fix token balance and TVL issues.

steegecs:
**Context:** 
In response to much of the drama and attention to the TVL issues that have been happening with Uniswap V3, this PR should fix the issues that were occurring in the calculation of TVL and token balances within the pool, protocol, and financials entities. 

The issue was that we were not indexing data from the `Collect` event in the Pool contracts. Because of this, the fees were not being accounted for as removed from the liquidity pools, leading to an overestimation of the amount of tokens locked in all of the liquidity pools. The fix involves just tracking this event, removing the liquidity from pools at the time of the event, and updating TVLs in the relevant entities. 

**Changes:**
- [x] Added a handler for the `Collect` event
- [x] Added `Collect` entity to account for new usage type, and update usage metrics to account for new usage type
- [x] Update pool balances and TVL using the data from the collect event.

### DIFF
--- a/subgraphs/uniswap-v3/schema.graphql
+++ b/subgraphs/uniswap-v3/schema.graphql
@@ -286,6 +286,9 @@ type UsageMetricsDailySnapshot @entity {
   " Total number of trades (swaps) in a day "
   dailySwapCount: Int!
 
+  " Total number of collections (remove accrued fees) in a day "
+  dailyCollectCount: Int!
+
   " Total number of pools "
   totalPoolCount: Int!
 
@@ -320,6 +323,9 @@ type UsageMetricsHourlySnapshot @entity {
 
   " Total number of trades (swaps) in an hour "
   hourlySwapCount: Int!
+
+  " Total number of collections (remove accrued fees) in an hour "
+  hourlyCollectCount: Int!
 
   " Block number of this snapshot "
   blockNumber: BigInt!
@@ -786,6 +792,50 @@ type Swap implements Event @entity {
 
   " Amount of token withdrawn from pool in USD "
   amountOutUSD: BigDecimal!
+
+  " The pool involving this transaction "
+  pool: LiquidityPool!
+}
+
+type Collect implements Event @entity {
+  " collect-{ Transaction hash }-{ Log index }"
+  id: ID!
+
+  " Transaction hash of the transaction that emitted this event "
+  hash: String!
+
+  " Event log index. For transactions that don't emit event, create arbitrary index starting from 0 "
+  logIndex: Int!
+
+  " The protocol this transaction belongs to "
+  protocol: DexAmmProtocol!
+
+  " Address that received the tokens "
+  to: String!
+
+  " Address that sent the tokens "
+  from: String!
+
+  " Block number of this event "
+  blockNumber: BigInt!
+
+  " Timestamp of this event "
+  timestamp: BigInt!
+
+  " Input tokens of the pool (not input tokens of the event/transaction). E.g. WETH and USDC from a WETH-USDC pool "
+  inputTokens: [Token!]!
+
+  " Output token of the pool (not output token of the event/transaction). E.g. the UNI-LP token "
+  outputToken: Token
+
+  " Amount of input tokens in the token's native unit "
+  inputTokenAmounts: [BigInt!]!
+
+  " Amount of output tokens in the token's native unit "
+  outputTokenAmount: BigInt
+
+  " USD-normalized value of the transaction of the underlying (e.g. sum of tokens withdrawn from a pool) "
+  amountUSD: BigDecimal!
 
   " The pool involving this transaction "
   pool: LiquidityPool!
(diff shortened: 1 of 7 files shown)


subgraphs/uniswap-v3/schema.graphql
```diff
@@ -791,6 +797,50 @@ type Swap implements Event @entity {
   pool: LiquidityPool!
 }
 
+type Collect implements Event @entity {
```
jaimehgb, line 800: Should have we used `Withdrawal` for these events? Since they are like a withdrawal of accrued fees from the liquidity pool. If anything we can add an `isCollect` field on the withdrawal entity or something. The rest of fields seem to be pretty much the same in both entities.

If we count fee collections separately we'll have to handle them downstream everywhere. I think sticking to the standard dex schema, when it makes sense to, is valuable. What do you think?

subgraphs/uniswap-v3/src/common/creators.ts
```diff
@@ -441,3 +442,79 @@ export function createSwapHandleVolumeAndFees(
...
+  ];
+
+  let token0 = getOrCreateToken(pool.inputTokens[INT_ZERO]);
+  let token1 = getOrCreateToken(pool.inputTokens[INT_ONE]);
+
+  // Get formatted amounts collected.
+  let amount0Converted = convertTokenToDecimal(amount0, token0.decimals);
+  let amount1Converted = convertTokenToDecimal(amount1, token1.decimals);
+
+  // Get the value in USD of the deposit
```
jaimehgb, line 470:
```suggestion
  // Get the value in USD of the withdrawal (or collection)
```

subgraphs/uniswap-v3/src/common/creators.ts
```diff
@@ -441,3 +442,79 @@ export function createSwapHandleVolumeAndFees(
...
+
+  // Update the pool balances for collected fees in converted amount
+  poolAmounts.inputTokenBalances = [
+    poolAmounts.inputTokenBalances[INT_ZERO].minus(amount0Converted),
+    poolAmounts.inputTokenBalances[INT_ONE].minus(amount1Converted),
+  ];
+
+  // Subtract the TVL of the pool from the protocol before fee collection
+  protocol.totalValueLockedUSD = protocol.totalValueLockedUSD.minus(
+    pool.totalValueLockedUSD
```
jaimehgb, line 483: I freaked out here for a second 😅 

Can you maybe put the addition and substraction closer together? Something like:

```
let oldPoolTVL = pool.totalValueLockedUSD;

// update pool TVL
[...]

protocol.totalValueLockedUSD = protocol.totalValueLockedUSD
    .minus(oldPoolTVL)
    .plus(pool.totalValueLockedUSD)
```

Or alternatively:
```
let delta = pool.totalValueLockedUSD.minus(oldPoolTVL);
protocol.totalValueLockedUSD = protocol.totalValueLockedUSD.plus(delta);
```

subgraphs/uniswap-v3/src/common/creators.ts
```diff
@@ -441,3 +442,79 @@ export function createSwapHandleVolumeAndFees(
...
+    event.transaction.hash
+      .toHexString()
+      .concat("-")
+      .concat(event.logIndex.toString())
+  );
+
+  collect.hash = event.transaction.hash.toHexString();
+  collect.logIndex = event.logIndex.toI32();
+  collect.protocol = protocol.id;
+  collect.to = owner.toHexString();
```
jaimehgb, line 506: Should we be using the owner of the position here or the recipient of the fee collection? Since an owner can collect fees to a different address

subgraphs/uniswap-v3/src/common/updateMetrics.ts
```diff
@@ -88,6 +88,9 @@ export function updateUsageMetrics(
   } else if (usageType == UsageType.SWAP) {
     usageMetricsDaily.dailySwapCount += INT_ONE;
     usageMetricsHourly.hourlySwapCount += INT_ONE;
+  } else if (usageType == UsageType.COLLECT) {
+    usageMetricsDaily.dailyCollectCount += INT_ONE;
+    usageMetricsHourly.hourlyCollectCount += INT_ONE;
```
jaimehgb, line 93: Related to my first comment, should we increase the counter of withdrawals too? Essentially a collection is a withdrawal. If someone expects the default DEX schema or is not aware of collections they might expect money out to be counted as a withdrawal?

subgraphs/uniswap-v3/schema.graphql
steegecs: Yeah - I think this makes sense and is a good idea!

subgraphs/uniswap-v3/src/common/creators.ts
steegecs: Sure! I fixed this

subgraphs/uniswap-v3/src/common/updateMetrics.ts
steegecs: Yeah - makes sense. I just removed the collect from usage metrics.

subgraphs/uniswap-v3/src/common/creators.ts
steegecs: Yeah - That makes sense to me after looking though the documentation.

jaimehgb (approved):
LGTM

PR #1093, opened 2022-09-28
fix (#minor); deployment.json; Updated deploying.json conventions for services and updated scripts accordingly

steegecs:
**Context:** 
In response to our slack discussions, we decided to updated `deployment-ids` to `services`, and added `subgraphs-slug` and `query-id` as references for deployments to a particular service. Making this change entailed making a few changes to the deployment and json validation scripts.


deployment/transform.js
```diff
@@ -0,0 +1,29 @@
...
+
+// write JSON string to a file
+fs.writeFile("new_deployment.json", data, (err) => {
+  if (err) {
+    throw err;
+  }
+  console.log("JSON data is saved.");
+});
+
+console.log(deploymentJsonData);
```
jaimehgb, line 29: I think this file was not intended to be commited? 🙈

steegecs: Removed.

jaimehgb (approved):

rysuds (approved):
LGTM

PR #1097, opened 2022-09-28
fix(): updated maple-finance dependencies to fix build

jaimehgb:
Maple finance build seems to be broken because of some mismatch between the graph-ts versions during build and defined in the project itself. This should fix it for now


tnkrxyz (approved):
looks good to me. thx!

PR #1103, opened 2022-09-29
feat/fix (#minor); deployment; Added tests for valid fields

steegecs:
**Context:**
Added tests on the deployment.json so that arguments that are not standard will trigger an error. It requirs a certain number of fields for the protocol, deployment, and version level data.


mrbrianhobo (approved):
LGTM

PR #1113, opened 2022-09-29
fix(): temporarily whitelisted non-standard subgraph dir structures

jaimehgb:
A few subgraphs have been added since I opened #1035 which didn't follow the directory structure enforced by folderslint. I didn't rebase to last master before merge so I didn't catch it. This is currently making all CI to fail on all new branches.

To make the CI pipelines to pass while the issue at hand is resolved I've temporarily whitelisted these new dir structures. As soon as they are fixed I'll revert this change :)


steegecs (approved):
Sounds good!

PR #1114, opened 2022-09-29
fix(): fixed radiant-capital deployment.json since it had the old structure

jaimehgb:
The deployment.json structure was slightly changed during the time #976 was open, which broke the deployment scripts 😅 

Screenshot 2022-09-30 at 00 27 50


steegecs (approved):
Looks good!

PR #1118, opened 2022-09-30
fix(#minor); deployment.json; Fixed miscellaneous issues in the deployment.json file

steegecs:
**Context:**
Venus and Radiant had their slugs and query ids mismatch the location on the hosted service. This PR fixes those. Additionally, Tornado Cash was nested inside LooksRare, so I fixed that as well.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1528,7 +1528,7 @@
     "base": "compound-forks",
     "protocol": "venus",
     "deployments": {
-      "venus-bsc": {
+      "venus-protocol-bsc": {
         "network": "bsc",
         "status": "prod",
         "versions": {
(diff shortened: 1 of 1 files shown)


deployment/deployment.json
```diff
@@ -4617,49 +4617,49 @@
...
+      "tornado-cash-bsc": {
+        "network": "bsc",
+        "status": "prod",
+        "versions": {
+          "schema": "1.3.0",
+          "subgraph": "1.0.0",
+          "methodology": "1.0.0"
+        },
+        "deployment-ids": {
+          "hosted-service": "tornado-cash-bsc"
```
jaimehgb, line 4655: Since you are at it 🙈 
We also had the old `deployment-ids` instead of `services` for TC

steegecs: Good catch - updated!

jaimehgb (approved):

Issue #1111, opened 2022-09-29
Compound Forks Total Borrow Accuracy


dmelotik:
We can improve the accuracy and reliability of `market.totalBorrowBalanceUSD` through using other events to get this data.

In compound forks the `AccrueInterest` event is emitted and ingested before an event for `RepayBorrow` or `Borrow`. This doesn't accurately represent the actual borrow amount afterwards, since we only update borrows on accrue interest, but it can change after.

The transaction events also include the new `totalBorrows`, so we can utilize that to get more accurate data!

tnkrxyz:
closed with #1136 merged

Issue #253, opened 2022-05-31
Compound/comp-forks: negative TVL


dmelotik:
Negative `market.totalValueLocked` and `market.inputTokenBalance` in some cases.

See [gql query](https://api.thegraph.com/subgraphs/name/messari/compound-ethereum/graphql?query=query+MyQuery+%7B%0A++market%28id%3A+%220xf5dce57282a584d2746faf1593d3121fcac444dc%22%29+%7B%0A++++id%0A++++name%0A++++inputTokenBalance%0A++++totalValueLockedUSD%0A++%7D%0A%7D) and [subgraph visualizer](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/compound-ethereum&tab=pool&poolId=0xf5dce57282a584d2746faf1593d3121fcac444dc)

![image](https://user-images.githubusercontent.com/60583747/171300811-b38ccbe3-4dd4-475f-941d-63747df857a8.png)

Issue: `inputTokenBalance` is being calculated from Mints and Redeems. This would be okay, but there is interest accrued on borrows which is not accounted for. So when a market's TVL is close to 0 it is at risk of this bug. In addition, inputTokenBalance and TVL are not accurate with this mapping.

Fix: Use `market.outputTokenSupply`: `inputTokenSupply` = `outputTokenSupply` * `exchangeRate`. This will also create an accurate TVL because it is calculated from `inputTokenSupply`.

PR #1136, opened 2022-10-03
fix (#patch); compound-forks; improve accuracy of tvl/totalBorrow

tnkrxyz:
This PR adds updates for tvl/deposit for mint/redeem event, and totalBorrowBalance for borrow/repay event and addresses issue described in #1111

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2611,8 +2611,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.1",
-          "methodology": "1.1.0"
+          "subgraph": "1.2.0",
+          "methodology": "1.0.1"
         },
         "files": {
           "template": "makerdao.template.yaml"
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2611,8 +2611,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.0",
-          "methodology": "1.0.1"
+          "subgraph": "1.2.1",
+          "methodology": "1.1.0"
         },
         "files": {
           "template": "makerdao.template.yaml"
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -978,7 +978,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.4",
+          "subgraph": "1.1.5",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1007,7 +1007,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.5",
+          "subgraph": "1.1.6",
           "methodology": "1.0.0"
         },
         "files": {
@@ -1036,7 +1036,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.5",
+          "subgraph": "1.1.6",
           "methodology": "1.0.0"
         },
         "files": {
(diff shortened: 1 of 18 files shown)


tnkrxyz:
This fixes the gap between totalBorrowBalanceUSD in the subgraph and the on-chain data; verified to work on the dforce-ethereum subgraph & produce a better match of totalBorrowBalanceUSD with on-chain data (see https://docs.google.com/spreadsheets/d/1SDnhWSiRuv3_7ykxykjGLFflruRtlQ-3/edit#gid=535577190 , protocol and ETH market).

I have tested the dforce and aurigami subgraphs. @dmelotik , let me know if you feel it is necessary to test all compound-forks subgraphs.

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1009,6 +1032,9 @@ export function _handleAccrueInterest(
 
   // creates and initializes market snapshots
 
+  // those were called in updateMarket()
+  // TODO: remove before merge after review
```
dmelotik, line 1036: extra note to remove this 😉

dmelotik (commented):
Nice work / refactor!  🚀 

Could you also test on Compound just to ensure the inputTokenBalance logic is sound, that is where it was prompted from: https://github.com/messari/subgraphs/issues/253

And should `updateProtocol()` and `snapshotFinancials()` also be moved in order to reflect the new updates on the transactions instead of `AccrueInterest()`?

deployment/deployment.json
```diff
@@ -1545,8 +1545,8 @@
         },
         "services": {
           "hosted-service": {
-            "slug": "venus-protocol-bsc",
-            "query-id": "venus-protocol-bsc"
+            "slug": "venus-bsc",
```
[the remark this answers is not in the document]
tnkrxyz: I had thought this is causing the build error for venus-bsc: https://github.com/messari/subgraphs/actions/runs/3186399411/jobs/5196948176 .

Looks like it may not be, as the error remains with the renaming. Still looking...

subgraphs/compound-forks/src/mapping.ts
tnkrxyz: Will do.

tnkrxyz:
> Could you also test on Compound just to ensure the inputTokenBalance logic is sound, that is where it was prompted from: #253

Started a test deployment of compound-v2: https://thegraph.com/hosted-service/subgraph/tnkrxyz/test-subgraph
 
> And should `updateProtocol()` and `snapshotFinancials()` also be moved in order to reflect the new updates on the transactions instead of `AccrueInterest()`?

Good point! Fixed.

subgraphs/compound-forks/src/mapping.ts
tnkrxyz: done.

deployment/deployment.json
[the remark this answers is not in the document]
tnkrxyz: I believe it is indeed the cause (there needs one more renaming). 

Not sure what the convention is. Since the directory for the venus protocol inside compound-forks/protocols is called "venus", the slug and query id should be "venus-bsc". It seems inconsistent to name them "venus-protocol-bsc"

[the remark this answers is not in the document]
dmelotik: Since we originally named the hosted service subgraph "messari/venus-protocol-bsc" we just left it.

I don't believe it is what is failing the build.

deployment/deployment.json
```diff
@@ -2555,8 +2555,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.0",
-          "methodology": "1.0.1"
+          "subgraph": "1.2.1",
```
dmelotik, line 2558: looks like this was accidentally changed in a rebase

dmelotik:
2 things, the versions should also be updated in the deployment.json, and now that we added the lint checker in the ci/cd checks that is what is failing now.

I think we should merge this, then create a new PR to tackle all of the linting issues. So don't worry about the linting check, we can merge regardless

dmelotik:
LGTM

dmelotik (approved):

PR #1140, opened 2022-10-04
feat (#minor); Eslint action; Add linting action when pushing to the repo.

steegecs:
**Context:**
We want to make sure that the code that is getting merged into the repository is linted. This PR adds an action to the workflow that runs our linter on changed files within the subgraph folder. When a file needs to be linted, this portion of the action will fail.

### DIFF
--- a/.github/workflows/build-checker.yaml
+++ b/.github/workflows/build-checker.yaml
@@ -23,9 +23,37 @@ jobs:
       
       - name: Install dependencies
         run: npm ci --ignore-scripts
-
+      
       - name: Folderslint
         run: npm run folderslint
+  
+  LintStagedFiles:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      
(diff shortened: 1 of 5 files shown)


.github/workflows/build-checker.yaml
```diff
@@ -23,9 +23,37 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      
+      - name: Get changed files in the docs folder
```
jaimehgb, line 40: In the docs folder?

jaimehgb (approved):

.github/workflows/build-checker.yaml
steegecs: Yeah - that should be subgraphs. Also, made another update to only grab changed files in the subgraphs folder.

Issue #1157, opened 2022-10-07
Fix compound fork linting


dmelotik:
Create a PR to fix the linting issues in compound-forks. This should be the only change made in that PR, but since it is a big directory we want to make a separate PR to do so.

dmelotik:
Looks to be fixed! Nice job @tnkrxyz. Going to close this out

PR #1159, opened 2022-10-07
fix(#patch); bastion; reward token reverting

dmelotik:
Reward token reverting and causing incorrect output.

To override since we know the name and decimals of the token we hardcode it.

Testing: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmXoLTGo3w6f1EWShayjCka7aP8n6jV11jpBw2JHTrXJcA&tab=protocol

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1021,7 +1021,7 @@
         "status": "dev",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.5",
+          "subgraph": "1.1.6",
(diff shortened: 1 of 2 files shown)


dmelotik:
Linting errors will be fixed with #1157

jaimehgb (approved):
🚀 
Their contract is not even verified? Was curious to know why the call was reverting :(

dmelotik:
Hard to say... A little weird why it is not verified. I am gonna throw a message in their discord

PR #1167, opened 2022-10-08
fix(#patch); euler; oneshot cancelled fix

dmelotik:
Fixing oneshot cancelled error. 

Testing graft here: https://okgraph.xyz/?q=dmelotik%2Feuler-finance-ethereum


dmelotik:
fixed in #1213

PR #1168, opened 2022-10-08
feat (#major) CLI; Command is now agnostic to call location, and scope no longer necessary

steegecs:
**2 New Features:**
- [x] The `messari build` command is now agnostic to the location that you call it from. You can `messari build` from any location in your terminal now :D. Feel free to build or deploy `apeswap` from the  `compound-forks` folder or even from any location outside the repo!
- [x] Because we enforce that singular deployments and deploying a whole protocol, we are able to omit the `scope` option and prompt. This PR updates the CLI so that when you pass either a `deployment-id` or `protocol` name, it can determine the scope automatically - and will know whether you are trying to deploy one or multiple subgraphs. If you would still like to use the functionality for deploying all subgraphs in a directory (`base`), just add the option `--base` to your command.


messari-cli/bin/env.ts
```diff
@@ -0,0 +1,9 @@
+const execSync = require('child_process').execSync
+
+// Change to the directory of this file
+process.chdir(__dirname)
+
+// Get the path to the head of the `Subgraphs` repo
+export const MESSARI_REPO_PATH = execSync('git rev-parse --show-toplevel')
```
jaimehgb, line 7: This assumes there will only be one local git clone of the repo on each system. It might be true 99% of the time, so it is probably fine. But if we ever make it an npm package we migth want to be able to set the `MESSARI_REPO_PATH` env var dinamically or something.

Don't need to change it now, but let's keep it in mind for potential updates. It might be worth documenting it too in the cli README so people are aware though.

jaimehgb (approved):
Looks correct to me 💪

messari-cli/bin/env.ts
steegecs: Yeah - I had considered this. I chose to do it this way, for now because it was easier. But on second thought, I might have to make this change soon, since I am not sure if the path will be set correctly in the build checker CI action.

PR #1170, opened 2022-10-09
(#patch); CLI; Move the creation of the deployment.json context class to inside src.

steegecs:
**Context:**
I think it fits better to put in the src folder, also, having it out causes linting errors, because I am putting a typescript file in the main body of the project.


harsh9200 (approved):
LGTM !!

PR #1173, opened 2022-10-09
(#patch); Eslint; Added eslint ignore and updated eslintrc to apply tsconfig same directly

steegecs:
**Context:**
Added a .eslintignore to the subgraphs linter so that the build and generated files do not get linted. Because they are types generated by the codegen and build commands, we do not want to lint them according to our standards. 

Additionally, eslint was not using the tsconfig file from the same location as the eslintrc. This does not make linting with our standards as flexible as we want it to be. Updated the `eslintrc.json` to `eslintrs.js` so that we could use `__dirname` within the configurations to always use the tsconfig withing the subgraphs/ folder.


harsh9200 (approved):
LGTM! 👍

PR #1177, opened 2022-10-10
feat(#major); alpaca-finance-lending-protocol; Add Alpaca Finance Lending Protocol Subgraph

corerouter:
The subgraph link is https://thegraph.com/hosted-service/subgraph/corerouter/alpaca


subgraphs/alpaca-finance-lending/README.md
```diff
@@ -0,0 +1,66 @@
+# Alpaca Finance Lending Protocol Subgraph
+
+## Calculation Methodology v1.0.0
+
+Alpaca Finance is a leading lending protocol allowing leveraged yield farming on BNB Chain. It helps lenders earn safe and stable yields, and offers borrowers undercollateralized loans for leveraged yield farming positions, vastly multiplying their farming principals and resulting profits.‌
+
+### Total Value Locked (TVL) USD
+
+Sum across all Reserves:
```
jaimehgb, line 9: Should we also include the deposits in active positions and deposits in their automated vaults.

When borrowing from alpaca finance you have to put down a certain amount which is not deposited in the lending pools. Should that add to the TVL? Or not because it is just passed through to Pancake/Biswap/etc?

Also, alpaca tokens locked in their governance contract?

this-username-is-taken: @ishraq8 thoughts on TVL here?

ishraq8: @jaimehgb 
1) Governance tokens that are locked or staked we do not include in TVL (similar to other tokens staked in their protocol). 
2) The token/stablecoin amount used for yield farming should be included in tvl, it works just like a yield aggregator added to their lending component based on their docs. Note: this should not be the interest bearing token that is produced from lending (receipt token) that is used in farming. 

Does that make sense? Simple answer: Include all non Alpaca tokens into TVL and exclude tokens produced by the protocol itself.

corerouter: @ishraq8 @jaimehgb @this-username-is-taken 

Totally agree with the governance token is not included in TVL. 

For the second one, as this subgraph focus on the alpaca finance lending protocol and doesn't include the yield farming protocol, I think the token used for the yield farming should not be included in the TVL for the lending protocol, as the token is not deposited in the lending protocol. As you can see from https://app.alpacafinance.org/lend and https://app.alpacafinance.org/farm, alpaca have different pools for lending and yield farming and this subgraph only take the lending pool into consideration. The total supply computed with this method from the subgraph also are in line with the data shown from their website at https://app.alpacafinance.org/lend

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
```diff
@@ -0,0 +1,405 @@
...
+    event,
+    market,
+    Address.fromString(market.inputToken),
+    event.params.posVal,
+    Address.fromString(market.inputToken),
+    event.params.prize,
+    bigDecimalToBigInt(
+      event.params.prize.toBigDecimal().times(protocolSideProfitRatio)
+    ),
+    event.params.killer,
```
jaimehgb, line 260: Also, alpaca finance runs its own liquidation bots. What they make from them is used to buyback and burn alpaca. Should we count liquidations from those addresses entirely as protocol revenue?

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
```diff
@@ -0,0 +1,405 @@
...
+    PositionSide.LENDER,
+    true
+  );
+}
+
+function handleMint(event: Transfer): void {
+  const contract = Vault.bind(event.address);
+  const tryTokenResult = contract.try_token();
+  if (tryTokenResult.reverted) {
+    return;
```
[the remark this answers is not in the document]
corerouter: Most of cases it should not revert. Just follow other messari subgraph best practice for defense coding. Added log in 71cba71400fa59bd478faf78c49fc2e8fd8708a6 Thanks!

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
```diff
@@ -0,0 +1,405 @@
...
+  }
+  createLiquidate(
+    event,
+    market,
+    Address.fromString(market.inputToken),
+    event.params.posVal,
+    Address.fromString(market.inputToken),
+    event.params.prize,
+    bigDecimalToBigInt(
+      event.params.prize.toBigDecimal().times(protocolSideProfitRatio)
```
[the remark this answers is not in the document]
corerouter: changed in 71cba71400fa59bd478faf78c49fc2e8fd8708a6

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
```diff
@@ -0,0 +1,405 @@
...
+    const tryGetInterestRate = configContract.try_getInterestRate(
+      tryVaultDebtVal.value,
+      poolTokenAmount
+    );
+
+    if (
+      !tryGetInterestRate.reverted &&
+      !tryVaultDebtVal.value.equals(BIGINT_ZERO) &&
+      !market.inputTokenBalance.equals(BIGINT_ZERO)
+    ) {
```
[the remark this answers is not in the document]
corerouter: changed in 71cba71400fa59bd478faf78c49fc2e8fd8708a6

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
corerouter: My original idea is to follow other subgraph's practice to just add the profits explicitly allocated to treasury as protocol side revenue. As said here https://docs.alpacafinance.org/tokenomics/proof-of-burn, there may be cases 3rd parties liquidation bots. Nevertheless, as right now the team are running the bots, we can count all profit as protocol side profit. Changed in 3https://github.com/messari/subgraphs/commit/71cba71400fa59bd478faf78c49fc2e8fd8708a6 Thanks!

subgraphs/alpaca-finance-lending/README.md
ishraq8: @corerouter Agreed with second point, their farms are auto compounders/yield aggregator and would require a second subgraph for the farms using yield aggregator. Thanks for the clarification

@jaimehgb In that case the TVL is fine as is.

subgraphs/alpaca-finance-lending/src/mappings/vault.ts
jaimehgb: Oh no, I meant the opposite 😅 Sorry if I wasn't clear.

What I meant was that **some** of the liquidators are owned by alpaca, not all of them. So for those liquidations that come from those bots, 100% of the prize would be protocolSideRevenue. Whereas for the rest of the bots, only the protocol liquidation fee.

corerouter: Sorry. Revert the commit and then set default protocolSideProfitRatio in liquidation to 80% which is in line with protocol config in 7a9b54c702bb4c6b5184aed9a455ec312a24a6c9 Thanks!

subgraphs/alpaca-finance-lending/src/entities/market.ts
```diff
@@ -0,0 +1,599 @@
...
+  market: Market,
+  balanceChange: BigInt
+): void {
+  const changeUSD = amountInUSD(
+    balanceChange,
+    getTokenById(market.inputToken),
+    event.block.number
+  );
+  market.totalBorrowBalanceUSD = market.totalBorrowBalanceUSD.plus(changeUSD);
+  if (market.totalBorrowBalanceUSD.lt(BIGDECIMAL_ZERO)) {
```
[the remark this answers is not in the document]
corerouter: The totalBorrowBalance would be negative if there is a problem in the codes logic. Add logging in that case at cf46ce06899cb5373e0d210bce9820ff79555937 Thanks!

subgraphs/alpaca-finance-lending/protocols/alpaca-finance-lending/config/templates/alpaca.finance.lending.template.yaml
```diff
@@ -0,0 +1,683 @@
+# prettier-ignore
```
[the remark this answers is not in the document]
corerouter: Change it at cf46ce06899cb5373e0d210bce9820ff79555937

jaimehgb (approved):

jaimehgb (approved):
Looks good to me! Very clean and well organised

corerouter:
> Looks good to me! Very clean and well organised

Thank you very much for your review!

PR #1186, opened 2022-10-11
fix(#patch): Saddle Finance missing pools and balance fixes

jaimehgb:
This PR aims to fix 2 issues:
1) In saddle finance, each the inputTokens in each pool have a specific order which is the one that needs to be used when listening to events. All the information related to any input token that comes from contract calls or events will follow that given order of inputTokens. But on our side, we sort input tokens alphabetically because of how the graph sorts arrays of entities. This caused a few issues, like **swaps where input and output tokens didn't match the actual swap tokens, or amounts that were computed using the wrong token decimals**, etc.
2) Some pools are deployed via a deployer contract, and others are deployed manually. Some of the manually deployed ones were missing from our subgraph. So this PR starts tracking a registry contract where all pools should be added eventually.


subgraphs/saddle-finance/src/entities/event.ts
```diff
@@ -15,10 +15,11 @@ import {
   handlePoolWithdraw,
 } from "./pool";
 import { getOrCreateProtocol, updateUsageMetrics } from "./protocol";
+import { sortValuesByTokenOrder } from "./pool";
 
 export function createDeposit(
   event: ethereum.Event,
-  inputTokenAmounts: BigInt[],
+  inputTokenAmounts: BigInt[], // these come sorted as in the contract, with might not match pool.inputTokens positions.
```
dmelotik, line 22: typo "which" might not match

subgraphs/saddle-finance/src/entities/event.ts
```diff
@@ -52,7 +53,7 @@ export function createDeposit(
 
 export function createWithdraw(
   event: ethereum.Event,
-  inputTokenAmounts: BigInt[],
+  inputTokenAmounts: BigInt[], // these come sorted as in the contract, with might not match pool.inputTokens positions.
```
dmelotik, line 56: same

subgraphs/saddle-finance/src/entities/pool.ts
```diff
@@ -56,42 +57,91 @@ import { SimpleRewarder } from "../../generated/templates/Swap/SimpleRewarder";
...
+
   const contract = Swap.bind(address);
   let lpTokenAddress = contract.swapStorage().value6;
   // Check if LP token exists
   if (!checkValidToken(lpTokenAddress)) {
     const v1contract = SwapV1.bind(address);
     lpTokenAddress = v1contract.swapStorage().value7;
     if (!checkValidToken(lpTokenAddress)) {
-      log.error("Invalid LP token address {} in pool {}", [
+      log.critical("Invalid LP token address {} in pool {}", [
```
dmelotik, line 125: we want to terminate the syncing if this happens?

dmelotik (commented):
the deployment.json still needs to be updated with the version

Also question: why remove the yarn.lock file?

LGTM once those small things are fixed. Not quite sure if @steegecs should do a review as well since i don't know the dexes as well..

steegecs (commented):
Update versions in the deployment.json. Otherwise, it looks good to me!

subgraphs/saddle-finance/src/entities/pool.ts
jaimehgb: Hmm, maybe not. I actually copied this from here: https://github.com/messari/subgraphs/pull/1186/files/53cb5cfddd4ef01fe3e589aa050a739e5c7b7a6c#diff-32d490cbe73170a2dc29bdba3bda4e19b4ca56bfc23bef0d8793b80b9a3ddd2cL500

The functionality was duplicated so I consolidated it into one function. In one place it logged an error and in the other it was a critical 😅 

But yes, might be safer in this case to keep it as an error. Changed it back! (it will now only panic when it fails if creating it from config, as it was initially)

jaimehgb:
Ready for review again. Added the last 4 commits, with the third one being the most relevant functionality-wise. It adds a fallback to calculate token prices from within saddle when the price lib fails to give a price. It does so by looking for pools where the token is traded against any whitelisted stable.

deployment/deployment.json
```diff
@@ -2757,7 +2757,7 @@
         "status": "dev",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.1.2",
+          "subgraph": "1.1.5",
```
steegecs, line 2760: I only see version updates for Arbitrum and Ethereum. Is that all that is impacted by these changes?

steegecs (commented):
There are a couple linting errors to fix. Other than the one other comment, everything looks good!

deployment/deployment.json
jaimehgb: Whoops, arbitrum wasn't intended to be changed. 

I still have to deploy on all the other networks, so far only ethereum has been deployed and QAd

jaimehgb:
> There are a couple linting errors to fix. Other than the one other comment, everything looks good!

I'm linting on a separate PR: https://github.com/messari/subgraphs/pull/1194

dmelotik (approved):
approved, just need to fix a couple linting tings

deployment/deployment.json
steegecs: SG

steegecs (approved):

PR #1195, opened 2022-10-12
feat(#major); Price Lib; Add support to five more EVM chains + Refactor

harsh9200:
**New Features**: 
- Added support to five more EVM chains (`aurora`, `gnosis`, `harmony`, `moonbeam`, and `optimism`).
- New `uniswapForksMethod` (`sushsiwapRouter` and `uniswapRouter`methods merged into one).
- New Config parameters for blacklisting price lib for specific tokens and particular methods. 
- New Config parameters for hardcoding missing/wrong-priced pegged coins.


steegecs (commented):
I have been testing this out, and it seems that the prices are pretty far off for polygon. They seem to often be off by 3 decimals and then sometimes by another 50% or so. Any idea what could be causing this? The prices on Ethereum were showing up just fine, though.

subgraphs/_reference_/src/prices/common/utils.ts
```diff
@@ -1,15 +1,68 @@
...
+    return new MOONBEAM.config();
+  } else if (network == OPTIMISM.NETWORK_STRING) {
+    return new OPTIMISM.config();
+  } else if (network == AVALANCHE.NETWORK_STRING) {
+    return new AVALANCHE.config();
+  } else if (network == ARBITRUM_ONE.NETWORK_STRING) {
+    return new ARBITRUM_ONE.config();
+  }
+
+  return new MAINNET.config();
```
dmelotik, line 67: It might make more sense for the mainnet config to sit in the if...else tree, then there is a proper warning message saying that the chain isn't supported if none are selected.

subgraphs/_reference_/src/prices/config/arbitrum.ts
```diff
@@ -0,0 +1,131 @@
...
+  "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"
+);
+export const WETH_ADDRESS = Address.fromString(
+  "0x82af49447d8a07e3bd95bd0d56f35241523fbab1"
+);
+export const USDC_ADDRESS = Address.fromString(
+  "0xff970a61a04b1ca14834a43f5de4533ebddb5cc8"
+);
+
+export class config implements Configurations {
```
dmelotik, line 75: not necessary to change, but one way to reduce the code and get rid of this class in every network config:
- you setup the network config as a data structure, and import that into utils.ts
- Then when the switch finds what network is being used it passes that as the constructor arg then all of the constants get set

subgraphs/_reference_/src/prices/config/gnosis.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+  curveRegistry(): Address[] {
+    return CURVE_REGISTRY_ADDRESSES;
+  }
+
+  hardcodedStables(): Address[] {
+    return HARDCODED_STABLES;
+  }
+
+  ethAddress(): Address {
```
dmelotik, line 109: you could consider naming this networkTokenAddress or nativeAddress

subgraphs/_reference_/src/prices/index.ts
```diff
@@ -69,58 +69,36 @@ export function getUsdPricePerToken(tokenAddr: Address): CustomPriceType {
...
-  let sushiswapPrice = getPriceUsdcSushiswap(tokenAddr, network);
-  if (!sushiswapPrice.reverted) {
-    log.warning("[SushiSwapRouter] tokenAddress: {}, Price: {}", [
-      tokenAddr.toHexString(),
-      sushiswapPrice.usdPrice.div(sushiswapPrice.decimalsBaseTen).toString(),
-    ]);
-    return sushiswapPrice;
-  }
-
   log.warning("[Oracle] Failed to Fetch Price, tokenAddr: {}", [
```
dmelotik, line 102: you should add the token name in there too, it makes it easier to quickly see what token failed

subgraphs/_reference_/src/prices/index.ts
```diff
@@ -132,7 +110,7 @@ export function getUsdPrice(
   tokenAddr: Address,
   amount: BigDecimal
 ): BigDecimal {
-  let tokenPrice = getUsdPricePerToken(tokenAddr);
+  const tokenPrice = getUsdPricePerToken(tokenAddr);
 
   if (!tokenPrice.reverted) {
     return tokenPrice.usdPrice.times(amount).div(tokenPrice.decimalsBaseTen);
```
dmelotik, line 116: CustomPrice.usdPrice should really be normalized to USD price with proper decimals. There is no reason it needs to be done after the fact.

After that you can also remove decimalsBaseTen

subgraphs/_reference_/src/prices/oracles/AaveOracle.ts
```diff
@@ -4,31 +4,24 @@ import { CustomPriceType } from "../common/types";
...
-    return new CustomPriceType();
-  }
 
-  let tokenPrice: BigDecimal = utils
+  const aaveOracleContract = getAaveOracleContract(config.aaveOracle());
+  if (!aaveOracleContract) return new CustomPriceType();
+
+  const tokenPrice: BigDecimal = utils
     .readValue<BigInt>(
       aaveOracleContract.try_getAssetPrice(tokenAddr),
```
dmelotik, line 26: Have you checked this? On mainnet aave v2 `getAssetPrice()` returns a value in ETH

dmelotik (changes requested):
Nice refactor and add! I have some requested changes, and I plan on trying this out tomo on aurora.

subgraphs/_reference_/src/prices/common/constants.ts
```diff
@@ -1,426 +1,24 @@
...
-import * as FANTOM from "../config/fantom";
-import * as POLYGON from "../config/polygon";
-import * as MAINNET from "../config/mainnet";
-import * as OPTIMISM from "../config/optimism";
-import * as ARBITRUM_ONE from "../config/arbitrumOne";
-
-import { Address, BigDecimal, BigInt, TypedMap } from "@graphprotocol/graph-ts";
-import { UniswapPair__getReservesResult } from "../../../generated/UniswapV2Factory/UniswapPair";
-import { SushiSwapPair__getReservesResult } from "../../../generated/UniswapV2Factory/SushiSwapPair";
+import { Address, BigDecimal, BigInt } from "@graphprotocol/graph-ts";
```
dmelotik, line 1: you should version the price lib in here somewhere

subgraphs/_reference_/src/prices/common/utils.ts
```diff
@@ -1,15 +1,68 @@
...
+import * as ARBITRUM_ONE from "../config/arbitrum";
+
 import * as constants from "./constants";
+import { Configurations } from "./types";
 import { _ERC20 } from "../../../generated/UniswapV2Factory/_ERC20";
-import { Address, BigInt, ethereum } from "@graphprotocol/graph-ts";
+import { Address, BigInt, dataSource, ethereum } from "@graphprotocol/graph-ts";
+
+export function isNullAddress(tokenAddr: Address): boolean {
+  return tokenAddr.equals(constants.NULL.TYPE_ADDRESS) ? true : false;
```
jaimehgb, line 19: `Address.equals()` already returns a boolean. You can skip the ternary operator

subgraphs/_reference_/src/prices/routers/UniswapForksRouter.ts
```diff
@@ -0,0 +1,200 @@
...
+  );
+  if (totalSupply == constants.BIGINT_ZERO || totalLiquidity.reverted) {
+    return new CustomPriceType();
+  }
+
+  let pairDecimals: number;
+  const pairDecimalsCall = uniSwapPair.try_decimals();
+
+  if (pairDecimalsCall.reverted) {
+    pairDecimals = constants.DEFAULT_DECIMALS.toI32() as u8;
```
jaimehgb, line 124: Is it safe to default to some decimals if the call reverted? Would it be preferable to return 0 price instead of potentially returning wrong values silently?

jaimehgb (commented):
Looking good! Just 2 minor things

subgraphs/_reference_/src/prices/routers/UniswapForksRouter.ts
dmelotik: Seconding this, I have ran into issues with this before. If it is 0 thought it will also be wrong, maybe throwing an error and returning the empty `CustomPrice` type is best here, then when we run into situations like this in the future we can have a whitelist?

subgraphs/_reference_/src/prices/README.md
```diff
@@ -171,10 +153,6 @@ Prices
 
 ### Polygon
 
-🔨 = In progress.  
-🛠 = Feature complete. Additional testing required.
-`MultiCall` = If the method uses more than two `JSON RPC Call`.
-
 | Method                                                                               |      Type      | StartBlock | MultiCall | Status |
 | ------------------------------------------------------------------------------------ | :------------: | :--------: | :-------: | :----: |
 | [QuickSwap](https://etherscan.io/address/0xa5E0829CaCEd8fFDD4De3c43696c57F7D7A678ff) |    `Router`    | `4931900`  |    ✅     |   🛠   |
```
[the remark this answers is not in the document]
harsh9200: yeah working on it

subgraphs/_reference_/src/prices/calculations/CalculationsSushiswap.ts
```diff
@@ -4,19 +4,34 @@ import { CustomPriceType } from "../common/types";
...
 
-export function getTokenPriceFromSushiSwap(tokenAddr: Address, network: string): CustomPriceType {
-  const curveContract = getSushiSwapContract(network);
+export function getTokenPriceUSDC(tokenAddr: Address): CustomPriceType {
+  const config = utils.getConfig();
+
+  if (!config || config.sushiCalculationsBlacklist().includes(tokenAddr))
+    return new CustomPriceType();
+
+  const curveContract = getSushiSwapContract(config.sushiCalculations());
```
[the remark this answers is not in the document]
harsh9200: nice catch

dmelotik:
Testing here: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmR9NRofS4PBy5GU4HywduByPJKBYGistTqwhVU2zoukAy&tab=protocol

The markets that use this are PLY, USN, TRI, and AURORA

subgraphs/_reference_/src/prices/oracles/ChainLinkFeed.ts
```diff
@@ -1,29 +1,42 @@
...
 
   if (!result.reverted) {
-    let decimals = chainLinkContract.try_decimals(tokenAddr, constants.CHAIN_LINK_USD_ADDRESS);
+    const decimals = chainLinkContract.try_decimals(
+      tokenAddr,
+      constants.CHAIN_LINK_USD_ADDRESS
+    );
 
     if (decimals.reverted) {
       new CustomPriceType();
```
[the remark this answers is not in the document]
harsh9200: yup, I will fix this

subgraphs/_reference_/src/prices/oracles/AaveOracle.ts
harsh9200: can you explain more? Mainnet does not support aave oracle method.

subgraphs/_reference_/src/prices/index.ts
harsh9200: Yeah, that makes more sense.

subgraphs/_reference_/src/prices/routers/UniswapForksRouter.ts
harsh9200: Sounds Good. I am now returning empty `CustomPrice` if the pair decimals call is reverted.

subgraphs/_reference_/src/prices/config/gnosis.ts
harsh9200: `ethAddress` and `wethAddress` parameters are named according to uniswap contract parameters. As all the uniswap forks contracts follow the same nomenclature, we should do the same.

subgraphs/_reference_/src/prices/index.ts
harsh9200: Done

subgraphs/_reference_/src/prices/config/arbitrum.ts
steegecs: I have not found a way to create the data structure using a type or interface to be passed along to a constructor.

subgraphs/_reference_/src/prices/routers/CurveRouter.ts
```diff
@@ -1,120 +1,280 @@
...
+
+export function isCurveLpToken(lpAddress: Address): bool {
+  const poolAddress = getPoolFromLpToken(lpAddress);
+  if (poolAddress.notEqual(constants.NULL.TYPE_ADDRESS)) return true;
+
+  return false;
+}
+
+export function getPoolFromLpToken(lpAddress: Address): Address {
+  let config = utils.getConfig();
```
dmelotik, line 17: I'd say to correct the lint issues in this PR, seems like a lot of them are from added code not existing. So it would not make the PR messier

subgraphs/_reference_/src/prices/common/constants.ts
dmelotik: Where does this live @harsh9200 ?

dmelotik (approved):

PR #1199, opened 2022-10-13
[GOV] Add delegate and voting power change entities to Aave / dYdX subgraphs

danielkhoo:
## Description

Propagate new entities from OpenZeppelin to Aave subgraphs

### DIFF
--- a/subgraphs/aave-governance/src/handlers.ts
+++ b/subgraphs/aave-governance/src/handlers.ts
@@ -158,7 +158,7 @@ export function getOrCreateDelegate(address: string): Delegate {
     delegate.save();
 
     if (address != ZERO_ADDRESS) {
-      let governance = getGovernance();
+      const governance = getGovernance();
       governance.totalDelegates = governance.totalDelegates.plus(BIGINT_ONE);
       governance.save();
     }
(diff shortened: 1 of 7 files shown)


subgraphs/aave-governance/src/handlers.ts
```diff
@@ -84,6 +86,54 @@ export function getGovernance(): Governance {
...
+  delegateChange.txnHash = event.transaction.hash.toHexString();
+  delegateChange.blockNumber = event.block.number;
+  delegateChange.blockTimestamp = event.block.timestamp;
+  delegateChange.logIndex = event.logIndex;
+
+  return delegateChange;
+}
+
+export function createDelegateVotingPowerChange(
+  event: ethereum.Event,
```
Bijan-Massoumi, line 112: Is it possible to put these handlers in a shared directory for both OZ and aave?

Bijan-Massoumi (approved):

subgraphs/aave-governance/src/handlers.ts
danielkhoo: I'm not sure tbh. But there are some minor differences in the handler functions / schema that would make sharing a bit awkward.

PR #1207, opened 2022-10-15
(#chore); Deployment Scripts; Remove old deployment scripts from deployments and package.json

steegecs:
**Context:**
The CLI is now polished and being used by a handful of our developers. It appears to be working very smoothly, and it is now being used in the build checker CI. For this reason, I am proposing this PR to remove the old scripts from the `deployments` folder as they are unnecessary unless using the old build/deployment paradigm. Additionally, this PR removes the old scripts from the package.json for all subgraphs, except for the `prepare:constants` as this is still required by the CLI when used. 

Since the CLI has been running smoothly, I think it is time to clean out old code and encourage people to use the new and better build/deployment method.


jaimehgb (approved):
I agree 🚀

PR #1218, opened 2022-10-17
fix(); aave forks; fix linting issues

dmelotik:


dmelotik (commented):
@steegecs Interesting thing to note here: In a handler you NEED to include the event and proper type even if you do not use the event. 

Is there any way to skip this in the linter?

jaimehgb:
> @steegecs Interesting thing to note here: In a handler you NEED to include the event and proper type even if you do not use the event.
> 
> Is there any way to skip this in the linter?

@dmelotik You can explicitly tell the linter to ignore a line (or a file). Add this comment on the line right above:
```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
(or just `eslint-disable-next-line`). This is one scenario when it can be justified to ignore the linter :D

dmelotik:
yay breaking shit, thanks @jaimehgb

subgraphs/aave-v2-forks/protocols/aave-v2/src/mapping.ts
```diff
@@ -79,15 +79,15 @@ import { IPriceOracleGetter } from "../../../generated/LendingPool/IPriceOracleG
 import { Transfer } from "../../../generated/templates/AToken/AToken";
 
 function getProtocolData(): ProtocolData {
-  let letants = getNetworkSpecificConstant();
```
jaimehgb, line 82: lol

jaimehgb (approved):

subgraphs/aave-v2-forks/protocols/aave-v2/src/mapping.ts
dmelotik: I think that happened when i was doing a ctrl+f change lmao

PR #1220, opened 2022-10-17
fix(); aave forks; add _ to helper fields

dmelotik:


jaimehgb (approved):
👍

jaimehgb:
Actually, thinking about this now. This will prevent grafting from subgraph versions earlier to this change, right? Since the schema won't be compatible.

Does aave sync fast? If so then I guess it's all good.

dmelotik:
No @jaimehgb aave doesn't sync fast. Maybe best to leave this out for now. I will make it a draft

dmelotik:
Just going to close this. If we make this change now it will prevent any grafting opportunity. Eventually we would move aave to the 3.0.0 schema and we can make those changes then.

Issue #1133, opened 2022-10-03
Aurigami Aurora Investigate issues


dmelotik:
Investigate Aurigami issues so we can get the subgraph prod ready (requested by @KremerMichael).

Known issues:
- crazy high prices

dmelotik:
The Aurigami price oracle does not work for PLY, USN, AURORA, and TRI. Will need some creative pricing to get the prices here.

in addition rewards are also missing.

dmelotik:
Finished

PR #1227, opened 2022-10-19
fix(#patch); aurigami; fixing price / rewards

dmelotik:
Fixes:
- Get price of 4 markets from LPs
- update to version 1.1.7

Rewards explanation: #1273 

Testing: https://okgraph.xyz/?q=dmelotik%2Faurigami-aurora

Relating to #1133


subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -302,11 +323,138 @@ function getOrCreateProtocol(): LendingProtocol {
...
+
+function getPrice(
+  marketAddress: Address,
+  priceOracle: string
+): ethereum.CallResult<BigInt> {
+  //
+  //
+  // There are 4 markets where the price oracle does not work
+  // PLY, AURORA, TRI, USN
+  // Use Trisolaris LP pool pairs to derive price
```
jaimehgb, line 347: You could remove this comment by rewriting the `if`'s below in a self descriptive way like:
```
if (marketUnuspportedByOracle(marketAddress)) {
  return getMarketPriceFromLp(marketAddress);
}
```

Then remove boilerplate like:
```
const customOracle = constants.CustomOracles[marketAddress];
return getPriceFromLP(priceOracle, customOracle.KnownMarket, marketAddress, customOracle.LP);
```

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -302,11 +323,138 @@ function getOrCreateProtocol(): LendingProtocol {
...
+//
+//
+// Gets the price of wantAddress.inputToken from an LP from Trisolaris
+function getPriceFromLp(
+  priceOracle: string, // aurigami price oracle
+  knownMarketID: Address, // address of the market we know the price of
+  wantAddress: Address, // market address of token we want to price
+  lpAddress: Address, // address of LP token
+  whichPrice: boolean // true to find token0 price in LP, false for token1
+): ethereum.CallResult<BigInt> {
```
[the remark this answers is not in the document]
dmelotik: We need to pass it into the handleAccrueInterest (shared src/mapping.ts) and this is the format.

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -302,11 +323,138 @@ function getOrCreateProtocol(): LendingProtocol {
...
+  if (whichPrice) {
+    const reserveBalance0 = tryReserves.value.value0
+      .toBigDecimal()
+      .div(exponentToBigDecimal(wantMarketDecimals));
+    const reserveBalance1 = tryReserves.value.value1
+      .toBigDecimal()
+      .div(exponentToBigDecimal(knownMarketDecimals));
+
+    // price of reserve0 = price of reserve1 / (reserve0 / reserve1)
+    priceBD = knownPriceUSD.div(reserveBalance0.div(reserveBalance1));
```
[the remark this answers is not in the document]
dmelotik: Isn't that the same thing? What i did is working so I kept it that way..

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -302,11 +323,138 @@ function getOrCreateProtocol(): LendingProtocol {
...
+
+//
+//
+// Gets the price of wantAddress.inputToken from an LP from Trisolaris
+function getPriceFromLp(
+  priceOracle: string, // aurigami price oracle
+  knownMarketID: Address, // address of the market we know the price of
+  wantAddress: Address, // market address of token we want to price
+  lpAddress: Address, // address of LP token
+  whichPrice: boolean // true to find token0 price in LP, false for token1
```
jaimehgb, line 399: I don't 100% understand why you need to tell which token from the pool to get the price from. Isn't that redundant if it will always be `wantAddress.inputToken`?

If wantAddress is token0, then you know you want the price for token0. Otherwise it's token1. Having that boolean could make someone accidentally pass it a wrong value resulting in inverted prices

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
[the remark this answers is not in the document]
jaimehgb: Ah, I see. 

NBD, but if you wanted to keep the function more "pure" you could make it return a BigInt (since it is returning a price), and then wrap the value into a CallResult wherever it is needed to be that way (in the mapping)

jaimehgb (approved):

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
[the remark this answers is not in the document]
jaimehgb: Ah I see.

NBD, but if you wanted to make the function more "pure" it would make sense to make it return a BigInt, since in the context of the function it is returning a price, so a number. And then wrap the value in a CallResult wherever it is needed that way

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
dmelotik: good point, changing this

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
dmelotik: Wouldn't I still need to have a list of if statements? In order to get the proper LP pair contract address?

jaimehgb: Well, yeah. Either that or put the constants in a map that allows you to fetch them directly. But yeah 😅

dmelotik: I am going to leave as is for now, but that is a good idea

PR #1232, opened 2022-10-19
[GOV] Add gov subgraphs to public deployments json

danielkhoo:
## Description

~As part of our governance subgraph ingestion we want to pull subgraph subgraph deployment info from a central location instead of hardcoding it. Adding a new section to the `dashboard/public/deployments.json` enables that.~

@this-username-is-taken does the additional section to this json break anything on subgraphs.messari.io? As far as I can tell it will just be ignored. I could also move it into a separate json if it make more sense

EDIT: After discussion with @this-username-is-taken and @steegecs, standardising reading of data from `https://github.com/messari/subgraphs/blob/master/deployment/deployment.json` instead.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -4122,7 +4122,7 @@
       }
     }
   },
-  "ens-governance-governance": {
+  "ens-governance": {
     "schema": "governance",
     "base": "openzeppelin-governor",
     "protocol": "ens-governance",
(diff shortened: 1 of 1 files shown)


dashboard/public/deployments.json
```diff
@@ -258,5 +258,37 @@
         "lido": {
             "ethereum": "https://api.thegraph.com/subgraphs/name/messari/lido-ethereum"
         }
+    },
+    "governance": {
+        "dydx-governance": {
+            "hosted-service": "messari/dydx-governance"
+        },
+        "aave-governance": {
```
Albert-Coding, line 266: nit: alphabetize, please

Albert-Coding (approved):

Issue #1191, opened 2022-10-12
Uniswap-V2 Ethereum - Outlier in Uniswap v2 subgraph


this-username-is-taken:
https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/messari/uniswap-v2-ethereum&tab=protocol

<img width="899" alt="Screen Shot 2022-10-11 at 11 14 57 PM" src="https://user-images.githubusercontent.com/2354636/195248587-bdd3bb46-7e7f-4c76-98d0-4e1761622b62.png">

steegecs:
#1237 
Being addressed and grafting from this PR.

steegecs:
Related to issue #1391

bye43:
Closing this, will be addressed in substreams

PR #1237, opened 2022-10-20
fix(#patch); Added return and warn when 2 input tokens are given in a swap; also fixed protocol getter. 

steegecs:
**Context:**
Issue #1191
Added a warning when 2 input tokens are given in a swap. This is a very rare occurrence and would require us to make changes to the schema to handle properly. This serves as a temporary fix and ignores these swaps. I think this is okay since it is very rare that people provide 2 input tokens, and is not a normal use case of the DEX.

Additionally, I made a small fix on the protocol getter to remove an early .save() of the entity that was causing errors.


subgraphs/uniswap-forks/src/common/creators.ts
```diff
@@ -251,6 +251,21 @@ export function createSwapHandleVolumeAndFees(
   const token0 = getOrCreateToken(pool.inputTokens[0]);
   const token1 = getOrCreateToken(pool.inputTokens[1]);
 
+  if (amount0In == BIGINT_ZERO && amount1In == BIGINT_ZERO) {
```
jaimehgb, line 254: Hmm, I think it should be `!=` ? So if both amounts are != 0 then we ignore the swap

jaimehgb (approved):

Issue #1139, opened 2022-10-04
euler subgraph improvements


tnkrxyz:
Besides the possible method problem with protocolSideRevenue (#1115), a few improvements to the euler subgraph implementation:

1. The implementation relies on calling `EulerGeneralView.doQuery()` on ~100 markets on every major events. It is so inefficient  that the current implenetation has to skip 119 blocks between calling it again. Call doQuery on the market emitted the event is an easy improvement; It is also possible to access individual value from relevant events/contract(s) for even better efficiency.
2. The subgraph uses the address of underlying token as the id of the corresponding market and it is different than the specification in the schema. For example, the id for the wstETH market is [0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0](https://etherscan.io/address/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0), and it is the address for the wstETH contract. It should be the euler eTokenAddr (asset) [0xbd1bd5C956684f7EB79DA40f582cbE1373A1D593](https://etherscan.io/address/0xbd1bd5C956684f7EB79DA40f582cbE1373A1D593) or dTokenAddr (debt) [0x436548baAb5Ec4D79F669D1b9506D67e98927aF7](https://etherscan.io/address/0x436548baAb5Ec4D79F669D1b9506D67e98927aF7)

tnkrxyz:
3. Possible simplifying protocolSideRevenue calculation (from #1115):

> Realized I am not comparing apple-to-apple here: at any time, the collateral asset amount > borrow amount. If taking into account the difference, they should be equal: borrow amount * borrow APY * (1 - reserve fee ratio) = collateral asset * lending APY.

> This means the current method for calculating protocol side revenue is correct, but more complex than needed
> `Market Deposits Balance * ((Market Supply APY / (1 - Market Reserve Fee)) - Market Supply APY)`

> =

> `Borrow Balance * Borrow APY * Reserve Fee`

dmelotik:
@tnkrxyz I wonder if there is a clean way to get revenue from euler. For example in compound we have `AccrueInterest()`. When looking at this a while back I couldn't find anything, but maybe it is more convoluted.

dmelotik:
This might be out event to watch https://github.com/euler-xyz/euler-contracts/blob/master/contracts/Events.sol#L32

And then every time it is emitted we take the difference from the last to the current value of `interestAccumulator` then multiply by the spot price of the asset. lmk what you think

[...]

PR #1242, opened 2022-10-20
fix (#minor); euler-finance; Refactor euler subgraph

tnkrxyz:
Refactor euler-subgraph to improve efficiency/syncing speed & improve robustness (reduce possibility of oneshot error) and address issues/enhancement listed in #1139

A test deployment is at https://thegraph.com/hosted-service/subgraph/tnkrxyz/test-subgraph?version=pending&selected=logs
validation dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmZVeMSKWTie1r9mS7DCQuSH7TueXe2rrcMmCdQM2sBnWW&tab=protocol

The syncing speed is not as impressive as I hyped - that was when I only update the price for market emitted AssetStatus. The code now updates prices for all markets every ~20 minutes and finishes syncing in ~2 hours.


subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+  if (totalBalances.notEqual(eTokenTotalSupply)) {
+    log.warning("[logAssetStatus]market={}/{},totalBalances={},eTokenTotalSupply={}", [
+      market.name!,
+      market.id,
+      totalBalances.toString(),
+      eTokenTotalSupply.toString(),
+    ]);
+  }
+
+  // update prices, tvl, totalDepositBalanceUSD, totalBorrowBalanceUSD every 100 blocks (~20 min)
```
dmelotik, line 125: can we make this between 10-15 mins. Our cron job that retrieves subgraph data from the subgraphs runs every 15 mins (ideally this would be updated every time)

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+      mrkt.totalDepositBalanceUSD = bigIntToBDUseDecimals(mrkt.inputTokenBalance, tkn.decimals).times(
+        underlyingPriceUSD,
+      );
+      mrkt.totalBorrowBalanceUSD = bigIntToBDUseDecimals(mrkt._totalBorrowBalance!, tkn.decimals).times(
+        underlyingPriceUSD,
+      );
+      mrkt.totalValueLockedUSD = mrkt.totalDepositBalanceUSD;
+      mrkt.save();
+
+      log.debug(
```
dmelotik, line 147: again, I am guessing this was for dev? Do you think it should be included in the master version?

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+    }
+
+    protocol.totalDepositBalanceUSD = totalDepositBalanceUSD;
+    protocol.totalBorrowBalanceUSD = totalBorrowBalanceUSD;
+    protocol.totalValueLockedUSD = totalDepositBalanceUSD;
+    protocol._lastUpdateBlockNumber = event.block.number;
+    protocol.save();
+  }
+
+  log.debug("[handleAssetStatus]block={},protocol.totalBorrowBalanceUSD={},protocol.totalDepositBalanceUSD={}", [
```
dmelotik, line 171: same here

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+    event.block.number.toString(),
+    protocol.totalBorrowBalanceUSD.toString(),
+    protocol.totalDepositBalanceUSD.toString(),
+  ]);
+
+  snapshotFinancials(event.block, BIGDECIMAL_ZERO, null, protocol);
+
+  assetStatus.totalBorrows = totalBorrows;
```
dmelotik, line 179: I'm going to argue that this would increase readability if you put this after you grab those values from the event. @jaimehgb do you have any thoughts here?

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
-  const queryParametersTuple = changetype<EulerGeneralView__doQueryInputQStruct>(queryParameters);
-  const eulerGeneralView = getEulerViewContract(block);
-  const result = eulerGeneralView.try_doQuery(queryParametersTuple);
-
-  if (result.reverted) {
-    return null;
-  }
-
-  return result.value;
+  //TODO: pToken?
```
dmelotik, line 288: if you want to leave this in, you should have more context so someone else could pick this up with enough context

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+  const repayUSD = bigIntToBDUseDecimals(event.params.repay, underlyingToken.decimals).times(
+    underlyingToken.lastPriceUSD!,
+  );
+  liquidation.profitUSD = liquidation.amountUSD.minus(repayUSD);
+  liquidation.save();
 
-  if (liquidation.amountUSD) {
-    market.cumulativeLiquidateUSD = market.cumulativeLiquidateUSD.plus(liquidation.amountUSD);
-    market.save();
+  log.info("[createLiquidation]seizedToken={}/{},amount={},price={},amountUSD={},blk={},tx={}", [
```
dmelotik, line 194: seems like another dev log that we could remove

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
-
-        market.outputTokenPriceUSD = eTokenPriceUSD;
-        market.outputTokenSupply = eTokenTotalSupply;
-        market.exchangeRate = eulerViewMarket.totalBalances
-          .toBigDecimal()
-          .div(tokenPrecision)
-          .div(eTokenTotalSupply.toBigDecimal().div(eTokenPrecision));
-      }
-    }
+  log.debug("[updatePrices]tx={},block={},token={}/{},currPriceUSD={}", [
```
dmelotik, line 261: same

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+  lenderRate.save();
+  market.rates = [borrowerRate.id, lenderRate.id];
+  market.save();
 
-    const reserveAPY = supplyAPY
-      .div(BIGDECIMAL_ONE.minus(eulerViewMarket.reserveFee.toBigDecimal().div(RESERVE_FEE_SCALE)))
-      .minus(supplyAPY);
-    const reserveYield = reserveAPY.times(secondsSinceLastUpdate.toBigDecimal()).div(SECONDS_PER_YEAR);
-    const protocolSideRevenueSinceLastUpdate = reserveYield.times(market.totalDepositBalanceUSD);
+  log.debug(
```
dmelotik, line 305: same

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+  assetStatus: _AssetStatus,
+  event: ethereum.Event,
+): void {
+  const marketId = market.id;
+  const underlying = Address.fromString(market.inputToken);
+  const block = event.block;
+  const timestamp = event.block.timestamp;
+  const token = getOrCreateToken(underlying);
+
+  if (assetStatus.timestamp && timestamp.lt(assetStatus.timestamp)) {
```
dmelotik, line 350: is there a way we can handle this without halting indexing?

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
 
-    updateSnapshotRevenues(
-      market.id,
-      block,
-      supplySideRevenueSinceLastUpdate,
-      protocolSideRevenueSinceLastUpdate,
-      totalRevenueSinceLastUpdate,
-    );
+  if (assetStatus.reserveBalance && reserveBalance.lt(assetStatus.reserveBalance)) {
+    log.critical("[updateRevenue]event reserveBalance {} < assetStatus.reserveBalance {} at block {} for tx {}", [
```
dmelotik, line 360: same

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+  market.cumulativeSupplySideRevenueUSD = market.cumulativeSupplySideRevenueUSD.plus(deltaSupplySideRevenue);
+  market.cumulativeProtocolSideRevenueUSD = market.cumulativeProtocolSideRevenueUSD.plus(deltaProtocolSideRevenue);
+  market.cumulativeTotalRevenueUSD = market.cumulativeTotalRevenueUSD.plus(deltaTotalRevenue);
+  market.save();
 
-    //market daily/hourly snapshots are updated in the handler function after updateProtocolAndMarkets() call
-  }
-  protocol.save();
-}
+  log.info(
```
dmelotik, line 397: same

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+    } else if (eventType == TransactionType.WITHDRAW) {
+      financialMetrics.dailyWithdrawUSD = financialMetrics.dailyWithdrawUSD.plus(amountUSD);
+    } else if (eventType == TransactionType.LIQUIDATE) {
+      financialMetrics.dailyLiquidateUSD = financialMetrics.dailyLiquidateUSD.plus(amountUSD);
+    }
+  }
+
+  financialMetrics.save();
+  const days = (block.timestamp.toI64() / SECONDS_PER_DAY).toString();
+  log.debug("[snapshotFinancials]days {}: protocol.totalValueLocked={}", [
```
dmelotik, line 489: same

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+export function snapshotMarket(
+  block: ethereum.Block,
+  marketId: string,
+  amountUSD: BigDecimal,
+  eventType: string | null = null,
+): void {
+  const marketDailyMetrics = getOrCreateMarketDailySnapshot(block, marketId);
+  const marketHourlyMetrics = getOrCreateMarketHourlySnapshot(block, marketId);
+
+  //if (block.timestamp.ge(marketDailyMetrics.timestamp)) {
```
dmelotik, line 562: remove these unused lines

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -162,316 +139,607 @@ export function createRepay(event: Repay): BigDecimal {
...
+    marketDailySnapshot.cumulativeTotalRevenueUSD.minus(deltaTotalRevenueUSD);
+  marketDailySnapshot.cumulativeSupplySideRevenueUSD =
+    marketDailySnapshot.cumulativeSupplySideRevenueUSD.minus(deltaSupplySideRevenueUSD);
+  if (marketDailySnapshot.dailyTotalRevenueUSD.ge(deltaTotalRevenueUSD)) {
+    marketDailySnapshot.dailyTotalRevenueUSD = marketDailySnapshot.dailyTotalRevenueUSD.minus(deltaTotalRevenueUSD);
+    marketDailySnapshot.dailySupplySideRevenueUSD =
+      marketDailySnapshot.dailySupplySideRevenueUSD.minus(deltaSupplySideRevenueUSD);
+    marketDailySnapshot.save();
+  } else {
+    log.error(
```
dmelotik, line 715: this seems like another dev log, but if you think it should be included then it should also be logged for the hourly snapshot.

dmelotik (commented):
Would it make more sense to update the price on a transaction event handler (just for that market) since the asset status occurs after that? Then you wouldn't have to update the prices every x blocks. Thoughts?

subgraphs/euler-finance/src/mappings/handlers.ts
jaimehgb: Hmm, I agree in a way because the function is very big, in my opinion, and after reading through it all, when you reach the end, you might not have clear where the `totalBorrows` value came from. 

But if we grabbed the value from the event, we should do it too for `totalBalances`, `reserveBalance`, `underlying`, `poolSize`, ... to keep it consistent. It might become too verbose, so I guess it's matter of preference.

What I'd definitely do is breaking down the function into smaller ones, reducing its size so that it doesn't matter if you use `totalBorrows` or `event.params.totalBorrows` because either way will be clear enough.

tnkrxyz:
> Would it make more sense to update the price on a transaction event handler (just for that market) since the asset status occurs after that? Then you wouldn't have to update the prices every x blocks. Thoughts?

I tried this in an earlier version, only updating the market is a lot faster (synced in ~45 mins), but the prices become stale for markets that are not actively. I think it is a good balance in the current implementation. 
 
> Also, there are a few linting things. And nice job on the refactor! A lot of my comments are small and should be quick resolves.
thx!

subgraphs/euler-finance/src/common/getters.ts
```diff
@@ -416,34 +456,44 @@ export function getOrCreateInterestRate(rateSide: string, rateType: string, mark
   return rate;
 }
 
+//TODO: DELETE from here
```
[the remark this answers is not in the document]
tnkrxyz: Good catch!

subgraphs/euler-finance/protocols/euler-finance/config/templates/euler.template.yaml
```diff
@@ -14,12 +14,12 @@ dataSources:
     name: euler
     network: mainnet
     source:
-      address: "{{ factoryContract }}"
+      address: "0x27182842E098f60e3D576794A5bFFb0777E025d3"
```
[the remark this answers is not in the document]
tnkrxyz: fixed.

subgraphs/euler-finance/protocols/euler-finance/config/templates/euler.template.yaml
```diff
@@ -62,4 +64,4 @@ dataSources:
           handler: handleRepay
         - event: Withdraw(indexed address,indexed address,uint256)
           handler: handleWithdraw
-      file: ./src/mappings/handlers.ts
+      file: ./src/mappings/handlers.ts
```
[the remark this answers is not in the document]
tnkrxyz: Fixed.

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
-import { Address, ethereum } from "@graphprotocol/graph-ts";
+import { Address, BigDecimal, log } from "@graphprotocol/graph-ts";
```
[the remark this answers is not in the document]
tnkrxyz: Uneeded, deleted

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+    const dTokenTotalSupply = dToken.totalSupply();
+    if (dTokenTotalSupply.gt(BIGINT_ZERO)) {
+      market._dTokenExchangeRate = bigIntToBDUseDecimals(totalBorrowBalance, token.decimals).div(
+        bigIntToBDUseDecimals(dTokenTotalSupply, DEFAULT_DECIMALS),
+      );
+    }
+  }
+  market.save();
+
+  if (interestRate.notEqual(assetStatus.interestRate)) {
```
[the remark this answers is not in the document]
tnkrxyz: Updating interestRate when reserveFee changes is invoked in `handleGovSetReserveFee()`.

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -1,170 +1,316 @@
...
+    // update interest rates if `interestRate` or `reserveFee` changed
+    updateInterestRates(market, interestRate, assetStatus.reserveFee, totalBorrows, totalBalances, event);
+  }
+  updateRevenue(reserveBalance, totalBalances, totalBorrows, protocol, market, assetStatus, event);
+  snapshotMarket(event.block, marketId, BIGDECIMAL_ZERO, null);
+
+  //for verification only
```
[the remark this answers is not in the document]
tnkrxyz: Deleted

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: Changed to 75 blocks (~15 minutes)

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: My preference is to keep them - users can filter what log level to show. Having them will help debug issues in the future.

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: Moved up the chunk & put some of the code into functions.

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: I believe there is no need to handle pToken in our subgraph. Deleted.

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: deleted

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: deleted

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: deleted

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: It is unlikely to happen, I am deleting them.

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: deleted

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: deleted

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: simplified & added.

PR #1252, opened 2022-10-21
fix(#patch); Update ubeswap version in deployment.json

steegecs:
**Context:** 
Deployment.json version was wrong for Ubeswap.


mrbrianhobo (approved):
🍠 🔄 LGTM

PR #1255, opened 2022-10-23
fix(#minor); Uniswap V3; Updated to use Collect instead of burn. Also, configured to use deployment versions

steegecs:
**Context:**
We recently incorporated collect events into Uniswap V3 in attempt to fix TVL issues that were occurring in PR #1092. I made the change based on a PR the Uni team was using to address this issue.

After making these changes, however, we were getting negative token values. I have discovered that when incorporating both the `burn` and the `collect` events, tokens are being removed from liquidity pools, in a sense, twice. This is because the `amount0` and `amount1` emitted from the burn event do not refer to the actual amount of tokens leaving the liquidity pool, rather, it refers to an increasing balance of tokens owed to the user that can be collected.

`Burn` and `collect` are often emitted together, and with the same amounts, but this is not guaranteed, and the `burn` event does not account for tracking the token pair flows out of the pool that were earned from accrued fees while `collect` events do.

By removing the `burn` event and retaining tracking withdraws with the `collect` event, we can maintain a proper balance of the liquidity pools over time. I have tested this out with Uniswap V3 Celo, and it has remained accurate when compared with token balances against etherscan until fully synced. 

Changes
- [x] Removed the use of the Burn event and instead track withdraws from the liquidity pools using the collect event. 
- [x] Update the protocol getter to use the `versions.ts` generated from the deployment.json

### DIFF
--- a/subgraphs/uniswap-v3/src/mappings/core.ts
+++ b/subgraphs/uniswap-v3/src/mappings/core.ts
@@ -1,6 +1,5 @@
 // import { log } from "@graphprotocol/graph-ts";
 import {
-  Burn as BurnEvent,
   Initialize,
   Mint as MintEvent,
   Swap as SwapEvent,
--- a/subgraphs/uniswap-v3/src/mappings/core.ts
+++ b/subgraphs/uniswap-v3/src/mappings/core.ts
@@ -45,21 +45,6 @@ export function handleMint(event: MintEvent): void {
   updatePoolMetrics(event);
 }
 
-// Handle a burn event emitted from a pool contract. Considered a withdraw into the given liquidity pool.
-export function handleBurn(event: BurnEvent): void {
-  createWithdraw(
-    event,
-    event.params.owner,
-    event.params.owner,
-    event.params.amount0,
-    event.params.amount1,
-    false
-  );
-  updateUsageMetrics(event, event.params.owner, UsageType.WITHDRAW);
-  updateFinancials(event);
-  updatePoolMetrics(event);
-}
-
 // Handle a swap event emitted from a pool contract.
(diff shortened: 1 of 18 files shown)


subgraphs/uniswap-v3/src/mappings/core.ts
```diff
@@ -47,17 +47,17 @@ export function handleMint(event: MintEvent): void {
...
+  //   event,
+  //   event.params.owner,
+  //   event.params.owner,
+  //   event.params.amount0,
+  //   event.params.amount1,
+  //   false
+  // );
+  // updateUsageMetrics(event, event.params.owner, UsageType.WITHDRAW);
+  // updateFinancials(event);
+  // updatePoolMetrics(event);
```
jaimehgb, line 60: Don't be afraid to delete. If anything it will be in the git history :D

jaimehgb (approved):

subgraphs/uniswap-v3/src/versions.ts
```diff
@@ -0,0 +1,17 @@
+import { Versions as VersionsInterface } from "../../../deployment/context/interface";
```
jaimehgb, line 1: This file isn't in the gitignore?

subgraphs/uniswap-v3/src/mappings/core.ts
steegecs: Deleted.

subgraphs/uniswap-v3/src/versions.ts
steegecs: I spelled it wrong. Fixed it!

PR #1256, opened 2022-10-23
feat(); CLI; Configure CLI so it can be installed from the npm registry

steegecs:
**Context:**
Currently, users of the CLI have to install it by navigating to the messari-cli folder and installing it locally. This PR configures the CLI so that it can be published and installed from the npm registry. 

Changes:
- [x] Moved some dev dependencies over to dependencies so they will be installed when you install the CLI.
- [x] Updated Env path derivation to the messari repo
- [x] Moved validation script into the CLI helper code.


messari-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "messari-subgraph-cli",
-  "version": "0.0.3",
+  "version": "0.1.39",
```
jaimehgb, line 3: What a giant bump haha

jaimehgb (approved):

messari-cli/package.json
steegecs: :)

PR #1258, opened 2022-10-23
fix(#patch); CI Linter; Fix changed files action 

steegecs:
Context:
Sometime the linter was not able to fetch the commit prior to the PR causing it to fail. This PR updates the changed file action to one that has worked better in the past.

Screen Shot 2022-10-23 at 3 56 06 PM


jaimehgb (approved):
🚀

Issue #1087, opened 2022-09-28
Using `rewards.ts` to dynamically get blocks/day on arbitrum does not work


dmelotik:
Trying to dynamically get blocks/day on arbitrum does not work due to the way block numbers are returned from the arbitrum one RPC: https://github.com/OffchainLabs/arbitrum/blob/master/docs/Time_in_Arbitrum.md

In order to find a working solution we could just hard code the block interval for arbitrum. Although, in the past it has been variable based on (https://arbiscan.io/chart/blocktime). The other option is to see if we can still figure out a way to get this dynamically using what we found above. 

Outcomes:
- However we choose to address this, all arbitrum subgraphs that used dynamic blocks/day calculation will need to be switched

tnkrxyz:
This affects dforce-arbitrum (to be fixed in PR #1049) and cream-finance-arbitrum (no fix yet).

dmelotik:
Looks like #1049 fixes some of them. It won't fix cream finance arbitrum, since it uses `rewards.ts` to get blocks/year. I reckon we should look across all of messari/subgraphs to fix any of this. I think rari fuse arb is another affected one.

tnkrxyz:
cream finance can be fixed in a similar way - by setting ARBITRUM_BLOCKS_PER_YEAR = ETHEREUM_BLOCKS_PER_YEAR in compound-forks/src/constants.ts. The fix in #1049 only fixed it for dforce because the change was made in compound-forks/protocols/dforce/src/constants.ts . Not sure about rari etc

dmelotik:
We still need to go through and ensure all the compound forks deployed to arbitrum are not affected. ie, I know rari fuse arbitrum needs to get off using reward.ts

steegecs:
Hows this coming along? Did you find a fix? Do you need help?

tnkrxyz:
@dmelotik , lmk if I can help with this.

dmelotik:
I think I will get it done rq this week since I know exactly what needs to happen. I will assign you to review @tnkrxyz, thanks

PR #1262, opened 2022-10-24
fix(#patch); rari fuse; fix arbitrum blocks/day

melotik:
#1087 

Test: https://okgraph.xyz/?q=dmelotik%2Ffuse-official

_Note_: This should be pused to prod once merged

### DIFF
--- a/subgraphs/compound-forks/protocols/rari-fuse/src/mappings.ts
+++ b/subgraphs/compound-forks/protocols/rari-fuse/src/mappings.ts
@@ -502,6 +502,7 @@ export function handleAccrueInterest(event: AccrueInterest): void {
   const oracleContract = PriceOracle.bind(Address.fromString(pool.priceOracle));
 
   let blocksPerYear: i32;
+  let blocksPerDay: BigDecimal;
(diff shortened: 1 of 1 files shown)


jaimehgb (approved):

subgraphs/compound-forks/src/versions.ts
```diff
@@ -0,0 +1,17 @@
...
+  getSubgraphVersion(): string {
+    return "1.1.9";
+  }
+
+  getMethodologyVersion(): string {
+    return "1.0.0";
+  }
+}
+
+export const Versions = new VersionsClass();
```
tnkrxyz, line 17: Does this change mean that all compound-forks subgraphs will have the exact same versions? Because versions.ts is in compound-forks/src, and there is no protocol-specific versions.ts file.

dmelotik: @tnkrxyz I just made a PR to remove this #1361

This is an auto-generated file that helps update versions automatically based on the deployment.json version. It shoudln't be in master, but you will notice it generates each time you use messari b -d

Issue #1081, opened 2022-09-27
New lending schema (Positional / new relations in markets)


dmelotik:
As we have developed lending protocol subgraphs, we have learned numerous things that will help us refine our lending positional data.

Learnings:
- Not every pool borrows the same asset as is in the pool. Take abracadabra for example.
- We learned that positions can be passed around outside of protocol owned contracts. cTokens can be transfered to wallets that never touch a market. Is this a new position, do we increment deposits, new user? We need to answer these questions
- Aave v2 is tough and needs more research to understand how swaps within the protocol create new positions.
- Lots of notes have been made here #389 
- SimpleFi has already done some work on this, planning to connect with them to learn more and collaborate

ishraq8:
Thanks for this, do you mind if i join the conversations with simplefi? i was going through the docs you shared.

https://github.com/SimpleFi-finance/subgraphs

dmelotik:
Linking notion of proposed changes: https://www.notion.so/messari/Lending-Protocol-Schema-cbc6f46a75044bb5829e29d12f7004d2

ishraq8:
Making note of issue https://github.com/messari/subgraphs/issues/966

https://twitter.com/omeragoldberg/status/1568118558982389760?s=46&t=YqJRaYypMHCTU9e4e92Qhg
https://chaoslabs.xyz/posts/chaos-labs-dives-deep-into-aave-v3-data-validity

For AAVE V3 positional data when withdrawing a supply position the contract burns the atokens withdrawn-the interest accumulated since last update. So for positional data someone can have no supply but still have atokens. 

Since our subgraph is tracking both deposit/withdraw and atokens seperately we can just make a note in the docs to address this potential descrepancy for aave v3.

dmelotik:
This is done now

PR #1265, opened 2022-10-25
feat(#major); lending schema 3.0.0

dmelotik:
New 3.0.0 lending schema #1081


schema-lending.graphql
```diff
@@ -80,6 +83,12 @@ enum LendingType {
 
   " Pooled protocols pool all users assets into a single market "
   POOLED
+
+  " Protocols that only allow KYC'd users "
+  WHITELIST_ONLY
```
this-username-is-taken, line 88: I would separate this out into another field (e.g. `PermissionType`) as it's a different dimension

schema-lending.graphql
```diff
@@ -529,11 +598,13 @@ type Market @entity {
   " Liquidation penalty (or the liquidation bonus for liquidators) as a percentage value. It is the penalty/bonus price on the collateral when liquidators purchase it as part of the liquidation of a loan that has passed the liquidation threshold "
   liquidationPenalty: BigDecimal!
 
-  " Token that need to be deposited to take a position in protocol "
-  inputToken: Token!
+  " Address of the price oracle used in this market "
+  priceOracle: Bytes
```
this-username-is-taken, line 602: We should keep the data type for addresses consistent across the schema, either as Bytes or as String

schema-lending.graphql
```diff
@@ -570,19 +641,78 @@ type Market @entity {
...
+  " Amount of input tokens borrowed in this market using variable interest rates (in native terms) "
+  variableBorrowedTokenBalance: BigInt
 
-  " Price per share of output token in USD "
-  outputTokenPriceUSD: BigDecimal!
+  " Amount of input tokens borrowed in this market using stable interest rates (in native terms) "
+  stableBorrowedTokenBalance: BigInt
+
+  " Total supply of output token (same order as outputTokens) "
+  outputTokensSupply: [BigInt!]
```
this-username-is-taken, line 712: I've had request before to change related fields into an array of structs instead of arrays of multiple fields, for example, something like:
```
type OutputToken @entity {
  supply: BigInt!
  priceUSD: BigDecimal!
  ...
}
```
This can eliminate the need to sort the items in the array to ensure alignment.

this-username-is-taken:
Great start! You might've already done it but might be worth going through this issue to make sure everything has been accounted for: https://github.com/messari/subgraphs/issues/389

@ishraq8 would this be a good time to consider splitting revenue into more detailed categories?

schema-lending.graphql
this-username-is-taken: If we change addresses to Bytes, might be a good time to change all IDs to Bytes as well, for performance improvements documented here: https://medium.com/edge-node-engineering/two-simple-subgraph-performance-improvements-a76c6b3e7eac

schema-lending.graphql
```diff
@@ -529,8 +568,8 @@ type Market @entity {
   " Liquidation penalty (or the liquidation bonus for liquidators) as a percentage value. It is the penalty/bonus price on the collateral when liquidators purchase it as part of the liquidation of a loan that has passed the liquidation threshold "
   liquidationPenalty: BigDecimal!
 
-  " Token that need to be deposited to take a position in protocol "
-  inputToken: Token!
+  " Tokens that need to be deposited in this market to take a position in protocol (should be alphabetized) "
+  inputTokens: [Token!]!
 
   " Token that is minted to track ownership of position in protocol (e.g. aToken, cToken). Leave as null if doesn't exist "
   outputToken: Token
```
tnkrxyz: I am not how common this is that it is worthwhile to change our schema to support it, but euler-finance has two outputTokens per market, eToken (asset) and dToken (debt). I am tracking dToken with a customized field '_dToken' in the euler refactor. If we want to supply this, then we'll have to change outputTokenSupply and outputTokenPriceUSD to an array too.

schema-lending.graphql
```diff
@@ -570,19 +609,48 @@ type Market @entity {
...
   " Total supply of output token "
-  outputTokenSupply: BigInt!
+  outputTokenSupply: BigInt
 
   " Price per share of output token in USD "
-  outputTokenPriceUSD: BigDecimal!
+  outputTokenPriceUSD: BigDecimal
 
   " Amount of input token per full share of output token. Only applies when the output token exists (note this is a ratio and not a percentage value, i.e. 1.05 instead of 105%) "
   exchangeRate: BigDecimal
```
tnkrxyz, line 718: Again, if we are to supply multiple outputToken, then exchangeRate will need to be an array too.

tnkrxyz:
Another special case of outputToken (not sure we want to support): in goldfinch, the outputToken from their tranched pools is an ERC721 token (https://dev.goldfinch.finance/docs/reference/contracts/overview).

this-username-is-taken:
also, thoughts on integrating this PR in?
https://github.com/messari/subgraphs/pull/962

schema-lending.graphql
dmelotik: created

schema-lending.graphql
dmelotik: If we want to keep the pattern {transaction count}-{hash}-{logindex} it won't be possible to make that of type Bytes.

A few options:
- We don't use that format (I'm against this)
- Or we just use Bytes on the IDs that we can

schema-lending.graphql
dmelotik: Added `TokenBalances` to clean this up and allow for more custom relationships within markets

dmelotik: See `TokenBalances` and let me know what you think

schema-lending.graphql
dmelotik: resolved by adding `TokenBalances` type

schema-lending.graphql
dmelotik: same

dmelotik:
> @ishraq8 would this be a good time to consider splitting revenue into more detailed categories?

An idea I had was to keep the revenues the same as they are currently, but to add a new type called `Revenues` and this will basically just allow you to specify the sources and balances of each in USD.

> also, thoughts on integrating this PR in? https://github.com/messari/subgraphs/pull/962

immutable added where possible ✅  / And the oracle for each market is added ✅  / How useful is the extra oracle information? From my perspective we want to add it mostly to keep track of the current oracle and show people where prices come from. Thoughts @ishraq8 @data-nexus

> @dmelotik What you think about adding the flashloan revenue, we talked about it doing it next update.

Imo, once this PR is merged I think we should reassess methodology on lending protocols with flashloans in order to determine if we need to include flashloans as revenue (ie, imo aave should include it in revenue)

ishraq8:
> How useful is the extra oracle information? From my perspective we want to keep track of the current oracle and show people where prices come from. Thoughts @ishraq8 @Data-Nexus

Considering oracle prices are being manipulated for a number of exploits (Mango Markets and Moo Finance in past two weeks), I think this data is useful for users.

dmelotik:
> Another special case of outputToken (not sure we want to support): in goldfinch, the outputToken from their tranched pools is an ERC721 token

@tnkrxyz do you have a recommended change to allow fit this in? I am not super familiar with goldfinch or tranches

schema-lending.graphql
```diff
@@ -205,6 +205,48 @@ type RevenueDetails @entity {
...
+  blockLU: BigInt!
+
+  " The timestamp this oracle was adopted for a market "
+  timestampCreated: BigInt!
+
+  " The most recent timestamp that the oracle was updated (Updated when the oracle address is updated) "
+  timestampLU: BigInt!
+
+  " Is the Oracle currently used as the source of truth for a market"
+  isActive: Boolean
```
this-username-is-taken, line 238: why not required?

schema-lending.graphql
```diff
@@ -205,6 +205,48 @@ type RevenueDetails @entity {
...
+
+  " True if the oracle returns prices in USD (e.g. generally the other case is the network's native token) "
+  isUSD: Boolean
```
this-username-is-taken, line 241: why not required?

schema-lending.graphql
```diff
@@ -205,6 +205,48 @@ type RevenueDetails @entity {
...
+}
+
+# Most lending protocols get token prices through onchain oracles.
+# As oracles are a source of truth it is important to understand the details.
+# This entity will help track oracle-related data in a market
+type Oracle @entity {
+  " { Market ID } "
+  id: ID!
+
+  " The block this oracle was adopted for a market "
```
this-username-is-taken, line 225: add `market` as a field

schema-lending.graphql
```diff
@@ -141,6 +155,101 @@ type InterestRate @entity {
...
+
+  " Revenue that is generated from fees from flashloans "
+  FLASHLOAN_FEE
+}
+
+# This entity offers a more nuanced view of revenue types
+# Use this to provide the sources of revenue and amounts of each
+type RevenueDetails @entity {
+  " { Market/Protocol ID } "
+  id: ID!
```
dmelotik, line 202: consider making this market-revenue type so that we don't have to keep the arrays in a specific order, then adding a market field here to derive this into each market.

Should this be at the protocol level?

schema-lending.graphql
dmelotik: Removed transaction count prefix per convo w/ suds and opted for Bytes wherever possible

schema-lending.graphql
dmelotik: Updated token balances to live as an array of structs in `market`. This essentially creates a mapping of inputToken:(multiple output tokens)

schema-lending.graphql
dmelotik: done

schema-lending.graphql
```diff
@@ -82,6 +91,17 @@ enum LendingType {
...
+  PERMISSIONED
+
+  " Protocols that allow a mix of KYC'd and non-KYC'd users "
+  PARTIALLY_PERMISSIONED
+
+  " Protocols that do not KYC. Can be used by any account "
+  PERMISSIONLESS
+}
+
 enum RiskType {
```
this-username-is-taken, line 105: Should we add something like a "Collateralization Type" to describe Over-collateralized, Under-collateralized, and uncollateralized lending protocol types?

schema-lending.graphql
```diff
@@ -82,6 +91,17 @@ enum LendingType {
   POOLED
 }
 
+enum PermissionType {
+  " Protocols that only allow KYC'd users "
+  PERMISSIONED
```
this-username-is-taken, line 96: I think this might need to be more nuanced. For example, there are protocols that allow *any* KYC-ed users, but some that allow only whitelisted users (e.g. institutions).

Also, I think we should describe permission for lender and borrowers separately. For example, a protocol can allow anyone to lend, but only KYCed users to borrow.

schema-lending.graphql
```diff
@@ -110,8 +140,8 @@ enum InterestRateSide {
 }
```
tnkrxyz, line 141: In addition to InterestRateType and InterestRateSide, there is need for an optional third enum for interest rate id: InterestRateTranche, to support tranched pools (Goldfinch, Bainbridge, tranched, etc) as each tranche has different rates.
```
enum InterestRateTranche {
      SENIOR
      JUNIOR
}
```

schema-lending.graphql
```diff
@@ -141,18 +171,151 @@ type InterestRate @entity {
   type: InterestRateType!
 }
```
tnkrxyz, line 172:
```
tranche: InterestRateTranche
```

schema-lending.graphql
```diff
@@ -141,18 +171,151 @@ type InterestRate @entity {
...
+  inputTokenBalance: BigInt!
+
+  " Prices in USD of the input tokens (same order as inputTokens) "
+  inputTokenPricesUSD: BigDecimal!
+
+  " Tokens that are minted to track ownership of position in protocol (e.g. aToken, cToken). Leave as null if doesn't exist (should be alphabetized) "
+  outputTokens: [Token!]
+
+  " Total supply of output tokens (same order as outputTokens) "
+  outputTokenBalances: [BigInt!]
```
tnkrxyz, line 212: -> outputTokenSupply?

schema-lending.graphql
```diff
@@ -514,34 +755,19 @@ type Market @entity {
...
 
+  " All oracles that have been used for this market "
+  oracles: [Oracle!]! @derivedFrom(field: "market")
+
   ##### Quantitative Data #####
 
-  " All interest rates / fees allowed in the market. Interest rate should be in APY percentage "
-  rates: [InterestRate!]!
+  " All fees allowed in the market. Fee should be in percentage format. e.g. 0.30% liquidation fee "
+  fees: [InterestRate!]
```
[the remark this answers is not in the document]
tnkrxyz: For lending protocols, rates are the most important factor, so maybe keep them in its own field?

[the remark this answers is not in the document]
dmelotik: @tnkrxyz I moved the rates field to `TokenData`. This was to separate rates from fees. Do you mean they should all be together again?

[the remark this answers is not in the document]
dmelotik: I can see making a second entity for fees themselves, thoughts?

schema-lending.graphql
dmelotik: I wanted to keep the same theme with inputTokenBalance. (and `outputTokenSupplies` looks kinda strange to me)

Do you have more thoughts?

jaimehgb: I kinda prefer supply 😅 

Since balances refer to the actual balance held at the market, and supply means the amount of outputTokens minted, which are not held in the market but in arbitrary wallets

schema-lending.graphql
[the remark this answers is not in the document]
tnkrxyz: Sorry - didn't catch that (fees are separated from rates), that makes sense then, but does it make sense to make it of type "InterestRate"? Using the liquidation fee example, looks like almost all interest rate id info are irrelevant: " { Interest rate side }-{ Interest rate type }-{ Market ID }-{ Optional: # days/hours since epoch time } "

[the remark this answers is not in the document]
dmelotik: I agree, I will update this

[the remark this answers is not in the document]
dmelotik: @tnkrxyz can you think of any fee types aside from liquidation, protocol, and admin?

schema-lending.graphql
```diff
@@ -323,6 +608,9 @@ type UsageMetricsDailySnapshot @entity {
   " ID is # of days since Unix epoch time "
   id: ID!
```
this-username-is-taken, line 609: should we make everything `Bytes` for consistency (and perf improvement)?

dmelotik: This is what happens when you try to do something like this: 

```
  " { Market Address }-{ Input Token Address }-{ Optional: Snapshot ID } "
  id: Bytes!
```
Screen Shot 2022-11-17 at 2 35 16 AM

dmelotik: Just kidding, I was able to get it to work by converting the i32 > Address > Bytes

schema-lending.graphql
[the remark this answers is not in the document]
tnkrxyz: withdraw, flashloan?

schema-lending.graphql
dmelotik: There are some entities that have words in them though, so those will have to remain ID

schema-lending.graphql
dmelotik: using relation entity now instead, the `TokenData` seemed like a force

jaimehgb:
> WATCH out:
> 
> - We need PositionCounter bc it is used in the library.

I’ve been adding auxiliar entities for use within the library too. I think those should be named with the `_`, since they are there as helpers more than to give info to subgraph clients?
> - There are a few fields in the Event interface that I removed: position and account

👍 
> - I added LendingProtocol._markets to support iteration through all the markets :/

Not a big deal, since the number of markets is usually bounded. But if the idea is purely iterating through them you might have a dedicated entity for that.
> - Not every entity ID is in Bytes, we can do this, but then we need to abstract some of the enums into an integer mapping. To me this is not ideal, but may be okay since the nuances are always reflected in the entity (in most cases)

Sounds good to me

dmelotik:
Done @jaimehgb

jaimehgb:
Awesome! Looks good to go then :D

jaimehgb (approved):

PR #1269, opened 2022-10-25
fix(#patch): lint action to only use added & modified files

jaimehgb:
When using the `all` files the linter fails when any file is removed, since it doesn't find it.


mrbrianhobo (approved):

steegecs (approved):
Thanks for fixing. Looks good!

PR #1279, opened 2022-10-27
fix(#bug); euler-finance; fixed negative revenues

tnkrxyz:
Fixed negative rev by reverting assetStatus update to the end of handleAssetStatus()


dmelotik:
good catch!

dmelotik (approved):

dmelotik:
@tnkrxyz can you update the version? Then merge it once you are done

PR #1283, opened 2022-10-27
Chore(); CLI; Remove subgraph cli from subgraph repo

steegecs:
**Context:**
A decision was made to move the CLI into another repo. Here is it - > https://github.com/messari/subgraph-cli


jaimehgb:
You'll need to update this too for the CI not to break?
https://github.com/messari/subgraphs/blob/master/.github/workflows/build-checker.yaml#L67

steegecs:
> You'll need to update this too for the CI not to break? https://github.com/messari/subgraphs/blob/master/.github/workflows/build-checker.yaml#L67

It works! I just had to update the action so that it installed the package from the npm registry.

jaimehgb (approved):

PR #1357, opened 2022-11-02
feat(#major); add compound iii

dmelotik:
Adding Compound V3 as it's own subgraph.

Test: https://okgraph.xyz/?q=dmelotik%2Fcompound-v3-ethereum


subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+      .concat("-")
+      .concat(this.market.id.toHexString())
+      .concat("-")
+      .concat(side);
+    if (interestType) {
+      counterID = counterID.concat("-").concat(interestType);
+    }
+    let positionCounter = PositionCounter.load(counterID);
+    if (!positionCounter) {
+      positionCounter = new PositionCounter(counterID);
```
jaimehgb, line 103: This is more of a schema question I guess. Why having a PositionCounter entity instead of using the counter in the Account entity?

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+      position.depositCount += INT_ONE;
+    } else if (transactionType == TransactionType.BORROW) {
+      position.borrowCount += INT_ONE;
+    } else if (transactionType == TransactionType.TRANSFER) {
+      position.receivedCount += INT_ONE;
+    }
+    // Note: liquidateCount is not incremented here
+    position.save();
+
+    if (openPosition) {
```
jaimehgb, line 147: You like these duplicated conditionals huh 😅 

I think you can refactor it into:
```
if (!openPosition) {
  updatePositionCounter(transactionType);
  takeSnapshot();
  return;
}

// create position and updatePositionCounter() on it
// update account open position counters
// update market open position counters
// update protocol open position counters
// takeSnapshot()

File: subgraphs/compound-v3/src/sdk/manager.ts [hunk 3]
```diff
@@ -0,0 +1,902 @@
...
+    const repayer = new AccountManager(
+      account,
+      this.market,
+      this.protocol,
+      this.event
+    );
+    if (repayer.isNewAccount()) {
+      this.protocol.cumulativeUniqueUsers += INT_ONE;
+      this.protocol.save();
+    }
```
jaimehgb, line 509: As an alternative, you could potentially embed these bits into the account manager. It would not work though if we counted independent type of users (lps, borrowers, lenders, etc)

subgraphs/compound-v3/src/sdk/manager.ts
```diff
@@ -0,0 +1,902 @@
...
+    );
+    liquidateeAccount.subtractPosition(
+      newBalance,
+      PositionSide.COLLATERAL,
+      TransactionType.LIQUIDATE,
+      this.market.inputTokenPriceUSD,
+      interestType
+    );
+    // Note:
+    //  - liquidatees are not considered users since they are not spending gas for the transaction
```
jaimehgb, line 592: Not considered users or active users (on liquidation)?

They should have already added to the cumulativeUniqueUsers since in order to be liquidated you have to have lent and borrowed first

subgraphs/compound-v3/src/sdk/manager.ts
```diff
@@ -0,0 +1,902 @@
...
+    }
+    this.protocol.totalValueLockedUSD = totalValueLockedUSD;
+    this.protocol.totalBorrowBalanceUSD = totalBorrowBalanceUSD;
+    this.protocol.save();
+  }
+
+  updateRevenue(
+    newTotalRevenueUSD: BigDecimal,
+    newProtocolRevenueUSD: BigDecimal,
+    newSupplySideRevenueUSD: BigDecimal,
```
jaimehgb, line 847: Is `total` always `protocol + supply`? If so you can omit the total.

Also, from the naming it looks as if these revenues are going to replace the old ones. Maybe rename to `addRevenue`, or the arguments to revenue deltas?

subgraphs/compound-v3/src/sdk/token.ts
```diff
@@ -0,0 +1,267 @@
...
+  constructor(address: Address, symbol: string, name: string, decimals: i32) {
+    this.address = address;
+    this.symbol = symbol;
+    this.name = name;
+    this.decimals = decimals;
+  }
+
+  // Get all tokens with a static defintion
+  static getStaticDefinitions(): Array<StaticTokenDefinition> {
+    const staticDefinitions = new Array<StaticTokenDefinition>(6);
```
jaimehgb, line 192: Should this be outside the sdk? Since it is protocol specific.

jaimehgb (commented):
Looking good!

The mapping feels simpler, and the fact that most of the entity management can just be reused is convenient 💪

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+  private position!: Position;
+  private market: Market;
+  private protocol: LendingProtocol;
+  private event: ethereum.Event;
+
+  constructor(
+    account: Bytes,
+    market: Market,
+    protocol: LendingProtocol,
+    event: ethereum.Event
```
jaimehgb, line 51: You should separate concerns here. The existence of an account doesn't need to be attached to some market. The actions it might perform are, but you should be able to instantiate accounts on their own. It might be the case that all supported account actions are always related to a Market now, but if we eventually added other non-market entities with which an account can interact (say some token staking, governance stuff, etc.) this would need to be decoupled.

Instead, when calling a method where a market is necessary, pass it as an argument.

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+      _account.save();
+      this.isNew = true;
+    } else {
+      this.isNew = false;
+    }
+    this.account = _account;
+  }
+
+  // returns true if the account was created in this instance
+  isNewAccount(): boolean {
```
jaimehgb, line 80: Since you already have the protocol entity when initializing the account, you can update the unique users counters directly from here instead of exposing an `isNewAccount` method. Let the caller forget about those metrics

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+    return this.isNew;
+  }
+
+  addPosition(
+    asset: Bytes,
```
jaimehgb, line 85: The asset and the market.inputToken are the same thing? If so, you can omit it

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+    newBalance: BigInt,
+    side: string,
+    transactionType: string,
+    priceUSD: BigDecimal,
```
jaimehgb, line 89: Not a big deal, but useful nevertheless. Let the library have access to pricing. Make it depend on an interface
```
interface Pricer {
    getAmountPriceUSD(amount: BigInt, token: Token): BigDecimal
}

class AccountManager {
    ...
    prices: Pricer;
}
```
this way you can omit the price argument from everywhere.

subgraphs/compound-v3/src/sdk/account.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  getPositionID(): string | null {
+    if (!this.positonError) {
+      return this.position.id;
+    }
+    return null;
+  }
+
+  setCollateral(isCollateral: boolean): void {
```
jaimehgb, line 297: It seems it is possible for this function (and the ones below) to be called without `this.position` being set. You could prevent it by not exposing these directly from the AccountManager, but have it expose some `Account.getCurrentPositionForMarket(market: Market): PositionManager`. If the user can only modify positions via its own manager you ensure that a position is always initialized before attempting to modify it.

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
```diff
@@ -0,0 +1,1246 @@
...
+  const totalRevenueDeltaUSD = baseBorrowIndexDiff
+    .toBigDecimal()
+    .div(BASE_INDEX_SCALE.toBigDecimal())
+    .times(bigIntToBigDecimal(totalBorrowBase, inputToken.decimals))
+    .times(inputToken.lastPriceUSD!);
+  const protocolRevenueDeltaUSD = totalRevenueDeltaUSD.times(reserveFactor);
+  const supplySideRevenueDeltaUSD = totalRevenueDeltaUSD.minus(
+    protocolRevenueDeltaUSD
+  );
+  dataManager.updateRevenue(
```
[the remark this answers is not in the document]
dmelotik: clean

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
```diff
@@ -0,0 +1,1246 @@
...
+    );
+    dataManager.getOrUpdateRate(
+      InterestRateSide.LENDER,
+      InterestRateType.VARIABLE,
+      bigIntToBigDecimal(supplyRate, DEFAULT_DECIMALS)
+        .times(BigDecimal.fromString(SECONDS_PER_YEAR.toString()))
+        .times(BIGDECIMAL_HUNDRED)
+    );
+  }
+  market.save();
```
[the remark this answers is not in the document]
dmelotik: good catch, that was from the og version

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
```diff
@@ -0,0 +1,1246 @@
...
+    marketID,
+    event.params.asset,
+    event,
+    getProtocolData()
+  );
+  const accountID = event.params.dst;
+  const accountActorID = event.params.from;
+  const asset = event.params.asset;
+  const amount = event.params.amount;
+  const token = updateMarketData(market);
```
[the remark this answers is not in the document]
dmelotik: it means i want a token lmao

subgraphs/compound-v3/src/sdk/manager.ts
dmelotik: I did that bc sometimes you make an account but don't count it, like the receiver in a transaction. Bc we are only counting users that spend gas.

But that opens the door to missing users if they interact w the protocol after the fact...

subgraphs/compound-v3/src/sdk/manager.ts
dmelotik: Yeah that is true, but just wanted to reinforce that we do not count (non-gas spenders) as cumuluative unique users

subgraphs/compound-v3/src/sdk/token.ts
dmelotik: Maybe I was wrong, but I always thought these were atypical tokens on etherreum. And we used this in general, but it may make sense to remove:

ex: https://etherscan.io/address/0xbdeb4b83251fb146687fa19d1c660f99411eefe3#readContract

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: Not always, for example, you could borrow MIM in a WETH market in abracadabra

subgraphs/compound-v3/src/sdk/manager.ts
dmelotik: this is true, fixed

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: The `positionCount` counter?

dmelotik: I don't think there is a counter in the account entity that would allow us to know which position to call if it is still open and positions are opened afterwards.

What I mean is that we will lose count of old positions as new ones open before the old ones are closed

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: Nice, I changed that

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: I guess this is the same thing as this: https://github.com/messari/subgraphs/pull/1357#discussion_r1041654183

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: I see what you are doing here. What would you expect to happen if the pricing function implementation can't get the price? $0 or an error?

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: True, and you may want to set something in the position even when you are not adding or subtracting to the position. Fixed in most recent commit

subgraphs/compound-v3/src/sdk/account.ts
jaimehgb: Yeah, my bad. Resolving this one

subgraphs/compound-v3/src/sdk/account.ts
jaimehgb: Ah right, that makes sense

subgraphs/compound-v3/src/sdk/account.ts
jaimehgb: Hmm, that’s a good point. Do we ever abort some process if we cannot fetch a price?

To keep things simple I would return `0` and work with that. If we realise it is necessary to sometimes abort when prices cannot be fetched we can add that functionality. Most times if the price lib returns 0 we just go with that value AFAIK, so it shouldn’t matter.

If when building a subgraph we wanted to not create a position if unable to fetch the price, we could make it return a `class { price: BigDecimal, error: boolean}` instead.

Another option is exposing both methods, one where you can pass the price and one which uses the pricer. But I think this can get confusing for a user of the lib.

 What do you think?

subgraphs/compound-v3/src/sdk/manager.ts
jaimehgb: Interesting ... Sounds like we should fix that then :P 
Maybe by only counting the users when increasing account counters? When increasing `Account.depositCount` (and the rest of counters), do that from the AccountManager, and if all counters are at `0` then you can increase the counter of unique total users.

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
```diff
@@ -0,0 +1,1210 @@
...
+  const supplySideRevenueDeltaUSD = totalRevenueDeltaUSD.minus(
+    protocolRevenueDeltaUSD
+  );
+  dataManager.updateRevenue(protocolRevenueDeltaUSD, supplySideRevenueDeltaUSD);
+}
+
+//
+//
+// Updates market TVL, borrows, prices
+// @return inputToken
```
jaimehgb, line 1069: You forgot this comment :)

subgraphs/compound-v3/src/sdk/token.ts
jaimehgb: Ah I see. It’s fine to leave them then 🤔 

Only thing is if we need to add additional ones for some protocols. We can either modify the lib to add that new token, or make it so that it is configurable, and allows to pass a list of other tokens. But we can leave it for now and work on that when we get there, if ever.

subgraphs/compound-v3/src/sdk/account.ts
jaimehgb: Ah okay, because we have one position counter per account per market. Got it I think

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
```diff
@@ -0,0 +1,1210 @@
...
+  const liquidator = event.params.absorber;
+  const borrower = event.params.borrower;
+  const baseAsset = cometContract.baseToken();
+  const amount = event.params.collateralAbsorbed;
+  const amountUSD = bigIntToBigDecimal(amount, COMPOUND_DECIMALS);
+  const liquidationPenalty =
+    marketEntity.liquidationPenalty.div(BIGDECIMAL_HUNDRED);
+  const profitUSD = amountUSD.times(liquidationPenalty);
+  updateMarketData(market);
+  updateRevenue(market, event.address);
```
tnkrxyz, line 786: Not familar with Compound v3, so forgive my stupid questions here: Is `event.params.collateralAbsorbed` in USD with COMPOUND_DECIMALS? Does liquidation profit (liquidationPenalty) only go to the liquidator?

tnkrxyz (commented):
Nice work! Hope to try out the sdk myself soon.

Made a bunch of comments on the schema 3.0.0. Let me know if you want them moved to the lending schema PR.

dmelotik:
> Let me know if you want them moved to the lending schema PR.

@tnkrxyz thanks for the review, in the future I would say yeah keep them separate (maybe not clear at first), but on this one I will parse out and make changes on the proper PRs

subgraphs/compound-v3/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+`Deposits`
+
+`Withdrawals`
+
+`Borrows`
+
+`Liquidations`
+
+`Repayments`
```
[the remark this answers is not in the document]
dmelotik: I mean, all of our other documentation says `Repayments`. Is this bc the schema says `Repay`?

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  " Collateralized Debt Position (CDP) protocols have singular isolated positions created by users. We aggregate them to give a single view of a market "
+  CDP
+
+  " Pooled protocols pool all users assets into a single market "
+  POOLED
+}
+
+enum PermissionType {
+  " Only users that have been whitelisted can interact. e.g. Only approved institutions can borrow "
+  WHITELIST_ONLY
```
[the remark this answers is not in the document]
dmelotik: In the `LendingProtocol` entity there are fields for `lenderPermissionType` and `borrowerPermissionType`. Does that cover what you are thinking?

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+
+enum FeeType {
+  " Fees from liquidations "
+  LIQUIDATION_FEE
+
+  " Fees given to an admin "
+  ADMIN_FEE
+
+  " Fees that are taken by the protocol "
+  PROTOCOL_FEE
```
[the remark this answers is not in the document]
dmelotik: I would say "INTEREST_FEE" is a product of the protocol (apart of interest rates). But there could be a fee that takes a cut of the interest, I think that is more what this is intended for.

I am adding WITHDRAW FEE

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  id: ID!
+
+  " Fee in percentage. E.g. 5.21% should be stored as 5.21 "
+  rate: BigDecimal!
+
+  " The type of fee (e.g. liquidation, admin, etc.) "
+  type: FeeType!
+}
+
+enum RevenueSource {
```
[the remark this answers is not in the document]
dmelotik: I see what you are saying, maybe RevenueSource should be emitted and replaced with FeeType. They are basically the same.

if there is a fee there is likely revenue generated from that. And we can link fees to RevenueDetails

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  CURVE
+  ## Can add more
+}
+
+# Most lending protocols get token prices through onchain oracles.
+# As oracles are a source of truth it is important to understand the details.
+# This entity will help track oracle-related data in a market
+type Oracle @entity {
+  " { Market Address }-{ Token Address } "
+  id: Bytes!
```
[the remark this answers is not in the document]
dmelotik: same as above, had no problems in comp v3 as well

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+
+  " The specific lending protocol type "
+  lendingType: LendingType
+
+  " The specific permissions required to lend in this protocol "
+  lenderPermissionType: PermissionType
+
+  " The specific permissions required to borrow from this protocol "
+  borrowerPermissionType: PermissionType
+
```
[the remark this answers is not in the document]
dmelotik: what is the use case here?

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  " Risk type of the lending protocol "
+  riskType: RiskType
+
+  " The way a positions can be collateralized "
+  collateralizationType: CollateralizationType
+
+  " Tokens that can be minted. Only applies to CDP "
```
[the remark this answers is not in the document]
dmelotik: Stable coins, aren't historically stable 😅 (joking)..

I added `usually stable coins`

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  " Total number of deposits "
+  depositCount: Int!
+
+  " Total number of withdrawals "
+  withdrawalCount: Int!
+
+  " Total number of borrows "
+  borrowCount: Int!
+
+  " Total number of repayments "
```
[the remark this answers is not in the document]
dmelotik: I mean they are synonymous

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  _markets: [Bytes!]!
+}
+
+###############################
+##### Protocol Timeseries #####
+###############################
+
+type UsageMetricsDailySnapshot @entity {
+  " ID is # of days since Unix epoch time "
+  id: Bytes!
```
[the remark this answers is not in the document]
dmelotik: https://medium.com/edge-node-engineering/two-simple-subgraph-performance-improvements-a76c6b3e7eac

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  " Block number of this snapshot "
+  blockNumber: BigInt!
+
+  " Timestamp of this snapshot "
+  timestamp: BigInt!
+}
+
+type UsageMetricsHourlySnapshot @entity {
+  " { # of hours since Unix epoch time } "
+  id: Bytes!
```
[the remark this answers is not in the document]
dmelotik: see above

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  createdTimestamp: BigInt!
+
+  " Creation block number "
+  createdBlockNumber: BigInt!
+
+  " Details about the price oracle used to get this token's price "
+  oracle: Oracle
+
+  " A unique identifier that can relate multiple markets together. e.g. a common address that they all share. This is useful for markets with multiple input tokens "
+  relation: Bytes
```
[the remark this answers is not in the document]
dmelotik: it can be anything really. The idea is that each market that is related will have the same value for field `relation`

subgraphs/compound-v3/README.md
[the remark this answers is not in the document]
tnkrxyz: I thought those are names of the entities/tables?

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  " Timestamp of this snapshot "
+  timestamp: BigInt!
+
+  " A unique identifier that can relate multiple markets together. e.g. a common address that they all share. This is useful for markets with multiple input tokens "
+  relation: Bytes
+
+  ##### Incentives #####
+
+  " Additional tokens that are given as reward for position in a protocol, usually in liquidity mining programs. e.g. SUSHI in the Onsen program, MATIC for Aave Polygon "
+  rewardTokens: [RewardToken!]
```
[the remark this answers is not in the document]
dmelotik: it could change (unlikely, but could). See rari fuse and a lot of reward contracts have an event for a new reward token

subgraphs/compound-v3/schema.graphql
```diff
@@ -0,0 +1,2020 @@
...
+  liquidates: [Liquidate!]! @derivedFrom(field: "liquidator")
+
+  " Number of times this account has been liquidated "
+  liquidationCount: Int!
+
+  " All liquidation events where this account got liquidated "
+  liquidations: [Liquidate!]! @derivedFrom(field: "liquidatee")
+
+  " Number of times this account has transferred "
+  transferredCount: Int!
```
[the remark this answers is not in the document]
dmelotik: I made it different bc I think there is a difference. `transferredCount` refers to a user making transfers. While `transferCount` just refers to the number of transfer events emitted

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
dmelotik: see the changes I made here: https://github.com/messari/subgraphs/pull/1265/commits/9dbdd9bb4c2c9158946aed077c979e42b95dc10d

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
tnkrxyz: It is different - e.g. creation a new pool in compound [requires admin](https://github.com/compound-finance/compound-protocol/blob/a3214f67b73310d547e00fc578e8355911c9d376/contracts/Comptroller.sol#L933), while theoretically [anyone can create a new pool in euler](https://github.com/compound-finance/compound-protocol/blob/a3214f67b73310d547e00fc578e8355911c9d376/contracts/Comptroller.sol#L933). Maybe this is a type of permission we don't care about, then we can ignore this.

Protocols like goldfinch requires borrowers to be KYC'ed AND added to their whitelist (GOListed).

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
tnkrxyz: See my comment above: https://github.com/messari/subgraphs/pull/1357#discussion_r1046472975, maybe this is something we don't care about.

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: Yeah, I think 0 would be fine. I like the ideal of having a price interface that we can use in all subgraphs and then implement the class specific to each protocol

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
tnkrxyz: Since we are contacting two bytes, I blieve we are not inserting a "-" in between, right?

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
tnkrxyz: OK, makes sense then.

subgraphs/compound-v3/src/sdk/manager.ts
dmelotik: Fixed

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: tbh @jaimehgb it doesn't necessarily make sense in this case. For example, I get prices bc I have a market loaded. I have no need to get prices for any arbitrary token.

subgraphs/compound-v3/protocols/compound-v3/README.md
```diff
@@ -0,0 +1,92 @@
+# Compound III Subgraph
```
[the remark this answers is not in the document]
dmelotik: good catch, no

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
dmelotik: To me whitelist implies KYC, but I guess that may not always be the case. We say in the comment: `Only approved institutions can borrow`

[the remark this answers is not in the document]
dmelotik: As far as poolcreatorpermissions, I am not quite sure where I stand on necessity, I can add it though

subgraphs/compound-v3/src/sdk/position.ts
```diff
@@ -0,0 +1,298 @@
...
+      this.account.openPositionCount -= INT_ONE;
+      this.account.closedPositionCount += INT_ONE;
+      this.account.save();
+
+      //
+      // update market position
+      //
+      this.market.openPositionCount -= INT_ONE;
+      this.market.closedPositionCount += INT_ONE;
+      this.market.save();
```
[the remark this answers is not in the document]
dmelotik: I guess it depends if we want to make this a current lending position count or a cumulative. Our precedent was cumulative, but I think the verbiage is a little ambiguous.

@jaimehgb any third opinion

subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts
dmelotik: Good catch, there is actually a param called `usdValue` that is 8 decimals offset

subgraphs/compound-v3/src/sdk/position.ts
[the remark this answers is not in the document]
jaimehgb: Yeah, I would keep it a cumulative, to keep it consistent with how it was before. Unless we want to change it for some particular reason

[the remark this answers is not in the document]
dmelotik: No, I can just make the comment more descriptive

subgraphs/compound-v3/src/sdk/account.ts
jaimehgb: I’m not sure I understand. But I think it still applies. Maybe one particular instance of `AccountManager` will be dealing with a single `Market` at any given time, but the code itself will deal with multiple depending on what market is instantiated. 

If it will always be just 1 token price per instance you could pass the price of the token at the time of the instantiation instead of a `Pricer` interface though.

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
jaimehgb: _We can move this discussion to a separate issue to not block the PR if you guys prefer._

Isn’t it the other way around? KYC implies whitelist, but whitelist doesn’t imply KYC. If some protocol decided to add a token or not based on a governance vote, it would be whitelist but not KYC.

About adding the field for `PoolCreatorPermissions`, I would do it. It feels strange to have it for general users but not for pool creators.

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
jaimehgb: The way I’ve been doing it for bridges was plain concatenation, without the `-`. When it renders as hex, the `-` will not show as `-`, so can be omitted IMO.

Only thing is we need to be careful with certain IDs. If it is concatenation of addresses, it’s fine. But if we concatenate say 2 arbitrary numbers we might end up with duplicate IDs:

For `{some num}{some other num}`:
- `11.concat(1)`
- `1.concat(11)`

It might be safer to have the `-` actually 🤔 
But then I think that `Bytes.fromUTF8(address-address) != Bytes.empty().concat(address).concat(“-“).concat(address)`. By not having the UTF8 char, it’s easier for people not to accidentaly generate IDs in different ways.

I think I’d leave the `-` out.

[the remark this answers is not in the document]
dmelotik: fixed schema comments

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
dmelotik: added

subgraphs/compound-v3/schema.graphql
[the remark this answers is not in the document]
dmelotik: added

jaimehgb (approved):

subgraphs/compound-v3/src/sdk/account.ts
dmelotik: Plan: see how it affects other subgraphs and iterate from there

PR #1368, opened 2022-11-04
format; maple finance; format & lint

dmelotik:
format and lint maple


jaimehgb (approved):

subgraphs/maple-finance/schema.graphql
```diff
@@ -322,9 +322,6 @@ type LendingProtocol implements Protocol @entity {
 
     " Treasury fee stored in decimals (e.g. 0.01 for 1%). This is the establishment fee to the treasury."
     _treasuryFee: BigDecimal!
-
-    " List of market IDs "
-    _marketIDs: [String!]!
```
jaimehgb: This is not necessary anymore I assume

dmelotik: that is not in the scope of this change, I added it in the next

PR #1369, opened 2022-11-04
fix(#patch); maple; fix protocol tvl

dmelotik:
test: https://okgraph.xyz/?q=dmelotik%2Fmaple-finance-ethereum

Added this to properly update the protocol tvl. Before we were seeing issue like this:
Screen Shot 2022-11-07 at 11 23 10 AM

### DIFF
--- a/subgraphs/maple-finance/src/common/mappingHelpers/update/intervalUpdate.ts
+++ b/subgraphs/maple-finance/src/common/mappingHelpers/update/intervalUpdate.ts
@@ -131,19 +131,28 @@ function intervalUpdateMarket(event: ethereum.Event, market: Market): Market {
 
 function intervalUpdateProtocol(event: ethereum.Event, marketBefore: Market, marketAfter: Market): void {
     const protocol = getOrCreateProtocol();
+    const marketIDs = protocol._marketIDs;
 
-    protocol.totalValueLockedUSD = protocol.totalValueLockedUSD.plus(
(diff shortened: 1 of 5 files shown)


subgraphs/maple-finance/src/common/mappingHelpers/update/intervalUpdate.ts
```diff
@@ -131,19 +131,28 @@ function intervalUpdateMarket(event: ethereum.Event, market: Market): Market {
...
 
-    protocol.totalDepositBalanceUSD = protocol.totalDepositBalanceUSD.plus(
-        marketAfter.totalDepositBalanceUSD.minus(marketBefore.totalDepositBalanceUSD)
-    );
+    for (let i = 0; i < marketIDs.length; i++) {
+        const market = Market.load(marketIDs[i]);
+        if (!market) {
+            // fail safe in case market does not exist for some reason
+            continue;
+        }
```
jaimehgb, line 145: I always have this internal debate on what we should do in these cases. This should not happen at all, unless there is a bug when creating the market, or some other function adds stuff to `marketIDs` for some reason.

Do we want it to fail silently? I would say to at least log an error (stopping the execution feels too extreme), but since our logs are lost in space it, it's not like it will make a difference 😅

dmelotik: log added

Issue #932, opened 2022-09-02
Banker Joe Missing Rewards


dmelotik:
Rewards are missing from our Banker Joe subgraph.

Currently, no rewards are accrued on banker joe, however in the past it seems like rewards have been accrued based on events from the `rewardDistributor`.

See the contract here: https://snowtrace.io/address/0x98a5d8b57d2cd80ffe961a275fb4d4f2aa1aa3c2

PR #1400, opened 2022-11-11
fix(#patch); banker joe; fix rewards

dmelotik:
Testing: https://okgraph.xyz/?q=dmelotik%2Fbanker-joe-avalanche not really working, but don't necessarily think it is worth the effort given there are no protocol rewards anymore

Note: no more rewards being emitted, so this is historical \(>_<)/ #932


dmelotik:
Not doing since banker joe is not emitting rewards anymore and it is not straight forward

Issue #1273, opened 2022-10-26
Aurigami Rewards Missing


dmelotik:
Aurigami missing rewards.

Investigation:
- Trying to call this function: https://aurorascan.dev/address/0xa200b567579a577f582d292f7a1b5c4ecce195f8#readContract#F42
- rewardSpeeds(`0`, `0x4f0d864b1ABf4B701799a0b30b57A22dFEB5917b`, `true`)
    - This should grab the supply side rewards for the USDC market in PLY, but the return is 0
- In addition no events have been emitted from [Comptroller](https://aurorascan.dev/address/0xa200b567579a577f582d292f7a1b5c4ecce195f8) showing that rewards have been claimed (`DistributedSupplierReward ()`)

bye43:
PR for this: https://github.com/messari/subgraphs/pull/1401

PR #1401, opened 2022-11-11
fix(#patch); aurigami; add rewards

dmelotik:
testing: https://okgraph.xyz/?q=dmelotik%2Faurigami-aurora

#1273


subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -333,6 +350,184 @@ function getOrCreateProtocol(): LendingProtocol {
...
+    ]);
+    return;
+  }
+
+  const rewardTokens: string[] = [];
+  const rewardsAmount: BigInt[] = [];
+  const rewardsAmountUSD: BigDecimal[] = [];
+
+  // PLY Borrow first since Reward tokens will be alphabetized
+  // Ply Borrow, Ply Supply, AURORA Borrow, AURORA Supply
```
jaimehgb, line 379: Shouldn’t Aurora go before PLY then?

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
```diff
@@ -333,6 +350,184 @@ function getOrCreateProtocol(): LendingProtocol {
...
+    const priceUSD = getPrice(AURORA_MARKET, protocol._priceOracle)
+      .value.toBigDecimal()
+      .div(mantissaFactorBD);
+    rewardsAmount.push(rewards.rewardsPerDayBI);
+    rewardsAmountUSD.push(rewards.rewardsPerDayBD.times(priceUSD));
+    log.warning("price: {}, token: {}", [
+      priceUSD.toString(),
+      rewards.token.name,
+    ]);
+  }
```
jaimehgb, line 476: If you want to reduce duplication you can do something like:

```
function calculateMarketRewards(token: Address, speed: BigInt, oracle: Address) {
  if (!speed.gt(BIGINT_ONE)) { return; }
  // get or create reward token
  // get rewards per day
  // get price use
  // rewardTokens.push(), rewardAmount.push() ….
  // etc
}
```

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
dmelotik: based off of the token address not the name

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
dmelotik: Would I create a class to pass the arrays in and out of the function? THe assemblyscript kinda makes this tricky

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
jaimehgb: Hmm, but the `rewardToken.id` includes the reward type. So then it should be PLY Borrow, Aurora Borrow, PLY Supply, Aurora Supply?

subgraphs/compound-forks/protocols/aurigami/src/mapping.ts
dmelotik: done

jaimehgb (approved):

ishraq8:
Double checked reward, they are close to what is on their mainsite

PR #1406, opened 2022-11-12
Venus rewards

jaimehgb:
This PR has 2 purposes (they are split on separate commits):
- Fixes an issue with `ActionPaused` events:
  - I had to change the Comptroller ABI to be able to use part of Venus specific functionality, and the project stopped building. It was because the ActionPaused event name differs in Venus from the original Compound contract.
- Adds XVS rewards.
  - Each market is given a `reward speed`, that every time it changes emits an event. This determines the `market.rewardTokenEmissionsAmount` values.
  - Every time that event or `AccrueInterest` are emitted, we'll recalculate the USD value of that `rewardTokenAmount`.
  - To translate `reward speed` to yearly rewards I'm using the code in `rewards.ts` which was copied as is [from here](https://github.com/messari/subgraphs/blob/master/subgraphs/compound-forks/protocols/compound-v2/src/rewards.ts). Let me know if that is accurate or there is a more recent method (@dmelotik )

Syncing here: https://okgraph.xyz/?q=QmUC7vLhpuwP82tAdgp6v9GTdYX5DCaG3dSuQDouVgcP7u

### DIFF
--- a/subgraphs/compound-forks/protocols/venus/src/mapping.ts
+++ b/subgraphs/compound-forks/protocols/venus/src/mapping.ts
@@ -1,4 +1,10 @@
-import { Address, BigInt, ethereum, log } from "@graphprotocol/graph-ts";
+import {
+  Address,
+  BigInt,
+  BigDecimal,
+  ethereum,
+  log,
+} from "@graphprotocol/graph-ts";
 // import from the generated at root in order to reuse methods from root
 import {
   NewPriceOracle,
@@ -8,6 +14,7 @@ import {
   ActionPausedMarket,
   MarketEntered,
   MarketExited,
+  VenusSpeedUpdated,
 } from "../../../generated/Comptroller/Comptroller";
 import {
   Mint,
@@ -19,12 +26,21 @@ import {
   NewReserveFactor,
   Transfer,
 } from "../../../generated/templates/CToken/CToken";
-import { LendingProtocol, Token } from "../../../generated/schema";
+import {
+  LendingProtocol,
+  Token,
+  RewardToken,
+  Market,
+} from "../../../generated/schema";
 import {
   cTokenDecimals,
   Network,
   BIGINT_ZERO,
   BSC_BLOCKS_PER_YEAR,
+  DAYS_PER_YEAR,
+  RewardTokenType,
+  BIGINT_TEN,
+  BIGDECIMAL_ZERO,
 } from "../../../src/constants";
 import {
   ProtocolData,
@@ -47,6 +63,8 @@ import {
   _handleActionPaused,
   _handleMarketEntered,
   _handleTransfer,
+  getOrCreateMarketDailySnapshot,
+  getOrCreateMarketHourlySnapshot,
 } from "../../../src/mapping";
 // otherwise import from the specific subgraph root
 import { CToken } from "../../../generated/Comptroller/CToken";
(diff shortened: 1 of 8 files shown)


subgraphs/compound-forks/protocols/venus/src/mapping.ts
```diff
@@ -293,6 +342,103 @@ export function handleTransfer(event: Transfer): void {
...
+// update market reward amounts based on event data.
+export function handleVenusSpeedUpdated(event: VenusSpeedUpdated): void {
+  const marketAddress = event.params.vToken;
+  const speed = event.params.newSpeed;
+  const rewards = getRewardsPerDay(
+    event.block.timestamp,
+    event.block.number,
+    speed.toBigDecimal(),
+    RewardIntervalType.BLOCK
+  ).times(BigDecimal.fromString(DAYS_PER_YEAR.toString()));
```
dmelotik, line 354: Rewards should be recorded on a per day basis, that is likely why they are so high on subgraphs.xyz

subgraphs/compound-forks/protocols/venus/src/mapping.ts
```diff
@@ -293,6 +342,103 @@ export function handleTransfer(event: Transfer): void {
...
+    return;
+  }
+
+  const protocol = getOrCreateProtocol();
+  const oracleContract = PriceOracle.bind(
+    Address.fromString(protocol._priceOracle)
+  );
+  const priceCall = oracleContract.try_getUnderlyingPrice(vXVS.address);
+  if (priceCall.reverted) {
+    log.error("unable to calculate rewards: priceCall reverted for vXVS {}", [
```
dmelotik, line 397: Help me understand why an error makes more sense here

subgraphs/compound-forks/protocols/venus/src/mapping.ts
```diff
@@ -293,6 +342,103 @@ export function handleTransfer(event: Transfer): void {
...
+      vXVS.address.toHexString(),
+    ]);
+    return;
+  }
+
+  const rewardAmount = market.rewardTokenEmissionsAmount![0];
+  const rewardAmountUSD = rewardAmount
+    .times(priceCall.value)
+    .divDecimal(BIGINT_TEN.pow(ORACLE_PRECISION as u8).toBigDecimal())
```
dmelotik, line 406: you could also use `exponentToBigDecimal()` instead of the pow operator here

dmelotik: 🤷‍♂️

dmelotik (approved):
lgtm (aside from the yearly reward part)

Note: there is a newer version of reward.ts in _reference_ but it shouldn't affect the functionality

subgraphs/compound-forks/protocols/venus/src/mapping.ts
jaimehgb: Damn, not sure why I put it to a year. I was convinced we normalized them to a year 😅

subgraphs/compound-forks/protocols/venus/src/mapping.ts
jaimehgb: I tend to avoid warnings, because of the unclear meaning from being in between an error and an info. If there is some action that should be taken I default to error. If not, info.

This call should not revert I think? If it does, we should maybe get the price from a different source for the blocks where it fails? So an error would inform us about that (if we could have alerts or something).

https://dave.cheney.net/2015/11/05/lets-talk-about-logging

dmelotik: makes sense, I think there are differing opinions on logging in software eng in general. But I will think about this moving forward. I usually default to warnings, and don't log anything else

jaimehgb: >  I usually default to warnings, and don't log anything else

Damn, haha 😅 

The problem there is basically that. It is not clear what we should do about it once a warning pops up somewhere. Is it safe to ignore? Or should we fix it somehow.

PR #1417, opened 2022-11-15
[GIN] Fix reflexer governance proposal with empty values and signatures

danielkhoo:
## Description

Reflexer governance proposal are created undefined values for `signatures` and `values`. This causes issues further down the pipeline as the target, calldata and signature lengths don't match. This PR adds a simple workaround to replace the undefined with empty values.

image


mmnavarr (approved):
-_-

PR #1418, opened 2022-11-15
fix(#minor); Uniswap-v3; Fix TVL spike down

steegecs:
**Context:**
The TVL spiked downward, in order to fix this I have blacklisted the culprit pool.


dmelotik (approved):
LGTM, contingent on adding a small note to the README for future devs to understand

PR #1419, opened 2022-11-16
chore(); clean up protocol README's

dmelotik:
Clean and organize protocol readme's into standard structure.

Skipped any governor subgraphs. And some subgraphs like NFT marketplaces and Network suibgraph it doesn't make sense to follow the structure


dmelotik:
not necessary

PR #1420, opened 2022-11-16
fix(#patch); Uniswap Forks; Updated swap function so that it can account for negative tokenOut balance

steegecs:
**Context**
It can sometimes be the case that there are 2 ingoing tokens in a swap and the outgoing token is less than the amount going into the liquidity pool. This essentially leads to a negative `TokenOut` amount. This PR adds handling so that it will identify these cases and allow the `TokenOut` to be negative on these rare instances.


subgraphs/uniswap-forks/src/common/creators.ts
```diff
@@ -349,37 +365,26 @@ class SwapTokens {
...
-    amountIn = amount0;
-    amountOut = amount1.times(BIGINT_NEG_ONE);
-  } else {
+  if (amount0Out.gt(BIGINT_ZERO)) {
     tokenIn = token1;
     tokenOut = token0;
-    amountIn = amount1;
-    amountOut = amount0.times(BIGINT_NEG_ONE);
+    amountIn = amount1In.minus(amount1Out);
+    amountOut = amount0In.minus(amount0Out).times(BIGINT_NEG_ONE);
```
dmelotik, line 382: amount0 out is like the overflow number? Could make a comment on how you know it's neg

dmelotik (approved):
small thing + a lint. Approving at your discretion. cheers 🎱

PR #1429, opened 2022-11-17
feat (#minor); euler-finance; add reward emission

tnkrxyz:
This PR adds reward emssion for euler-finance.

Test deployment: https://subgraphs.messari.io/subgraph?endpoint=https%3A//api.thegraph.com/subgraphs/name/tnkrxyz/euler-ethereum&tab=pool&poolId=0xe025e3ca2be02316033184551d4d3aa22024d9dc


deployment/deployment.json
```diff
@@ -1976,8 +1976,8 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.2.3",
-          "methodology": "1.1.0"
+          "subgraph": "1.3.3",
```
dmelotik, line 1979: should make it 1.3.0 when bumping like this

deployment/deployment.json
```diff
@@ -1976,8 +1976,8 @@
...
+          "methodology": "1.2.0"
```
dmelotik, line 1980: this is 1.3.0

subgraphs/euler-finance/protocols/euler-finance/config/templates/euler.template.yaml
```diff
@@ -66,3 +66,24 @@ dataSources:
...
+      language: wasm/assemblyscript
+      entities: []
+      abis:
+        - name: EulStakes
+          file: ./abis/EulStakes.json
+      eventHandlers:
+        - event: Stake(indexed address,indexed address,address,uint256)
+          handler: handleEULStake
+          receipt: true
+      file: ./src/mappings/handlers.ts
```
dmelotik, line 89: new line at the end of the file

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +324,110 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
...
+
+      const cutoffAmount = getCutoffValue(marketStakeAmounts, 10);
+      const totalRewardAmount = BigDecimal.fromString((EUL_DIST[prevEpochID - START_EPOCH] * EUL_DECIMALS).toString());
+
+      const EULToken = getOrCreateToken(Address.fromString(EUL_ADDRESS));
+      const rewardToken = getOrCreateRewardToken(Address.fromString(EUL_ADDRESS), RewardTokenType.BORROW);
+      const dailyScaler = BigDecimal.fromString((BLOCKS_PER_DAY / (BLOCKS_PER_EPOCH as f64)).toString());
+      for (let i = 0; i < protocol._marketIDs!.length; i++) {
+        const mktID = protocol._marketIDs![i];
+        const mrkt = getOrCreateMarket(mktID);
```
dmelotik, line 395: Kinda the same idea with what you mentioned on my PR. My take is that it is supposed to be there, so if it is not you would want to throw an error and skip

subgraphs/euler-finance/src/mappings/helpers.ts
```diff
@@ -666,3 +667,29 @@ function getRepayForLiquidation(event: ethereum.Event): BigInt | null {
...
+    const blocksLapsed = endBlock.minus(mrkt._borrowLastUpdateBlock!);
+    const addWeightedTotalBorrowUSD = mrkt.totalBorrowBalanceUSD.times(blocksLapsed.toBigDecimal());
+    mrkt._weightedTotalBorrowUSD = mrkt._weightedTotalBorrowUSD!.plus(addWeightedTotalBorrowUSD);
+    mrkt._borrowLastUpdateBlock = endBlock;
+    mrkt.save();
+  }
+}
+
+export function updateWeightedStakeAmount(market: Market, endBlock: BigInt): void {
+  const blocksLapsed = endBlock.minus(market._stakeLastUpdateBlock!);
```
dmelotik, line 691: does `market._stakeLastUpdateBlock` need to be updated after this function?

dmelotik (commented):
Code looks good! But the rewards seem to be off, cc @ishraq8 to keep this on your radar to QA rewards.

subgraphs/euler-finance/src/mappings/helpers.ts
tnkrxyz: Good point! Currently, it is done after the function call (needed before a refactoring), but it can be moved inside the function.

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: To clarify, are you referring to `mrkt = getOrCreateMarket(mktID)` or `if (!mrktRewardToken || mrktRewardToken.length == 0)`. I think, in my case, `getOrCreateMarket` always returns a valid market, and `rewardToken = getOrCreateRewardToken()` should alwasys returns a valid rewardToken?

tnkrxyz:
> Code looks good! But the rewards seem to be off, cc @ishraq8 to keep this on your radar to QA rewards.

Thanks! I will have to dig into this more to understand how euler calculates the APR. 

@ishraq8 , please let me know what you find when you get to it.

subgraphs/euler-finance/src/mappings/helpers.ts
dmelotik: Would increase readability that way (for me at least)

dmelotik:
@tnkrxyz although, if you find this is the best we can do then we can make a note in the README.md and move on (as long as @pietydespises is OK with that). But given the nature of how we have to calculate this, maybe this is the best we can do?

subgraphs/euler-finance/src/mappings/handlers.ts
dmelotik: `mrkt = getOrCreateMarket(mktID)` is what I am talking about.

And I agree it will always return a Market entity. I guess my point is that if it had to make a Market entity that would be an error. It really shouldn't happen.

I usually do something like this:
```
const mrkt = Market.load(mktID);
if (!mrkt) {
     log.error();
     return;
}
```

Idk it rly isn't a major thing, I'm just being picky and we are really just informally defining standards as we do this

ishraq8:
@dmelotik @tnkrxyz Rewards are accurate now, moving inside the function worked. I recommend pulling this request. Great work from both!

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: OK, makes sense. I am switching to load & check.

tnkrxyz:
> @dmelotik @tnkrxyz Rewards are accurate now, moving inside the function worked. I recommend pulling this request. Great work from both!

@ishraq8 , the changes suggested by Dylan should not affect the reward emission amount.

I am getting some clarification on the distribution algorithm on Euler's discord server. Will report back what I learned.

dmelotik:
Nice! Approving, feel free to merge after you hear back or if you think it is ready

dmelotik (approved):

tnkrxyz:
After getting [some clarification from the euler discord server](https://discord.com/channels/742749441697513632/1021868855557116006/1049334533511725188), I am more comfortable with [the fixed implementation of reward emission](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/euler-ethereum&tab=poolOverview). Since euler just changed their [EUL distribution schema](#1488) and EUL distribution for past epoch is not available on their front end, it is hard to QA the numbers. 

@dmelotik , @ishraq8 I know the PR is approved, but I appreciate it if you can take a look of the last 5 commits since Dylan's review before I merge.

image

ishraq8:
@tnkrxyz From a QA perspective the new schema is fairly close in terms of numbers to the old one. As long as @dmelotik is ok with the changes we can go ahead and pull the additional lines so to avoid coming back in the future.

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +310,132 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
     event,
   );
 }
+
+export function handleStake(event: Stake): void {
```
dmelotik, line 314: tbh, I am kinda struggling to follow all of the moving parts in this function.. Could you also include notes about the function to find rewards that you linked in the PR?

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +310,132 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
...
+  const epochID = getCurrentEpoch(event);
+  if (epochID < 0) {
+    market._stakedAmount = market._stakedAmount.plus(deltaStakedAmount);
+    market.save();
+    return;
+  }
+
+  const epochStartBlock = getStartBlockForEpoch(epochID)!;
+  let epoch = _Epoch.load(epochID.toString());
+  if (!epoch) {
```
dmelotik, line 332: This could be an early return statement as well:

```
if (epoch) {
    log;
    return;
}
```

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +310,132 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
...
+  // keep track of staked amount from epoch 1
+  const epochID = getCurrentEpoch(event);
+  if (epochID < 0) {
+    market._stakedAmount = market._stakedAmount.plus(deltaStakedAmount);
+    market.save();
+    return;
+  }
+
+  const epochStartBlock = getStartBlockForEpoch(epochID)!;
+  let epoch = _Epoch.load(epochID.toString());
```
dmelotik, line 331: each epoch is ~6 minutes, does that mean the emissions are updated every 6 minutes? Or is it usually the same rewards so we are just updating the price?

dmelotik (commented):
Hey, sorry I have some more requests. But the reward function is pretty difficult to digest so I offered some things to improve readability

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +310,132 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
...
+        EULPriceUSD = EULToken.lastPriceUSD!;
+      }
+
+      const rewardToken = getOrCreateRewardToken(Address.fromString(EUL_ADDRESS), RewardTokenType.BORROW);
+      const totalRewardAmount = BigDecimal.fromString((EUL_DIST[prevEpochID - START_EPOCH] * EUL_DECIMALS).toString());
+      // select top 10 staked markets, calculate sqrt(weighted staked amount)
+      const cutoffAmount = getCutoffValue(marketWeightedStakedAmounts, 10);
+      let sumAccumulator = BIGDECIMAL_ZERO;
+      for (let i = 0; i < marketWeightedStakedAmounts.length; i++) {
+        // TODO: reflect changes in eIP 24 and 28
```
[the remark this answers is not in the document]
tnkrxyz: Not yet: https://github.com/messari/subgraphs/issues/1488 . They only made the change a few days ago. My preference is to merge this PR first and come back to implement the new distribution scheme.

subgraphs/euler-finance/src/mappings/handlers.ts
```diff
@@ -293,3 +310,132 @@ export function handleGovSetReserveFee(event: GovSetReserveFee): void {
...
+        if (marketWeightedStakedAmounts[i].ge(cutoffAmount)) {
+          sumAccumulator = sumAccumulator.plus(marketWeightedStakedAmounts[i].sqrt().toBigDecimal());
+        }
+      }
+
+      // scale to daily emission amount
+      const dailyScaler = BigDecimal.fromString((BLOCKS_PER_DAY / (BLOCKS_PER_EPOCH as f64)).toString());
+      for (let i = 0; i < protocol._marketIDs!.length; i++) {
```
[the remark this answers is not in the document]
tnkrxyz: I believe I need all three loops in the function. The first one finalizing the weighted staked amount, the second (this comment) sums up the total weighted staked amount (`sumAccumulator`) among top 10 markets, the third distributes the award according to `sqrt(weighted staked amount) / sqrt(sumAccumulator)` among top 10 markets.

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: If I do this, I will need to repeat [lines 433-440](https://github.com/tnkrxyz/subgraphs/blob/16d7fb2e879f7c9cea658ebb2a8cb3996db1143f/subgraphs/euler-finance/src/mappings/handlers.ts#L433-L440)
```
  if (!market._stakeLastUpdateBlock) {
    market._stakeLastUpdateBlock = epochStartBlock;
    market._weightedStakedAmount = BIGINT_ZERO;
  }
  // update _weightedStakeAmount before updating _stakedAmount
  updateWeightedStakedAmount(market, event.block.number);
  market._stakedAmount = market._stakedAmount.plus(deltaStakedAmount);
  market.save();
```

And these lines have to be after the [new epoch logic (`if (!epoch) {`)](https://github.com/tnkrxyz/subgraphs/blob/16d7fb2e879f7c9cea658ebb2a8cb3996db1143f/subgraphs/euler-finance/src/mappings/handlers.ts#L332)

tnkrxyz:
@dmelotik , what do you think about refactoring some of the code into separate functions in https://github.com/messari/subgraphs/pull/1429/commits/54870bc95c6c68a457f23e3b1ae38e52a0be86c2 ?

subgraphs/euler-finance/src/mappings/handlers.ts
dmelotik: hmmm, we should put that in there. So from my understanding the price only gets updated every 15 days? It could change pretty drastically. We should have logic to update the price at least every 30 mins.

subgraphs/euler-finance/src/mappings/handlers.ts
dmelotik: Fair, I imagine it may change a little as well when we implement the new logic?

subgraphs/euler-finance/src/mappings/handlers.ts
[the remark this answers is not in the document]
dmelotik: Yeah why don't we do that. As long as we cover everything I am okay with doing that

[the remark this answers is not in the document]
dmelotik: Feel free to resolve the last 3 convos and merge unless you have anything

dmelotik:
> what do you think about refactoring some of the code into separate functions in https://github.com/messari/subgraphs/commit/54870bc95c6c68a457f23e3b1ae38e52a0be86c2 ?

That helped!

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: Correct.

subgraphs/euler-finance/src/mappings/handlers.ts
tnkrxyz: Good question. They only ever distribute rewards at the end of each epoch, so the RewardTokenEmissionAmount will not change within an epoch. Do you think it is necessary to update RewardTokenEmissionUSD every 30 minutes? If so, we can update RewardTokenEmissionUSD within `updatePrices()` function when we update prices for all markets (there is a EUL market, so EUL price will be update there).

dmelotik: Yes I think so. In all of the other protocol rewards we calculate the rate based on the current price of EUL. That sounds like the right spot ✅

tnkrxyz: OK, added code to update rewardTokenEmissionsUSD when EUL price is updated: https://github.com/messari/subgraphs/pull/1429/commits/939d3fd1b56e1321ec78aff6a77946208cc62404

tnkrxyz: OK, sync'ed successfully. I will merge.

PR #1438, opened 2022-11-18
feat(): Added new bridge schema

jaimehgb:
## Some considerations:

Bridges subgraphs are a bit tricky because, by nature, these protocols data is scattered among different networks. This is different than other protocols that are also deployed in multiple networks at the same time in the fact that usually these deployments are completely independent from each other. This is not the case with bridges, where transactions start in one chain and finish in another one.

This requires paying extra attention to some metrics, to avoid double counting them. If we deploy a subgraph for the same bridge on the multiple chains where it operates, metrics like fees can be accounted for in all deployments. Both where the transaction originates and where it lands. In this example, it would duplicate the revenue.

At the same time, if we wanted to combine the data among deployments, we'd need to be able to link each entity with its equivalent on other chains. For example, some token address might be `A` on the source chain, and `B` on the destination. Or they might have the same address if deployed with the same deployer under certain circumstances. Same as with tokens, happens with pools. When bridging tokens from some pool, they'll land in its homologous in the other chain, and their addresses might be different. 

This schema has been designed with this in mind.  

## Entity Overview

Quick comment on the entities of this schema.

`Pool`, is the contract where tokens are locked/burned/minted/released in our chain. In the case of liquidity networks and one side of the lock/mint bridges they'll hold some token balance. Pools always have an equivalent contract on the chains the bridge connects with. These equivalents don't need to be stored in a subgraph. With this schema we only care about the Pools that are deployed on the observed network.

Pools connect with their homologous in other chains. This is represented by what's called `PoolRoutes`. This entity holds info about the networks it connects, the token bridged and its equivalent and the volumes sent through it.

`CrosschainToken` is an entity to keep all the information known from the token on the other side. `DAI` address might be some `A` on ethereum, but `B` on fantom. We store `A` as a `Token` entity, but `B` as a `CrosschainToken` one to make sure we know they are not in the current chain. Tokens that don't live in the current chain need to be represented with `CrosschainToken`.

Comments on directions. Event entities, like `BridgeTransfer`, and `BridgeMessage` have the fields `to, from, transferTo and transferFrom`. They are different, although might match sometimes. `to and from` refer to the actual EVM transaction we are handling. They might be an EOA and a contract, or 2 contracts, etc. Whereas `transferTo and transferFrom` refer to the addresses bridging the funds. User `A` might be bridging funds to themselves on the same address `A`. But the transaction will interact with some bridge router contract `X`. In this case, `to and from` will be `A and X`, and `transferTo and transferFrom` will be `A and A`.

More on directions. Volumes are split into volumeIn and volumeOut across different entities. The "in" and "out" refer to the direction in which funds are flowing taking the current chain as reference. VolumeIn means some token was bridged from some chain to the one we are working with. VolumeOut means the opposite: we are bridging funds out. This can be confusing because when bridging out we usually lock funds into the pool. And when bridging in, funds are leaving the pool. Always interpret them having the chain as a reference (not the pool).

## Currently open questions:
- [x] How should we count unique users? 
  - **Problem**: We can count every single address we detect: source and destination of funds when bridging out, and also source and destination of funds when bridging in. Should we count only one side (like we do for revenue) or both?
  - **Options**:
    - Every address that interacts with the protocol, including the ones that bridged funds from other chains to the current.
    - Only addresses that bridge funds out from the current chain observed, and maybe destination if different than source.
- [x] Same as above, but with transactions. In this case counting everything might make sense because they are real transactions happening on the chain (since funds need to be burned, released, locked, minted).
- [x] Even if not counted, should we create an entity for transactions originated in other chains that land here? I think yes, because their data might not be entirely available on the other side.
- [x] I made the `address` field required in `CrosschainToken` entity. I can see how in some bridges this info might just not be available from the contracts. I'm assuming that permissionless bridges should disclose the relationship between token addresses. And the permissioned ones might not. But since permissioned bridges will have less tokens by nature, can we hardcode all supported token relationships? Or better make the address field optional.
- [x] I'm not entirely happy with the Pool entity. In most bridges pools have a single token. But some of them might have multiple tokens, which allow for swaps to happen while bridging. Here there are 2 options:
  - When building a subgraph for a bridge with multiple tokens in the same pool, breaking the pool down into separate ones. For example, consider the pool DAI/USDC/USDT, which contains liquidity for all those assets and allows swapping while bridging. Since liquidity will be denominated in the 3 different assets we could create 3 pools to hold all the data of that one pool. One with each inputToken. What I dislike here is that it doesn't match the actual thing.
  - Creating a new Pool entity dedicated to multi token pools. Problem here will be bloating the schema since it will also need all the snapshots for that entity. Also, they'll need to be treated as different entities downstream, when they'll be really similar.
  - Adding support for multiple tokens in the current pool entity. It would consist on renaming the following fields (in both the pool entity and the snapshots):
    - `inputToken: BigInt` -> `inputTokens: [BigInt]`
    - `inputTokenBalance: BigInt` -> `inputTokenBalances: [BigInt]`
    - `cumulativeVolumeIn: BigInt` -> `cumulativeVolumesIn: [BigInt]`
    - `cumulativeVolumeInUSD: BigDecimal` -> `cumulativeVolumesInUSD: [BigDecimal]`
    - `cumulativeVolumeOut: BigInt` -> `cumulativeVolumesOut: [BigInt]`
    - `cumulativeVolumeOutUSD: BigDecimal` -> `cumulativeVolumesOutUSD: [BigDecimal]`
    - `netVolume: BigInt` -> `netVolumes: [BigInt]`
    - `netVolumeUSD: BigDecimal` -> `netVolumesUSD: [BigDecimal]`
    - The problems with this rename are:
      - Increased complexity since we need to ensure all those arrays match their ordering with that of `inputTokens`
      - Most pools are single token. We'll be losing the ability to query like `query all pools where volumeInUSD > someVal`, because they'll be arrays now.


schema-bridge.graphql
```diff
@@ -100,32 +135,147 @@ type Protocol @entity {
...
 
-  " Current TVL (Total Value Locked) of the entire protocol "
+  " Current TVL (Total Value Locked) of the entire protocol on this chain. "
   totalValueLockedUSD: BigDecimal!
 
   " Current PCV (Protocol Controlled Value). Only relevant for protocols with PCV. "
   protocolControlledValueUSD: BigDecimal
 
-  " Revenue claimed by suppliers to the protocol. LPs on DEXs (e.g. 0.25% of the swap fee in Sushiswap). Depositors on Lending Protocols. NFT sellers on OpenSea. "
+  " Revenue claimed by suppliers to the protocol. LPs on DEXs (e.g. 0.25% of the swap fee in Sushiswap). Depositors on Lending Protocols. NFT sellers on OpenSea. In the case of bridges, revenue only includes that of transactions initiated in this chain ""
```
fortysevenlabs, line 149: extra " at the end

schema-bridge.graphql
```diff
@@ -100,32 +135,147 @@ type Protocol @entity {
...
+
+type BridgeProtocol implements Protocol @entity {
+  " Smart contract address of the protocol's main contract (Factory, Registry, etc) "
+  id: ID!
+
+  " Name of the protocol, including version. e.g. Uniswap v3 "
+  name: String!
+
+  " Slug of protocol, including version. e.g. uniswap-v3 "
+  slug: String!
```
dmelotik, line 188: there should be a field that will be the same in all chains for a specific bridge, this seems like what it would be. Maybe just add that to the comment

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
...
+"""
+An event is any user action that occurs in a protocol. Generally, they are Ethereum events
+emitted by a function in the smart contracts, stored in transaction receipts as event logs.
+However, some user actions of interest are function calls that don't emit events. For example,
+the deposit and withdraw functions in Yearn do not emit any events. In our subgraphs, we still
+store them as events, although they are not technically Ethereum events emitted by smart
+contracts.
+"""
+interface Event {
+  " { Event type }-{ Transaction hash }-{ Log index } "
```
dmelotik, line 812: There is no enum called `EventType`, should probably have that or change the id.

Another question is, why have this identifier instead of unique entities for each event type. That way is the precedent and seems to be more clear on the ingestion side.

dmelotik: There is also this: https://medium.com/edge-node-engineering/two-simple-subgraph-performance-improvements-a76c6b3e7eac

I think we should use this going forward in any new schemas as it seems to be better practice. Immutable entities and `Bytes` as the id type

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
...
+  to: String!
+
+  " Address that sent the tokens "
+  from: String!
+
+  " Are we briding in or out? "
+  isOutgoing: Boolean!
+
+  " This is the address that's receiving the bridged funds (not a pool contract). When bridging out, it will be the address on the destination chain. "
+  transferTo: String!
```
dmelotik, line 937: We can make this `Bytes` assuming we adopt that in this schema

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
   rewardTokenEmissionsUSD: [BigDecimal!]
 }
 
+type PoolRouteSnapshot @entity {
```
dmelotik, line 761: what is the time interval on this snapshot?

schema-bridge.graphql
```diff
@@ -55,6 +66,30 @@ type Token @entity {
...
 
+# CrosschainToken is the representation of a token from a different chain.
+# For example, if you bridge USDC from Ethereum to Polygon, the token on Polygon is a CrosschainToken.
+# CrosschainToken might or might not be canonical. It refers to the counterpart of the token on the other chain.
+# If the token is wrapped on the current chain, CrosschainToken will be canonical on the opposite chain; and viceversa.
+type CrosschainToken @entity {
+  # We prepend the chainID because it is possible for a token to have the same address on multiple chains.
+  # ChainID is the id of the destination chain (not the one the subgraph is running on), and tokenID is the 
+  # address of the token on the current chain.
+  " {chainID}-{tokenID} "
```
dmelotik, line 77: why use a seperate identifier for chains when we have the `network` enum already?

schema-bridge.graphql
```diff
@@ -100,32 +135,147 @@ type Protocol @entity {
...
+  methodologyVersion: String!
+
+  " The blockchain network this subgraph is indexing on "
+  network: Network!
+
+  " The type of protocol (e.g. DEX, Lending, Yield, Bridge, etc) "
+  type: ProtocolType!
+
+  " Indicates if the bridge is permissionless: i.e. anyone can add tokens to it "
+  permissionless: Boolean!
```
dmelotik, line 206: Is there more nuance to permission type? Like KYC, permissionless, whitelist? not sure, something we have in the lending subgraph.

dmelotik: Also, I guess if you are going to remain consistent this should be `isPermissioned`

schema-bridge.graphql
```diff
@@ -310,6 +561,45 @@ type Pool @entity {
   hourlySnapshots: [PoolHourlySnapshot!]! @derivedFrom(field: "pool")
 }
 
+type PoolRoute @entity {
+  # Still thinking about how to make this ID in a way that allows routes to be linked among chains
```
dmelotik, line 565: could it be linked by inputToken and crossToken? chain `A` crossToken.token == chain `B` inputToken and vice versa

schema-bridge.graphql
```diff
@@ -310,6 +561,45 @@ type Pool @entity {
...
+type PoolRoute @entity {
+  # Still thinking about how to make this ID in a way that allows routes to be linked among chains 
+  " {Pool ID}-{lowerChainID}-{greaterChainID} "
+  id: ID!
+
+  " Pool this route belongs to "
+  pool: Pool!
+
+  " Type of the pool on opposite chain "
+  counterType: BridgePoolType
```
dmelotik, line 573: This seems mistyped, I am not sure what it is

dmelotik:
Some initial thoughts I am having while reading the description:

> How should we count unique users?

I think it is worth keeping track of unique users on each chain. Then for total bridge unique users you can take the cardinality of the set of unique addresses on chain `A` (appears on chain `A` but not `B`), chain `B`, and the same addresses. This could be computed post-subgraph as long as we have an `ActiveAccount` entity.

> Pools with multiple tokens...

Same issue with lending subgraphs, maybe you could take the same approach we came up with and keep them separate, but have a relating field to show a relationship.

schema-bridge.graphql
jaimehgb: This makes sense. It should be this same one right? Since it’s how we do it on all other protocols. I’ll add it to the comment

schema-bridge.graphql
jaimehgb: Sounds good. Will add that too.
How do bytes show when querying? I’m thinking, are we relying on the value of the ID for something anywhere? For example, liquidity pools in DEXes have their contract address as the ID. When querying, do they render as a string?

`Event type` refers to if it is a transfer, deposit, withdrawal, message. Will clarify it 👍

schema-bridge.graphql
jaimehgb: Depends on its parent. If it is inside `PoolHourlySnapshot.routes` it will be hourly. Same with the daily ones.

I made it this way because since they are a sub-entity of the PoolSnapshot they won’t be queried directly, they will always be retrieved as part of the pool snapshot.

schema-bridge.graphql
jaimehgb: @pietydespises I defer to you here

schema-bridge.graphql
jaimehgb: That would be ideal. I fear that we might not be able to always know the address of the token on the other chains. 

Although I’m assumig that we are by making `CrosschainToken.token` a non-nullable field 😅

schema-bridge.graphql
jaimehgb: Lol, wasn’t sure how to name it. When you send tokens from chain `A` to chain `B`, they are sent from a pool in `A` and received in a pool in `B`. Each of those pools has a type, they go in pairs:
- Lock/Release <> Mint/Burn
- Liquidity <> Liquidity
(not sure if there are more combinations)

`counterType` refers to the type of the pool on the other side of the route.

schema-bridge.graphql
```diff
@@ -1,5 +1,5 @@
 # Subgraph Schema: Bridge
-# Version: 1.3.0
+# Version: 1.0.0
```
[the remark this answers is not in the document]
jaimehgb: Looking at the `deployment.json` doesn’t seem like it. 2 bridges are being built using the original, but the original schema was a copy of the `generic` one so we can label as `generic` if they are not updated to support this one

jaimehgb:
> I think it is worth keeping track of unique users on each chain. Then for total bridge unique users you can take the cardinality of the set of unique addresses on chain `A` (appears on chain `A` but not `B`), chain `B`, and the same addresses. This could be computed post-subgraph as long as we have an `ActiveAccount` entity.

Yeah, discussing it yesteday with @pietydespises we thought it was better to only count users on each chain interacting directly with that chain: sending out or receiving funds via the bridge on the particular chain. Could also split it into users sending and users receiving counters. Then, to get the bridge count of unique users there is no other way around it than combining all the `Account` entities and finding uniques amond all deployments.

Same with transactions, although here it’s easier. Split into in and out counters. Then total transaction count would be the sum of all transactions out for all deployments (which should match transactions in).


> Same issue with lending subgraphs, maybe you could take the same approach we came up with and keep them separate, but have a relating field to show a relationship.

Do you have an example of this? to take a look

schema-bridge.graphql
jaimehgb: I was thinking because of simplicity (bridge contracts already use this chainID to identify the different chains), and to avoid having to map our `network` enum to chainIDs.

It is not a new ID I’m inventing, but these: https://chainlist.org/

schema-bridge.graphql
dmelotik: yeah I would think so, just wanted to throw that out since we will probably want on ingestion side

schema-bridge.graphql
dmelotik: single markets show up as the address, it gets a little funky and unreadable when you concat things onto it.

> are we relying on the value of the ID for something anywhere?

What I did to mediate this is add the days or hours as a field to the snapshot

schema-bridge.graphql
dmelotik: `receiverPoolType`? But that makes sense, so maybe `counterType` is sufficient

dmelotik:
> Do you have an example of this? to take a look

https://github.com/messari/subgraphs/pull/1265/files#diff-ac4294b9bbcc88f965ed2359faa615e92707aafea95a1e344d5b46c8722f736eR791

schema-bridge.graphql
pietydespises: I think permissionless, whitelist, private is probably best

schema-bridge.graphql
pietydespises: that should be all the combinations! I hope

schema-bridge.graphql
```diff
@@ -154,18 +304,54 @@ type UsageMetricsDailySnapshot @entity {
...
+  dailyMessageSentCount: Int!
+
   " Total number of pools "
   totalPoolCount: Int!
 
+  " We call a pool route each connection a token has with another chain. For example, if USDC can be bridged from the current chain to 2 others, it will have 2 pool routes. Not all assets can be bridged to all chains. "
+  totalPoolRouteCount: Int!
+
+  "  Number of routes where the destination asset is canonical (i.e. it is not wrapped)"
+  canonicalRouteCount: Int!
```
dhruv-chauhan, line 347: I was confused if this would be total or daily, suggest renaming to `totalCanonicalRouteCount` (or `dailyCanonicalRouteCount`?)

same with `wrappedRouteCount` and `supportedTokenCount`

schema-bridge.graphql
```diff
@@ -55,6 +66,30 @@ type Token @entity {
   lastPriceBlockNumber: BigInt
 }
 
+# CrosschainToken is the representation of a token from a different chain.
+# For example, if you bridge USDC from Ethereum to Polygon, the token on Polygon is a CrosschainToken.
+# CrosschainToken might or might not be canonical. It refers to the counterpart of the token on the other chain.
+# If the token is wrapped on the current chain, CrosschainToken will be canonical on the opposite chain; and viceversa.
```
dhruv-chauhan, line 72: Is this always true? If pools are liquidity <> liquidity, could the tokens on each side not be wrapped?

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
...
+  from: String!
+
+  " Block number of this event "
+  blockNumber: BigInt!
+
+  " Timestamp of this event "
+  timestamp: BigInt!
+}
+
+type LiquidityDeposit implements Event @entity {
```
dhruv-chauhan, line 837: I think `LiquidityDeposit`, `LiquidityWithdraw`, `BridgeTransfer` events should also have an `amount` field

schema-bridge.graphql
```diff
@@ -100,32 +135,147 @@ type Protocol @entity {
...
+  " Daily financial metrics for this protocol "
+  financialMetrics: [FinancialsDailySnapshot!]! @derivedFrom(field: "protocol")
+
+  ##### Pools #####
+
+  " All pools that belong to this protocol "
+  pools: [Pool!]! @derivedFrom(field: "protocol")
+}
+
+type BridgeProtocol implements Protocol @entity {
```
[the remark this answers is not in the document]
jaimehgb: Whoops, that is correct 👍

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
...
+"""
+interface Event {
+  " { Event type }-{ Transaction hash }-{ Log index } "
+  id: ID!
+
+  " Transaction hash of the transaction that emitted this event "
+  hash: String!
+
+  " Event log index. For transactions that don't emit event, create arbitrary index starting from 0 "
+  logIndex: Int!
```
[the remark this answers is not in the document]
jaimehgb: We seem to be using `Int` in all the other schemas. Any reason as to why changing it?

schema-bridge.graphql
```diff
@@ -426,11 +758,271 @@ type PoolHourlySnapshot @entity {
...
+}
+
+enum TransferType {
+  MINT,
+  BURN,
+  LOCK,
+  RELEASE,
+}
+
+type BridgeTransfer implements Event @entity {
```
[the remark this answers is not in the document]
jaimehgb: It is implied in the `crosschainToken` field, which is something messages don’t have. But thinking about it now I prefer to have those fields explicitly in the transfer too. Thanks 👍

schema-bridge.graphql
jaimehgb: I’m not sure how I missed that 😅

schema-bridge.graphql
jaimehgb: `total` :P 
Renamed!

schema-bridge.graphql
jaimehgb: Added those

schema-bridge.graphql
jaimehgb: Oh, yeah, not really. If one is wrapped the other one will be canonical for sure. But one being canonical doesn’t imply the other one being wrapped. The vice versa is wrong there. Added more clarity there.

schema-bridge.graphql
jaimehgb: Have done the same thing then

schema-bridge.graphql
jaimehgb: I was hoping to be able to have the same ID for the same route in all the different deployments, but the ways I can think of achieving that are either not simple or would make it impossible on one of the sides of the chain to load a given route from a mapping knowing only about the pool and the chains involved.

Will leave it as is. Identifying same routes by the relationship of their inputToken and crossToken should work.

schema-bridge.graphql
jaimehgb: Added a more descriptive comment on that field then just in case. Kept the same name, since `receiverPoolType` implies that the other pool is a receiver but can also send funds through the route.

schema-bridge.graphql
```diff
@@ -176,21 +429,94 @@ type UsageMetricsDailySnapshot @entity {
...
+  id: Bytes!
+
+  " Same as the ID, but readable. Number of hours since Unix "
+  hour: Int!
 
   " Protocol this snapshot is associated with "
   protocol: Protocol!
 
   " Number of unique hourly active users "
   hourlyActiveUsers: Int!
```
dhruv-chauhan, line 442: `hourlyActiveUsers` repeated on line#462

schema-bridge.graphql
```diff
@@ -428,15 +897,338 @@ type PoolHourlySnapshot @entity {
...
+  fromChainID: Int!
+
+  " Chain ID where the transfer is being sent to "
+  toChainID: Int!
+
+  " This is the address that's receiving the bridged funds (not a pool contract). When bridging out, it will be the address on the destination chain. "
+  transferTo: Bytes!
+
+  " This is the address initiating the transfer, where the initial funds came from. It might not match the address initiating the actual EVM transaction. "
+  transferFrom: Bytes!
```
dhruv-chauhan, line 1106: For multichain, I do not get `transferFrom` from the bridge-in event params, can we make this optional?

schema-bridge.graphql
```diff
@@ -428,15 +897,338 @@ type PoolHourlySnapshot @entity {
...
+  deposits: [LiquidityDeposit!]! @derivedFrom(field: "account")
+
+  " Total number of liquidity withdrawals this account has made "
+  withdrawCount: Int!
+
+  " All withdraw events for this account "
+  withdraws: [LiquidityWithdraw!]! @derivedFrom(field: "account")
+
+  " Total number of messages this account has sent or received through this bridge "
+  messageCount: Int!
```
dhruv-chauhan, line 1224: suggest to keep `messageSentCount` (optionally also `messageReceivedCount`) instead, since this field will be used to calculate `protocol.cumulativeUniqueMessageSenders`

dhruv-chauhan (commented):
Also, I'm not sure if we can use `Bytes` for `id`'s. Running codegen gives me the following error,

<details>
  <summary>Failed to load subgraph from subgraph.yaml: Error in schema.graphql:</summary>

  ```
  Account:
  - Field 'id': Entity IDs must be of type ID!

  ActiveAccount:
  - Field 'id': Entity IDs must be of type ID!

  BridgeMessage:
  - Field 'id': Entity IDs must be of type ID!

  BridgeProtocol:
  - Field 'id': Entity IDs must be of type ID!

  BridgeTransfer:
  - Field 'id': Entity IDs must be of type ID!

  CrosschainToken:
  - Field 'id': Entity IDs must be of type ID!

  FinancialsDailySnapshot:
  - Field 'id': Entity IDs must be of type ID!

  LiquidityDeposit:
  - Field 'id': Entity IDs must be of type ID!

  LiquidityWithdraw:
  - Field 'id': Entity IDs must be of type ID!

  Pool:
  - Field 'id': Entity IDs must be of type ID!

  PoolDailySnapshot:
  - Field 'id': Entity IDs must be of type ID!

  PoolHourlySnapshot:
  - Field 'id': Entity IDs must be of type ID!

  PoolRoute:
  - Field 'id': Entity IDs must be of type ID!

  PoolRouteSnapshot:
  - Field 'id': Entity IDs must be of type ID!

  RewardToken:
  - Field 'id': Entity IDs must be of type ID!

  Token:
  - Field 'id': Entity IDs must be of type ID!

  UsageMetricsDailySnapshot:
  - Field 'id': Entity IDs must be of type ID!

  UsageMetricsHourlySnapshot:
  - Field 'id': Entity IDs must be of type ID!
  ```
</details>

cc @dmelotik

dmelotik:
@dhruv-chauhan what version of graph-cli are you using? this was not introduced until 0.28.0 and graph-ts version 0.26.0

schema-bridge.graphql
jaimehgb: Hmm, that makes sense 🤔 

Although, does multichain allow to send the transaction to a different address? Maybe it’s the same as the sender?

schema-bridge.graphql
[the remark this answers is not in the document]
jaimehgb: Will keep it as is, for consitency with the rest of schemas

Issue #1453, opened 2022-11-24
Lido Ethereum missing hourly snapshots


lowjiansheng:
It seems like all the hourly snapshots are timestamped exactly 24 hours apart. This is perfectly valid if no activities occur on lido in 24 hours, but it seems too much of a coincidence for all the timestamps to be around 12pm EST separated 24 hours apart.

Query:
```
{
  poolHourlySnapshots (orderBy:timestamp, orderDirection: desc) {
    id
    timestamp
    pool {
      id
      name
    }
  }
}
```

Response:
```
{
  "data": {
    "poolHourlySnapshots": [
      {
        "id": "463692",
        "timestamp": "1669292699" (GMT: Thursday, 24 November 2022 12:24:59),
        "pool": {
          "id": "0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
          "name": "Liquid Staked Ether 2.0"
        }
      },
      {
        "id": "463668",
        "timestamp": "1669206251" (GMT: Wednesday, 23 November 2022 12:24:11) ,
        "pool": {
          "id": "0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
          "name": "Liquid Staked Ether 2.0"
        }
      },
      {
        "id": "463644",
        "timestamp": "1669119779" (GMT: Tuesday, 22 November 2022 12:22:59),
        "pool": {
          "id": "0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
          "name": "Liquid Staked Ether 2.0"
        }
      },
      {
        "id": "463620",
        "timestamp": "1669033355" (GMT: Monday, 21 November 2022 12:22:35),
        "pool": {
          "id": "0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84",
          "name": "Liquid Staked Ether 2.0"
        }
      }
...
```

PR #1454, opened 2022-11-25
fix(#patch): lido-ethereum; update snaphsots and tvl more often

jaimehgb:
We were only updating snapshots when there was a mint of stETH. This only happens once every 24 hours, so hourly snapshots were only being created once daily.

Had to make some readjustments to the way we take snapshots in this subgraph and the times we update TVL and revenues:
- TVL was not being updated when a deposit was made, only on daily revenue mint. **Fixed this.**
- Snaphots are not updated as a whole. Only certain parts of it at a time (tvl alone, supply side revenue alone, total revenue alone, protocol revenue alone). Since revenue only increases every 24h, taking a snapshot before then would make 0 cumulative values on revenue on the hourly snapshots. **Added all the copying cumulative values on all snapshot taking functions.**
- Total revenue was calculated by observing mint events, and then calling the oracle to fetch the last daily reward update. **I moved that to its own event handler emitted from the oracle.** Functionality is the same, but slightly more efficient and clearer to read and understand.
  - We still look for mint events to see which ones go to treasury and operators (protocol revenue).
- We were using Lido’s internal contract shares as `OutputTokenSupply`. The number of shares is not the supply of stETH. **Fixed that too.**

Solves this: #1453

Syncing here (name is saddle-finance because of reusing eth deployments): https://okgraph.xyz/?q=Qmd13EkC2ksj4ETcnSpJfncorf4ynfwyK7UUTkHrxiL4hY

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2075,7 +2075,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.0.1",
+          "subgraph": "1.0.2",
           "methodology": "1.0.0"
         },
         "files": {
--- a/subgraphs/lido/protocols/lido/config/templates/lido.template.yaml
+++ b/subgraphs/lido/protocols/lido/config/templates/lido.template.yaml
@@ -76,3 +76,75 @@ dataSources:
         - event: Transfer(indexed address,indexed address,uint256)
           handler: handleTransfer
       file: ./src/mappings/Lido.ts
+  - kind: ethereum
+    name: LidoOracle
+    network: mainnet
+    source:
+      address: "0x442af784a788a5bd6f42a01ebe9f287a871243fb"
+      abi: LidoOracle
+      startBlock: 11473216
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.6
+      language: wasm/assemblyscript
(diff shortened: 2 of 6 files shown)


dmelotik:
> name is saddle-finance because of reusing eth deployments

@jaimehgb you can acutally create new ethereum hosted service subgraphs..

jaimehgb:
> > name is saddle-finance because of reusing eth deployments
> 
> @jaimehgb you can acutally create new ethereum hosted service subgraphs..

Oh really? 😅 
I thought it had been deprecated?

dmelotik (approved):
nice update! lgtm

jaimehgb:
@bye43 can you take a quick glance at this Lido version?
Relevant changes are:
- Fixing hourly snapshots (they were only appearing in 24h intervals)
- Refactor of revenue calculation (but should have the same values)
- Changes to `outputTokenSupply`, before it was tracking internal contract shares. Now it tracks stETH supply

https://thegraph.com/hosted-service/subgraph/jaimehgb/saddle-finance (it has a pending version, but it is exactly the same as the main one. Depends on when you get to review this it might still be syncing or not. In any case, using the current version will give the same result)

bye43:
@jaimehgb This looks good to go!

Issue #1425, opened 2022-11-16
Liquity Ethereum Rewards


bye43:

jaimehgb:
This is ready for review: https://okgraph.xyz/?q=QmSSafuj1YQ4aMCv6zxofF2L2cvUQ7acCjBBnmdDfbwELy

bye43:
@pietydespises can you do a quick scan on this?

pietydespises:
the "rewardAPR" graph in the "MarketDailySnapshots" section for the Stability Pool under the "Pool" tab looks off near the beginning (looks wrong is wrong). Any idea what's happneing here https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fid%2FQmSSafuj1YQ4aMCv6zxofF2L2cvUQ7acCjBBnmdDfbwELy&tab=pool&poolId=0x66017D22b0f8556afDd19FC67041899Eb65a21bb&view=marketDailySnapshots-rewardAPR 
? @jaimehgb everything else looks good!

jaimehgb:
@pietydespises Seems to be because of LQTY price being pretty high on the day of launch:
<img width="854" alt="Screenshot 2022-12-03 at 21 16 13" src="https://user-images.githubusercontent.com/17258946/205470840-2475f19b-52e8-4a1d-b25a-6d4da546e2da.png”>

If daily emissions were 60,000, and price (from coingecko) was $73, that’s 4,380,000 daily USD reward emissions. Ours is at ~5.1M, slightly higher, which might be due to using the price at some point during the day instead of an average.

Do we want to replace the price at the day of launch with something more realistic? Or since it was the price at which it traded should we leave it as is? I can do either option :)

jaimehgb:
Zooming in on that day gives even crazier prices:
<img width="292" alt="Screenshot 2022-12-03 at 21 25 29" src="https://user-images.githubusercontent.com/17258946/205470999-e0686aa9-7726-4188-9df3-082301503c68.png">

pietydespises:
this makes sense! I think it's ok to leave it in as is, users will be able to see a more sensible graph as long as their timeline isn't set to "All"

PR #1458, opened 2022-11-28
feat(#minor): added liquity rewards

jaimehgb:
This PR contains 2 main changes:
- Adding the StabilityPool as a Market entity
  - The stability pool is a pool where people can deposit LUSD which gets then converted to ETH as liquidations occur on the protocol. These depositors are rewarded in LQTY. This pool has its own TVL, deposits, withdrawals, and revenue
  - Revenue from this pool was added directly to the protocol before. Now it is added to the stability pool market entity (and to the protocol as a consequence)
  - Deposits and withdrawals from this pool are now tracked (they weren’t before)
  - LUSD price was hardcoded to 1, now it is fetched from the price lib
- Adding LQTY rewards to the stability pool

Comparing TVL, revenue and deposits from this version and the previous, everything mostly matches. Revenue is a bit different because of LUSD pricing changes. Deposit volume has increased because we weren’t tracking deposits on this pool before.

https://okgraph.xyz/?q=QmNuqVWM1nXknfcgPqmY1DR75spEuKwLCFKAmyxTHgV8bg


jaimehgb:
@pietydespises You QA’d liquity before. Can you take a quick look to the new values and rewards? I’ve already compared most of it to our currently deployed version and everything makes sense to me, but would be good to get a review from a fresh pair of eyes before deploying it to prod.

subgraphs/liquity/src/mapping/TroveManager.ts
```diff
@@ -185,10 +212,99 @@ function liquidateTrove(event: TroveUpdated, trove: _Trove): void {
...
+ *
+ * @param event Liquidation event
+ */
+export function handleLiquidation(event: Liquidation): void {
+  // To accurately calculate the revenue generated by the Stability pool we
+  // compare the amount of LUSD that was burned to the amount of ETH that was
+  // sent to the StabilityPool. The USD difference of both amounts will be the
+  // StabilityPool revenue.
+  if (!event.receipt) {
+    log.info(
```
dmelotik, line 232: should this be an error as receipt is expected?

dmelotik (commented):
Increment versions 😉. It would also be good to add a few things to the readme ab rewards as well

dmelotik (approved):
sweet another one done!

pietydespises:
responded to @jaimehgb here: https://github.com/messari/subgraphs/issues/1425#issuecomment-1336295573

PR #1460, opened 2022-11-29
Added eslint custom rule to encourage address literals to be lowercase

jaimehgb:
I’ve been having issues because of the casing of addresses. Am I the only one? 😅 

When we get an `Address` anywhere in our mappings, if we call `toHexString()` it returns the address all in lowercase (not checksumed). We usually do address comparison using their string values instead of `Address.equals` or `Address.notEqual`. 

This creates scenarios where `SomeEntity.id = SOME_CONSTANT_ADDRESS`, which if we happen to compare `event.address.toHexString() == SomeEntity.id` it won’t match. This might lead to duplicate entities (one with the lowercase ID and one with the checksumed ID), events going unhandled if some entity is missing (`Market.load(event.address.toHexString()` when ID is checksumed), and so on. There are a couple ways of preventing this:
- Enforce all our hardcoded constants to always be lowercase
- Enforce all address comparisons to be done with `Address.equals` instead of `toHexString() == …`

I’ve created a simple eslint rule to check for the first option (easier to implement). Also, the second option might still lead to errors if we do something like:
```
if (event.address.equals(Address.fromString(SOME_CONSTANT)) {
  // do something
  Entity.load(SOME_CONSTANT);
}
...
// And then somewhere else
Entity.load(event.address.toHexString();
```

### Things against

Having the linter enforce addresses to be lowercase will probably cause issues in already existing subgraphs if we change a constant and graft. Because if some entity ID is checksumed and we change all the code logic to now deal with lowercase addresses there will be a mismatch there. It should be safe to enforce on new subgraphs, and subgraphs we index from scratch. It might be reasonable to only enable this rule for currently non-existing files somehow. Let me know your thoughts on this.


dmelotik:
I have ran into this, I usually opt to compare `Address` types to prevent this

tnkrxyz:
I ran into this mysel. Very helpful improvement.

In terms of potential impact to grafting of existing subgraphs, is there a way we can know beforehand whether grafting works after changing an address constant?

subgraphs/_eslint-rules/index.js
```diff
@@ -0,0 +1,6 @@
+
+module.exports = {
+    rules: {
+        "no-checksum-addresses": require("./no-checksum-addresses"),
+    },
+};
```
dmelotik, line 6: add newline at EOF

subgraphs/_eslint-rules/no-checksum-addresses.js
```diff
@@ -0,0 +1,27 @@
...
+                        message: 'Ethereum addresses should be lowercase',
+                        fix: function (fixer) {
+                            return fixer.replaceText(node, node.raw.toLowerCase());
+                        },
+                    });
+                }
+            },
+        };
+    }
+}
```
dmelotik, line 27: new line at EOF

dmelotik (commented):
> Because if some entity ID is checksumed and we change all the code logic to now deal with lowercase addresses there will be a mismatch there.

This is a good point as `ID` type id's are case sensitive. But in the case we graft and it doesn't work we just ignore the lint rule imo. There shouldn't be too many cases of this.

dmelotik:
@jaimehgb can you do a test file to see if it works properly?

jaimehgb:
Works :D 
Screenshot 2023-02-03 at 21 29 37

dmelotik (approved):
Nice!

Issue #1424, opened 2022-11-16
Vesta-Finance Arbitrum - Add Rewards


bye43:

bye43:
PR: https://github.com/messari/subgraphs/pull/1470

bye43:
@ishraq8 Can you QA these rewards when you get the chance? https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/vesta-finance-arbitrum&tab=poolOverview

ishraq8:
Rewards for stability pool are accurate. this should get merged.

PR #1470, opened 2022-11-30
feat (#minor); vesta-finance; added rewards emission

tnkrxyz:
This PR adds rewards emission for vesta-finance #1424 

- Test deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/vesta-finance-arbitrum?selected=logs
- Validation dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/vesta-finance-arbitrum&tab=protocol


dmelotik:
@tnkrxyz there are now negative revenues?

tnkrxyz:
> @tnkrxyz there are now negative revenues?

Oops, I must have messed up sth. Looking into it.

subgraphs/vesta-finance/src/entities/stabilitypool.ts
```diff
@@ -1,42 +1,31 @@
...
+  const VSTValue = bigIntToBigDecimal(VSTAmount).times(VSTPrice);
+  const assetToken = getOrCreateAssetToken(asset);
+  const totalAssetValue = bigIntToBigDecimal(assetAmount).times(
+    assetToken.lastPriceUSD!
   );
+  const stabilityPoolTVL = VSTValue.plus(totalAssetValue);
+  const stabilityPool = getOrCreateStabilityPool(event.address, asset, event);
+  stabilityPool.inputTokenBalance = VSTAmount;
+  stabilityPool.totalValueLockedUSD = stabilityPoolTVL;
+  stabilityPool.totalDepositBalanceUSD = stabilityPoolTVL;
```
jaimehgb, line 28: Should we make different the TVL and depositBalance? Users only deposit VST into the stability pool. The asset amount is given by the protocol to the users in exchange of their deposits.

So, TVL = vst + assetAmount, and deposit = vst?

jaimehgb: Oh, just saw that for liquity I made it the same way 😅 

What do you think? Should we change it?

subgraphs/vesta-finance/src/utils/logs.ts
```diff
@@ -0,0 +1,76 @@
+import { ethereum, Address, BigInt } from "@graphprotocol/graph-ts";
+
+// Topic0 of ERC20 Transfer event.
+// Transfer(address,address.uint256)
+export const ERC20_TRANSFER_TOPIC0 =
+  "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef";
+
+export class ERC20TransferLog {
```
jaimehgb, line 8: ❤️

jaimehgb:
> @tnkrxyz there are now negative revenues?

It is possible actually. Since now we track stability pool revenues as being (LiquidatedAmount - VST burned) it might make revenue negative sometimes.

tnkrxyz:
> @tnkrxyz there are now negative revenues?

OK, this is due to lack of price data for VST before 2022-11-23, before that I assume VST price = $1 (it is a stable coin), but this may be off. For example, transaction [0x58b2290bb2f54f31f54f0fa225d132e66a984091b756ddb1c1449949575dd20d](https://arbiscan.io/tx/0x58b2290bb2f54f31f54f0fa225d132e66a984091b756ddb1c1449949575dd20d#eventlog) (logIndex 22) liquidates 683636363636363637 collateral (gOHM token price =$75.95 according to first event in the same transaction, total value $51.92) for 2460000000000000000000 VST debt (~=$2460 assuming VST = $1). This leads to a loss (negative revenue) of $2408 to the supply side of gOHM stability pool. 

It is likely that my assumption of VST price = $1 is off, but the subgraph has always made the assumption when [calculating BorrowBalanceUSD](https://github.com/tnkrxyz/subgraphs/blob/03e6ebeabbfccec704e41bfc14f76eb0669a53c0/subgraphs/vesta-finance/src/entities/market.ts#L305)

I tried to use [this balancer pool](https://app.balancer.fi/#/arbitrum/pool/0x5a5884fc31948d59df2aeccca143de900d49e1a300000000000000000000006f) to [get VST Price](https://github.com/tnkrxyz/subgraphs/blob/03e6ebeabbfccec704e41bfc14f76eb0669a53c0/subgraphs/vesta-finance/src/entities/token.ts#L186), but the pool was deployed at block 9455379 (April 7). Also for some reason, `getVSTTokenPriceFromBalancer` call was not successful. I am adding more log to see what is happening.  

Although, even with balancer oracle fixed, it won't help with the negative revenue problem as it happened before April. I can set the VST price to be 0, but it will create bigger supply side revenue than real. 

@dmelotik, @jaimehgb , I appreciate your input here.

subgraphs/vesta-finance/src/entities/stabilitypool.ts
tnkrxyz: That's a good point! It makes sense to change it.

jaimehgb:
VST has always been around `0.9-1.0` USD: https://www.coingecko.com/en/coins/vesta-stable

It might be something with the pricing of gohm

subgraphs/vesta-finance/src/entities/stabilitypool.ts
jaimehgb: We can leave it as is then 👍

jaimehgb (approved):
Looks correct to me! 🚀

PR #1472, opened 2022-11-30
fix(#patch); dForce; reward array mismatch

dmelotik:
Reward arrays could be mismatched and were. We cannot allow this so this was updated.

Also refactored rewards to be more readable and update emissions with events.

Testing:
- arb: https://okgraph.xyz/?q=dmelotik%2Fdforce-arbitrum
- bsc: https://okgraph.xyz/?q=dmelotik%2Fdforce-bsc
- eth: https://okgraph.xyz/?q=dmelotik%2Fdforce-ethereum
- optimism: https://okgraph.xyz/?q=dmelotik%2Fdforce-optimism
- polygon: https://okgraph.xyz/?q=dmelotik%2Fdforce-polygon

### DIFF
--- a/subgraphs/compound-forks/protocols/dforce/src/mapping.ts
+++ b/subgraphs/compound-forks/protocols/dforce/src/mapping.ts
@@ -151,7 +151,7 @@ export function handleMarketAdded(event: MarketAdded): void {
       underlyingName = "Maker";
       underlyingSymbol = "MKR";
       underlyingDecimals = 18;
-    } else if (underlyingTokenAddr == DF_ADDRESS) {
+    } else if (underlyingTokenAddr.toHexString() == DF_ADDRESS) {
       underlyingName = "dForce";
       underlyingSymbol = "DF";
       underlyingDecimals = 18;
@@ -619,19 +619,11 @@ function initRewards(marketAddr: Address): void {
 // reward is always DForce token
 function getOrCreateRewardTokens(): string[] {
   let token = Token.load(rewardTokenAddr.toHexString());
-  if (token == null) {
+  if (!token) {
     token = new Token(rewardTokenAddr.toHexString());
-    // special handle for DF token since its name and symbol is byte32
-    if (rewardTokenAddr == DF_ADDRESS) {
-      token.name = "dForce";
-      token.symbol = "DF";
-      token.decimals = 18;
-    } else {
-      const tokenContract = ERC20.bind(rewardTokenAddr);
-      token.name = getOrElse<string>(tokenContract.try_name(), "unknown");
-      token.symbol = getOrElse<string>(tokenContract.try_symbol(), "unknown");
-      token.decimals = getOrElse<i32>(tokenContract.try_decimals(), 18);
-    }
+    token.name = "dForce";
+    token.symbol = "DF";
+    token.decimals = 18;
     token.save();
   }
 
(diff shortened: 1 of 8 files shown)


subgraphs/compound-forks/protocols/dforce/src/mapping.ts
```diff
@@ -152,7 +151,7 @@ export function handleMarketAdded(event: MarketAdded): void {
       underlyingName = "Maker";
       underlyingSymbol = "MKR";
       underlyingDecimals = 18;
-    } else if (underlyingTokenAddr.toHexString() == DF_ADDRESS) {
+    } else if (underlyingTokenAddr == DF_ADDRESS) {
```
jaimehgb, line 154: Hmm, are you sure this works? Usually on JS/TS, comparing objects or arrays only returns true when you are comparing to references to the same instance:
```
const a = {};
const b = {}
let c = a;
a == b // false
a == c // true
```

```

Comment (tnkrxyz) [reply subgraphs/compound-forks/protocols/dforce/src/mapping.ts:154]: I have been doing this myself - I think it works for string?

```
> const a = "0x01234567890abcdef"
undefined
> const b = "0x01234567890abcdef"
undefined
> a == b
true
```

File: subgraphs/compound-forks/protocols/dforce/src/mapping.ts [hunk 2]
```diff
@@ -704,3 +551,166 @@ export function getOrCreateMarketStatus(marketId: string): _DforceMarketStatus {
...
+}
+
+function updateRewardPrices(marketAddress: Address): void {
+  const market = Market.load(marketAddress.toHexString());
+  if (!market) {
+    log.error("Market not found for address {}", [marketAddress.toHexString()]);
+    return;
+  }
+  const rewardToken = RewardToken.load(market.rewardTokens![0]);
+  if (!rewardToken) {
```
tnkrxyz, line 670: This is unlikely to happend, but if it does, should we call `getOrCreateRewardTokens` fix the error?

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
jaimehgb: Yeah, strings, booleans and numbers are fine. Usually the problem is with variables whose value is a reference: objects, classes, arrays…

Taking a look at the `Address` implementation, it extends from `ByteArray`, which overloads the `==` operator: https://github.com/graphprotocol/graph-ts/blob/main/common/collections.ts#L259

So whenever you do `Address == Address`, what happens under the hood is that it runs `Address.equals(address)`. It should work then 🤔 

I’d suggest not relying on this in general, since you cannot expect it to always be the case. Safer to call `.equals` explicitly, you always know what to expect from that.

tnkrxyz (commented):
Nice improvement! Looks good other than the Address equal check Jaime pointed out.

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
tnkrxyz: Good point! I learned something. thx!

dmelotik: Good catch! fixed

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
dmelotik: I guess if this isn't there then there are likely more issues. So I guess I would want to throw an error and quit.. What do you think?

jaimehgb: Since this is a very common occurence when we build subgraphs we can try to come up with some standard decision. I think the options are:
- Log an error and skip the calculation that follows
  - good: the subgraph doesn’t break, keeps indexing
  - bad: no one looks at errors, and if we do, they get buried in other logs. Might never get fixed unless it affects data quality noticeably 
- Have some fallback or use `getOrCreate` to ensure the entity is always there
  - good: the subgraph doesn’t break and keeps indexing
  - bad: if the entity was supposed to be there and wasn’t, we won’t even log an error. Might be affecting something else somewhere and make it harder to debug when we create the entity “again”.
- Fail with a critical error, to make sure someone comes and fix the underlying issue
  - good: we ensure someone will take a look at it
  - bad: subgraph stops syncing until fixed. Whomever comes to fix it might just default to any of the above options, adding one step.

Most of this I think will be fixed with the libraries we build for each schema though. And maybe, instead of this, we can think on what each log level means. If we purely keep errors logs at a minimum, only logging what really signals that something needs to get fixed, together with the log persistence tool we could both know that an error happened and keep the subgraph going while we fix it. Will bring up this discussion on an issue or discord to not block this PR.

Getting back to the topic at hand 😅 I don’t like any of the options at hand, but I prefer the `log an error and return` option too.

jaimehgb (approved):

subgraphs/compound-forks/protocols/dforce/src/mapping.ts
tnkrxyz: makes sense.

dmelotik:
Does not change values, merging, no need to QA @ishraq8

PR #1492, opened 2022-12-05
Bridge Schema Lib

jaimehgb:


subgraphs/synapse/src/sdk/common/accounts.ts
```diff
@@ -0,0 +1,60 @@
+import { Address, ethereum } from "@graphprotocol/graph-ts";
```
dmelotik, line 1: Should this be under bridge? At least for lending it is different

PR #1497, opened 2022-12-06
feat(); Uniswap V3 Grant; This includes upgrade and refactor of Uniswap V3  (Grant + Schema 3.0.0)

steegecs:
**Context:**
The purpose of this PR is to upgrade the Uniswap V3 schema and refactor the code. The purposes of these changes are to upgrade the schema to version 2.0.0 and get the relevant data for the Uni Grant.


schema-dex-amm-extended.graphql
```diff
@@ -652,6 +805,121 @@ type LiquidityPoolHourlySnapshot @entity {
...
+  closedPositionCount: Int!
+}
+
+####################################
+##### Pool Tick-Level Metadata #####
+####################################
+
+type Tick @entity {
+  " { pool address }-{ tick index } "
+  id: Bytes!
```
jaimehgb, line 827: It’s probably still early to know, but do you think these ticks will be the way other concentrated liquidity DEXes will work?

schema-dex-amm-extended.graphql
```diff
@@ -960,7 +1291,7 @@ type Swap @entity {
 
 type Stat @entity {
   " { protocol/pool }-{ [Optional] Pool ID }-{ deposit/withdraw/swap }-{ Days/hours since Unix epoch } "
-  id: ID!
+  id: Bytes!
```
jaimehgb, line 1294: Are you still using `Stat`?

subgraphs/uniswap-v3/src/common/entities/account.ts
```diff
@@ -0,0 +1,33 @@
+import { Address } from "@graphprotocol/graph-ts";
+import { Account } from "../../../generated/schema";
+import { INT_ZERO } from "../constants";
+
+export function createAccount(address: Address): Account {
```
jaimehgb, line 5: I think this one is unused? Do you need to export it while having already `getOrCreateAccount`?

subgraphs/uniswap-v3/src/common/entities/pool.ts
```diff
@@ -0,0 +1,208 @@
...
+  pool._totalAmountEarned = [BIGINT_ZERO, BIGINT_ZERO];
+  pool._mostRecentSnapshotsDayID = INT_ZERO;
+  pool._mostRecentSnapshotsHourID = INT_ZERO;
+
+  poolAmounts.inputTokens = [token0.id, token1.id];
+  poolAmounts.inputTokenBalances = [BIGDECIMAL_ZERO, BIGDECIMAL_ZERO];
+  poolAmounts.tokenPrices = [BIGDECIMAL_ZERO, BIGDECIMAL_ZERO];
+
+  // Used to track the number of deposits in a liquidity pool
+  const poolDeposits = new _HelperStore(poolAddress);
```
jaimehgb, line 111: Oh, I like this 🤔 
Why did you decide to only use it for deposits

subgraphs/uniswap-v3/src/common/utils/backfill.ts
```diff
@@ -56,40 +53,70 @@ export function populateEmptyPools(event: ethereum.Event): void {
...
+
+    pool.positionCount = INT_ZERO;
+    pool.openPositionCount = INT_ZERO;
+    pool.closedPositionCount = INT_ZERO;
+
+    pool._totalAmountWithdrawn = [BIGINT_ZERO, BIGINT_ZERO];
+    pool._totalAmountCollected = [BIGINT_ZERO, BIGINT_ZERO];
+    pool._totalAmountEarned = [BIGINT_ZERO, BIGINT_ZERO];
+    pool._mostRecentSnapshotsDayID = INT_ZERO;
+    pool._mostRecentSnapshotsHourID = INT_ZERO;
```
jaimehgb, line 100: Is this the only place where you are setting the default 0 values for a pool? If not, can you combine them all into one? If not, I fear that whenever a new field is added for the Pool entity we’ll forget to add it here too.

subgraphs/uniswap-v3/src/mappings/core.ts
```diff
@@ -1,72 +1,216 @@
...
+    event,
     event.transaction.from,
-    event.params.sqrtPriceX96
+    pool,
+    true,
+    swapHelperObj.balanceChangesNet,
+    BIGINT_ZERO,
+    swapHelperObj.activeLiquidityDelta,
+    swapHelperObj.supplyUncollectedTokenChangesNet,
+    swapHelperObj.protocolUncollectedTokenChangesNet
```
jaimehgb, line 137: Does it make sense to have `DexEventHandler` have `SwapHelperObj` as an argument instead of its fields? This way it is easier for a user (of DexEventHandler) to know what to pass to it. Also reduces the chances of passing the fields swapped.

Or is it used in more cases than just Swaps? If this is the case, instead of instantiating it with the constructor you might have different functions (static maybe) that initialize the `DexEventHandler` in different ways, depending on what might be available at each time: `Deposit, Swap, Withdraw, etc`

schema-dex-amm-extended.graphql
```diff
@@ -295,6 +317,15 @@ type UsageMetricsDailySnapshot @entity {
...
 
+  " Total number of deposits (add liquidity) in an hour "
+  dailyDepositCount: Int!
+
+  " Total number of withdrawals (remove liquidity) in an hour "
+  dailyWithdrawCount: Int!
+
+  " Total number of trades (swaps) in an hour "
+  dailySwapCount: Int!
+
```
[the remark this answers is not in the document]
steegecs: Fixed!

schema-dex-amm-extended.graphql
```diff
@@ -295,21 +322,33 @@ type UsageMetricsDailySnapshot @entity {
...
 
-  " Statistics around deposit events "
-  depositStats: Stat!
+  # " Statistics around deposit events "
+  # ## depositStats: Stat!
 
-  " Statistics around withdraw events "
-  withdrawStats: Stat!
+  # " Statistics around withdraw events "
+  # ## withdrawStats: Stat!
```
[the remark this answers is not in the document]
steegecs: Removed

schema-dex-amm-extended.graphql
```diff
@@ -435,21 +519,24 @@ type LiquidityPool @entity {
...
+
   " Weights of input tokens in the liquidity pool in percentage values. For example, 50/50 for Uniswap pools, 48.2/51.8 for a Curve pool, 10/10/80 for a Balancer pool "
   inputTokenWeights: [BigDecimal!]!
 
-  " Total supply of output token. Note that certain DEXes don't have an output token (e.g. Bancor) "
-  outputTokenSupply: BigInt
-
-  " Price per share of output token in USD "
-  outputTokenPriceUSD: BigDecimal
-
```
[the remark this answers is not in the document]
steegecs: Yes on `totalLiquidity`, I removed the price per unit of the liquidity as it should be easily derivable from `totalLiquidity` and `totalLiquidityUSD`. WDYT?

schema-dex-amm-extended.graphql
```diff
@@ -499,7 +586,10 @@ type LiquidityPool @entity {
 
 type LiquidityPoolDailySnapshot @entity {
```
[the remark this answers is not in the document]
steegecs: Actually, I can make these immutable now. I have already set it so that the snapshots occur once per time interval. Updated this too.

schema-dex-amm-extended.graphql
steegecs: I think yes, but I am not certain.

schema-dex-amm-extended.graphql
steegecs: No I am not still using stat.

steegecs: I removed it.

subgraphs/uniswap-v3/src/common/entities/account.ts
steegecs: No I don't suppose so XD. Removed

subgraphs/uniswap-v3/src/common/entities/pool.ts
steegecs: This has been used in the pricing mechanism. But I agree that it would make sense to start tracking swap, deposit, and  withdraw counts at the pool level. Do you think I should add that in now ?

subgraphs/uniswap-v3/src/mappings/core.ts
steegecs: So I have it set up so that the constructor takes a set of deltas. Then the deltas are propagated to update all the fields in the entities. What do you think about creating a function for each that instantiates the same `Deltas` type, and then pass that along for the last 5 args? Or one function for Deposits + Withdraws and one for Swaps.

steegecs: I like passing the deltas, because the constructor initializes more or less the same based on the deltas. The only thing that is different is that the deltas count for volume with swaps, which is why I pass the flag. wdyt?

subgraphs/uniswap-v3/src/common/utils/backfill.ts
steegecs: Agreed. Fixed.

schema-dex-amm-extended.graphql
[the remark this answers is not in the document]
jaimehgb: Makes total sense. Only downside is not being able to query directly by liquidity price, but I don’t think that is so relevant?

subgraphs/uniswap-v3/src/common/entities/pool.ts
jaimehgb: Actually, the opposite. It might make sense to put those counters away from the pool entity because every change to them results in a whole copy of the Pool, increasing the DB and index size. So having values that change often in their dedicated store would make the `Pool` table more lightweight.

Let’s not add it for now, but we can keep it in mind for new schemas / subgraphs. Specially for the ones where we expect a lot of `Pools` to exist.

subgraphs/uniswap-v3/src/mappings/core.ts
jaimehgb: >  What do you think about creating a function for each that instantiates the same Deltas type, and then pass that along for the last 5 args?

Yes, I like that. That way you also encapsulate the logic that creates the deltas from a given event type. Which could eventually be generalized.

subgraphs/uniswap-v3/src/common/entities/stat.ts
```diff
@@ -0,0 +1,28 @@
+import { Bytes, ethereum } from "@graphprotocol/graph-ts";
```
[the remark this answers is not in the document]
steegecs: rm rf stat

jaimehgb (approved):

PR #1515, opened 2022-12-13
chore(); Uniswap Forks; Add immutable entities to uniswap forks

steegecs:
**Context:**
Add immutable entities to Uniswap Forks.


dmelotik:
I belv acct and activeacct can also be immutable, but I will let you decide if that is the best thing to do c✅

Issue #1484, opened 2022-12-01
CI / Lint additions


dmelotik:
- Add check for folder/file naming to stay standard file name scheme being https://google.github.io/styleguide/tsguide.html#identifiers-imports
- ~~Add a new GH action to check for formatting separate of the other actions~~ Done by @jaimehgb 
- Add lint check so we have no string literals

jaimehgb:
> Add lint check for literals 

This one refers to the addresses? @dmelotik

dmelotik:
@jaimehgb I mean no string literals

dmelotik:
TODO:
- https://eslint.org/docs/latest/rules/camelcase ignore globals, and ensure that it doesn't break any graph imports
- https://eslint.org/docs/latest/rules/max-lines-per-function add max lines AND max indents to make functions easier to read
- https://eslint.org/docs/latest/rules/no-magic-numbers helping with literals

There are no "out of the box" eslint rules to prevent string literals. We would either have to make our own or use a third party. Lowest priority given the work req'd.

tnkrxyz:
@dmelotik , I started to work on this task, draft PR in #1955.

> - [X] Add check for folder/file naming to stay standard file name scheme being https://google.github.io/styleguide/tsguide.html#identifiers-imports

done.

> - Add lint check so we have no string literals

I need a bit clarification what is allowed, what is not:
- allowed:
   - template literal: `${var1}-${var2}`
   - log.error("[error] variable value: {}", [])

- disallowed:
   - string concatenation: var1 + "-" + var2
 
What else?

> * https://eslint.org/docs/latest/rules/camelcase ignore globals, and ensure that it doesn't break any graph imports

[camelcase is deprecated](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/camelcase.md). it is suggested to use naming-convention. I suppose we will exclude imported names. What about single word variable name, like `vault`? leading "_" ok?

> * https://eslint.org/docs/latest/rules/max-lines-per-function add max lines AND max indents to make functions easier to read

I could not find this rule in the aslant-plugin installation: https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin/src/rules . Maybe also deprecated?

> * https://eslint.org/docs/latest/rules/no-magic-numbers helping with literals

done.

> 
> There are no "out of the box" eslint rules to prevent string literals. We would either have to make our own or use a third party. Lowest priority given the work req'd.

[...]

PR #1525, opened 2022-12-15
feat(): Added first version of SDK for bridges

jaimehgb:
This PR adds an initial version of the SDK for bridges. It is well documented, but the current way of using it is to copy and paste. I’ve tried to make it independent of any ABIs, so that it is easier to auto-generate or package into npm or something. 
But we’ll cross that bridge 🙃 

It is a big PR, so let me break it down a bit:
- Made some changes to the bridge schema:
  - Added a bunch of network names
  - Changed chainIDs from `Int` to `BigInt`, since some chains have absurdly big IDs
  - Minor docs updates
  - Added auxiliary entities for the SDK to track account activity and snapshots
- Added the SDK itself:
  - Consists on classes that wrap around the schema entities. When we want to modify any of these entities we do so through these wrappers. This takes care of proper array ordering, updating downstream fields, taking snapshots automatically.
  - Protocol wraps around the Protocol entity and exposes methods to modify all its fields
  - Pool, same thing as Protocol, but for the Pool
  - Account, is a utility to load accounts, track unique users, and create account actions:  deposit, withdraw, transfer, etc
  - Tokens is a utility to instantiate tokens and crosschain tokens
  - All what’s contained under the `util` folder has been copied over from already existing subgraphs.

I recommend reading the docs first, to get an understanding of how things work to then jump at the code with a better idea. Read the docs here: https://github.com/jaimehgb/subgraphs/tree/bridge-sdk/subgraphs/_reference_/src/sdk
There are a few examples on how to use it too.


subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -0,0 +1,638 @@
...
+  log.error("Unknown pool type at inferCounterType {}", [poolType]);
+  log.critical("", []);
+  return poolType;
+}
+
+// A function which given 3 arrays of arbitrary types of the same length,
+// where the first one holds the reference order, the second one holds the same elements
+// as the first but in different order, and the third any arbitrary elements. It will return
+// the third array after sorting it according to the order of the first one.
+function sortArrayByReference<T, K>(
```
dmelotik, line 619: these are nice, we could consider adding them to the constants for all sdks to use

subgraphs/_reference_/src/sdk/protocols/bridge/pool_snapshot.ts
```diff
@@ -0,0 +1,321 @@
+import { ethereum, Bytes } from "@graphprotocol/graph-ts";
```
dmelotik, line 1: again, I have been using camelCase for ts files

dmelotik: But here it says underscores: https://google.github.io/styleguide/tsguide.html#identifiers-imports

🫤

subgraphs/_reference_/src/sdk/protocols/bridge/protocol.ts
```diff
@@ -0,0 +1,457 @@
...
+    protocol.cumulativeMessageSentCount = 0;
+    protocol.cumulativeMessageReceivedCount = 0;
+    protocol.supportedNetworks = [];
+    protocol.totalPoolCount = 0;
+    protocol.totalPoolRouteCount = 0;
+    protocol.totalCanonicalRouteCount = 0;
+    protocol.totalWrappedRouteCount = 0;
+    protocol.totalSupportedTokenCount = 0;
+
+    const proto = new Bridge(protocol, pricer, event);
```
dmelotik, line 102: why not get this at the beginning of the function so you don't need to load it 2x

subgraphs/_reference_/src/sdk/protocols/bridge/protocol_snapshot.ts
```diff
@@ -0,0 +1,408 @@
...
+  }
+}
+
+function initProtocolHelper(): _ProtocolSnapshotHelper {
+  let helper = _ProtocolSnapshotHelper.load(SnapshotHelperID);
+  if (helper) {
+    return helper;
+  }
+  helper = new _ProtocolSnapshotHelper(SnapshotHelperID);
+  helper._lastActivityTimestamp = BigInt.fromI32(0);
```
dmelotik, line 379: We should probably make that a constant imo

subgraphs/_reference_/src/sdk/protocols/config.ts
```diff
@@ -0,0 +1,45 @@
...
+  getVersions(): Versions;
+}
+
+export class ProtocolConfig implements ProtocolConfigurer {
+  id: string;
+  name: string;
+  slug: string;
+  versions: Versions;
+
+  constructor(id: string, name: string, slug: string, versions: Versions) {
```
dmelotik, line 18: I'm surprised this line didn't get formatted...

dmelotik (commented):
Looks pretty nice. I made some comments, some are debatable

schema-bridge.graphql
```diff
@@ -1,5 +1,5 @@
 # Subgraph Schema: Bridge
-# Version: 1.0.0
+# Version: 1.0.1
```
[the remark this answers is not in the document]
jaimehgb: Well, technically it should be major, since it is a breaking change. Old code won’t work with this schema change. But felt too premature to make it major 😅 

Can definitely do minor though 👍

subgraphs/_reference_/src/sdk/util/constants.ts
```diff
@@ -0,0 +1,212 @@
+import { BigDecimal, BigInt } from "@graphprotocol/graph-ts";
```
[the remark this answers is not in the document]
jaimehgb: Sure, let me rebase my branch onto yours and merge that.

schema-bridge.graphql
```diff
@@ -690,10 +751,13 @@ type Pool @entity {
...
   " Pool hourly snapshots "
   hourlySnapshots: [PoolHourlySnapshot!]! @derivedFrom(field: "pool")
+
+  _lastDailySnapshotTimestamp: BigInt
+  _lastHourlySnapshotTimestamp: BigInt
 }
 
 type PoolRoute @entity {
-  " Bytes.fromUTF(({Pool ID}-{lowerChainID}-{greaterChainID}) "
+  " Bytes.fromUTF(({Pool ID}-{crossToken}-{lowerChainID}-{greaterChainID}) "
```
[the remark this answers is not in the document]
jaimehgb: I know 😅 
In this case, the chainIDs might have the issue of `12-23` being equal to `1-223`.

An alternative would be to always pad these numbers with a given number of zeroes or something. Since it is going to be abstracted by the library it’s fine. What do you think?

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
```diff
@@ -0,0 +1,423 @@
...
+    acc.messageCount = 0;
+    acc.save();
+
+    this.protocol.addUser();
+
+    return new Account(this.protocol, acc, this.tokens);
+  }
+}
+
+namespace ActivityType {
```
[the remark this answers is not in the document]
jaimehgb: It is not exported, nor needed by any other files. I think it makes sense to keep here.

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
```diff
@@ -0,0 +1,423 @@
...
+  export const TRANSFER_OUT = "transferOut";
+  export const TRANSFER_IN = "transferIn";
+  export const LIQUIDITY_PROVISIONING = "deposit";
```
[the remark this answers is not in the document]
jaimehgb: Oh, that’s a mistake 🙈

subgraphs/_reference_/src/sdk/protocols/bridge/chain-ids.ts
```diff
@@ -0,0 +1,101 @@
+import { BigInt, TypedMap } from "@graphprotocol/graph-ts";
```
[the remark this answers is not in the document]
jaimehgb: No particular reason, I’ll change to camelCase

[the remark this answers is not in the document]
jaimehgb: Resolving duplicates

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -0,0 +1,638 @@
...
+    this.pool.cumulativeVolumeIn = BIGINT_ZERO;
+    this.pool.cumulativeVolumeOut = BIGINT_ZERO;
+    this.pool.netVolume = BIGINT_ZERO;
+    this.pool.cumulativeVolumeInUSD = BIGDECIMAL_ZERO;
+    this.pool.cumulativeVolumeOutUSD = BIGDECIMAL_ZERO;
+    this.pool.netVolumeUSD = BIGDECIMAL_ZERO;
+
+    this.pool._lastDailySnapshotTimestamp = BIGINT_ZERO;
+    this.pool._lastHourlySnapshotTimestamp = BIGINT_ZERO;
+    this.save();
```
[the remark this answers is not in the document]
jaimehgb: Yeah 🤔 
Now it just calls `entity.save()`, but before it would also take care of updating the snapshots.

It’s also a place where to add any hooks that we want to run on save. Or if these wrappers managed more than 1 entity at a time you’d only need one save. It’s hypotheticals though

subgraphs/_reference_/src/sdk/protocols/bridge/pool_snapshot.ts
jaimehgb: Haha damn. Which one do you prefer? Let’s settle on one and add it to the CI

subgraphs/_reference_/src/sdk/protocols/bridge/protocol.ts
jaimehgb: I think I still need it 2x, unless I put all the protocol default values inside the `if`?

subgraphs/_reference_/src/sdk/protocols/config.ts
jaimehgb: It seems to break down into multiple lines when the number of arguments is >=5 or the line is longer than 81 chars. So its right at the limit 😅

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
```diff
@@ -0,0 +1,99 @@
...
+export class TokenManager {
+  protocol: Bridge;
+  initializer: TokenInitializer;
+
+  constructor(protocol: Bridge, init: TokenInitializer) {
+    this.protocol = protocol;
+    this.initializer = init;
+  }
+
+  initToken(address: Address): Token {
```
tnkrxyz, line 27: I don't have a better naming suggestion atm, but can we break it into `getToken` and `initToken` or just use the `getOrCreateToken` name we have been using? I don't expect to call initToken to return a token (already existing).

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -0,0 +1,638 @@
...
+   * @param token The actual token being rewarded
+   * @param amount The daily amount of reward tokens emitted to this pool
+   */
+  setRewardEmissions(
+    type: RewardTokenType,
+    token: Token,
+    amount: BigInt
+  ): void {
+    const rToken = this.tokens.initRewardToken(type, token);
+    const amountUSD = this.protocol.pricer.getAmountPrice(token, amount);
```
tnkrxyz, line 490: `getAmountPrice` -> `convertAmountToUSD`?

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -0,0 +1,638 @@
...
+
+  log.error("Unknown pool type at inferCounterType {}", [poolType]);
+  log.critical("", []);
+  return poolType;
+}
+
+// A function which given 3 arrays of arbitrary types of the same length,
+// where the first one holds the reference order, the second one holds the same elements
+// as the first but in different order, and the third any arbitrary elements. It will return
+// the third array after sorting it according to the order of the first one.
```
tnkrxyz: It'd helpful to add example inputs and output:
`sortArrayByReference(['a', 'c', 'b'], ['a', 'b', 'c'], [1, 2, 3]) => [1, 3, 2]`

Personally, I prefer switch the first and second argument.

subgraphs/_reference_/src/sdk/util/numbers.ts
```diff
@@ -0,0 +1,46 @@
+import { BigDecimal, BigInt } from "@graphprotocol/graph-ts";
```
tnkrxyz, line 1: If we are to consolidate code between lending, dex and bridge, this file is a candidate for consolidation too.

subgraphs/_reference_/src/sdk/util/events.ts
```diff
@@ -0,0 +1,9 @@
+import { ethereum } from "@graphprotocol/graph-ts";
+
+export function getUnixDays(event: ethereum.Event): i32 {
+  return event.block.timestamp.toI32() / 86400;
```
tnkrxyz, line 4: Use constant `SECONDS_PER_DAY`

subgraphs/_reference_/src/sdk/util/events.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export function getUnixHours(event: ethereum.Event): i32 {
+  return event.block.timestamp.toI32() / 3600;
```
tnkrxyz, line 8: use constant `SECONDS_PER_HOUR`

tnkrxyz (commented):
Nice work! I made a few minor comments

schema-bridge.graphql
```diff
@@ -23,6 +23,67 @@ enum Network {
   OSMOSIS
   MATIC # aka Polygon
   XDAI # aka Gnosis Chain
+  UBIQ
```
this-username-is-taken, line 26: I don't think there is a need to add any network that's not yet supported by The Graph

subgraphs/_reference_/src/sdk/protocols/bridge/README.md
```diff
@@ -0,0 +1,375 @@
...
+  );
+
+  pool.setRewardEmissions(RewardTokenType.DEPOSIT, rewardToken, dailyEmissions);
+}
+```
+
+## A Note On Snapshots
+
+You don't need to worry about snapshots anymore. At all. But you might want to know how they work under the hood.
+Snapshots from each entity that has them are taken when the given entity is loaded. If the time since the last snapshot was taken is longer than the snapshot period, a snapshot will be taken with the data the entity had at the time of the load. Any changes made to an entity after a load will be part of the next snapshot.
```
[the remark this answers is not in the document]
jaimehgb: I made it this way for 2 reasons:
- It’s easier to implement (now that snapshots are immutable)
- Taking the snapshot on save would potentially lead to some events being more than 1h/1d apart being part of the same snapshot. Since snapshots are immutable, and because we take them when the period has elapsed, taking them on some entity.save() would include the changes of that entity together with all the previous ones even if a whole month elapsed.

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -0,0 +1,638 @@
...
+    newBalance: BigInt,
+    updateMetrics: boolean = true
+  ): void {
+    this.pool.inputTokenBalance = newBalance;
+    if (updateMetrics) {
+      this.refreshTotalValueLocked();
+    }
+  }
+
+  getBytesID(): Bytes {
```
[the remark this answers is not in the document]
jaimehgb: It was because I initially made `Pool` and `PoolManager` to depend on an interface instead of of `Protocol` directly. And AssemblyScript doesn’t support defining fields in interfaces, only functions. So I had to make a getter instead.

But I think it is not necessary anymore, so I can remove it

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
jaimehgb: Sure, yeah. Better to be consistent with what we already have 👍

subgraphs/_reference_/src/sdk/util/numbers.ts
jaimehgb: Yes! This `util` folder is meant to be shared by all protocol types

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
jaimehgb: Moved them to the `util` folder

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
jaimehgb: I renamed it instead to `getAmountValueUSD`

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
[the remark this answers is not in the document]
jaimehgb: Oh, it’s actually in use by `account.ts`. I think I prefer to keep it, so it is `this.pool.GetBytesID()` instead of `this.pool.pool.id`

subgraphs/_reference_/src/sdk/protocols/bridge/constants.ts
```diff
@@ -0,0 +1,32 @@
...
+  export const MESSAGE_RECEIVED = "MESSAGE_RECEIVED";
+  export const OTHER = "OTHER";
+}
+export type TransactionType = string;
+
+export namespace CrosschainTokenType {
+  export const WRAPPED = "WRAPPED";
+  export const CANONICAL = "CANONICAL";
+}
+export type CrosschainTokenType = string;
```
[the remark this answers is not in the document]
jaimehgb: I renamed it to `enums`, otherwise it was getting messy and confusing

subgraphs/_reference_/src/sdk/protocols/bridge/README.md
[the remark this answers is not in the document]
tnkrxyz: Makes sense!

tnkrxyz (approved):
Looks good to me & I approve. Great work!
I will let Dylan handle his open conversations.

schema-bridge.graphql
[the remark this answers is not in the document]
dmelotik: Since we can do hyphens in the schema I think that is okay to keep in there. I like the look of it anyway

subgraphs/_reference_/src/sdk/protocols/bridge/pool_snapshot.ts
dmelotik: We should go with the google guide imo. So you are good, the ci will catch the file names as we edit those files

subgraphs/_reference_/src/sdk/protocols/bridge/protocol.ts
dmelotik: I see, resolved

dmelotik (approved):
Approved and added the file name schema to the issue here: https://github.com/messari/subgraphs/issues/1484

Issue #1499, opened 2022-12-06
Goldfinch Ethereum: Membership vault and reward & more


tnkrxyz:
Goldfinch recently deployed their membership vault that allows users to stake their output token (from senior pool and tranched pools) and GFI token to earn rewards: https://docs.goldfinch.finance/goldfinch/protocol-mechanics/membership

The subgraph in #1298 not yet include this. Will come back to implement this once merged.

tnkrxyz:
Another [open comment from @dmelotik](https://github.com/messari/subgraphs/pull/1298#discussion_r1018210259) re: `updatePosotion()` in helper.ts

https://github.com/tnkrxyz/subgraphs/blob/847f33e3f09faadedaa797734612007f414bcfb3/subgraphs/goldfinch/src/common/helpers.ts#L501-L503

```
  let openPosition = getOpenPosition(account.id, market.id, side);
  if (!openPosition) {
    openPosition = getNewPosition(account.id, market.id, side, event);
```

> Can a user not add onto an existing position? If that is the case this looks good, otherwise there is double counting

tnkrxyz:
More comments/TODOs [from Dylan](https://discord.com/channels/@me/984342328510652416/1050462217012785175):

> openPositionCount = cumulativePositionCount, this is likely not true
> And is there any other identifier we could append to the Market names?

dmelotik:
And Senior Pool rates ✅

ishraq8:
This is resolved here: https://github.com/messari/subgraphs/pull/1536

PR #1298, opened 2022-10-29
feat (#major); goldfinch; add goldfinch subgraph

tnkrxyz:
## TODO
- [x] README & Methodology
- [x] testing deployment
- [x] QA 
   - test deployment: https://api.thegraph.com/subgraphs/name/tnkrxyz/benqi (the url is correct & sorry have to re-use the handle)
   - validation dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/benqi&tab=protocol

PR #1536, opened 2022-12-19
feat (#patch); goldfinch; added membership rewards

tnkrxyz:
Added goldfinch membership rewards (staking of output ERC 720 tokens) #1499

- test deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/goldfinch-ethereum
- validation dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/goldfinch-ethereum

### Tasks
- [X] Membership vault / rewards
- [X] senior pool rates
- [ ] ~~[Add onto an existing position](https://github.com/messari/subgraphs/pull/1298#discussion_r1018210259)~~ (Dylan doesn't recall the exact concern here)
- [X] openPositionCount = cumulativePositionCount
- [ ] ~~additional identifier/names to add to Market names (tranched pool)~~ (confirmed with goldfinch folk, no tranched pool identifier available on chain)
- [x] remove compoundBalance field/code

### DIFF
--- a/subgraphs/goldfinch/src/common/log.ts
+++ b/subgraphs/goldfinch/src/common/log.ts
@@ -1,0 +1,297 @@
+import {
+  Address,
+  ethereum,
+  BigInt,
+  ByteArray,
+  crypto,
+  log,
+} from "@graphprotocol/graph-ts";
+import {
+  _MembershipStakingPosition,
+  _MembershipStakingTx,
+  _MembershipCapitalStaked,
+  _PoolToken,
+} from "../../generated/schema";
+import {
+  BIGINT_ZERO,
+  POOL_TOKENS_ADDRESS,
+  SENIOR_POOL_ADDRESS,
+  STAKING_REWARDS_ADDRESS,
+} from "./constants";
+
+const AdjustedHoldingsSig = crypto.keccak256(
+  ByteArray.fromUTF8("AdjustedHoldings(address,uint256,uint256)")
+);
+
+export class AdjustedHoldingsLog {
+  owner: Address;
+  eligibleAmount: BigInt;
+  nextEpochAmount: BigInt;
+
+  private constructor(txLog: ethereum.Log) {
+    this.owner = txLog.address;
+    this.eligibleAmount = AdjustedHoldingsLog.getEligibleAmount(txLog);
+    this.nextEpochAmount = AdjustedHoldingsLog.getNextEpochAmount(txLog);
+  }
+
+  static parse(txLog: ethereum.Log): AdjustedHoldingsLog | null {
+    if (!AdjustedHoldingsLog.isAdjustedHoldingsLog(txLog)) {
+      return null;
+    }
+
+    return new AdjustedHoldingsLog(txLog);
+  }
+
+  static isAdjustedHoldingsLog(txLog: ethereum.Log): boolean {
+    return txLog.topics[0].equals(AdjustedHoldingsSig);
+  }
+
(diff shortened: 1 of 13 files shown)


ishraq8:
@dmelotik Did the QA on this newer version. it is good, senior pool has accurate APR and the rewards are accurate too. Note that the rewards and interest on the site are APY. but if you calculate the amount emitted or interest due the APR we have is accurate.

subgraphs/goldfinch/src/common/log.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  static getAssetTokenId(txLog: ethereum.Log): BigInt {
+    const decoded = ethereum
+      .decode("(uint256,uint256,uint256)", txLog.data)!
+      .toTuple();
+    return decoded[1].toBigInt();
+  }
+
+  static getusdcEquivalent(txLog: ethereum.Log): BigInt {
```
dmelotik, line 121: should probably be named `getUsdcEquivalent()`?

subgraphs/goldfinch/src/common/log.ts
```diff
@@ -0,0 +1,297 @@
...
+      .decode("(uint256,address,uint256)", txLog.data)!
+      .toTuple();
+    return decoded[2].toBigInt();
+  }
+
+  setmarketId(marketId: string): void {
+    this.marketId = marketId;
+  }
+
+  setusdcEquivalent(usdcEquivalent: BigInt): void {
```
dmelotik, line 237: same?

subgraphs/goldfinch/src/mappings/senior_pool.ts
```diff
@@ -355,6 +355,9 @@ export function handleInvestmentMadeInSenior(
   market.totalBorrowBalanceUSD =
     market.totalBorrowBalanceUSD.plus(newBorrowUSD);
   market.cumulativeBorrowUSD = market.cumulativeBorrowUSD.plus(newBorrowUSD);
+  if (!market._interestTimestamp) {
```
dmelotik, line 358: It would save all these if not statements if you set the interest timestamp at market creation

subgraphs/goldfinch/src/mappings/tranched_pool.ts
```diff
@@ -95,6 +92,14 @@ export function handleDepositMade(event: DepositMade): void {
       .toHexString();
   }
 
+  if (!market._interestTimestamp) {
```
dmelotik, line 95: same here in this file too

dmelotik (commented):
Nice work on this!

Don't forget to update the version, there also seems to be an error in the subgraph.

subgraphs/goldfinch/src/mappings/senior_pool.ts
tnkrxyz: @dmelotik  Right! But there is a certain migrated tranched pool that does not have the market creation event, so I am adding this statement to handle it.

subgraphs/goldfinch/src/mappings/tranched_pool.ts
tnkrxyz: See comment above.

tnkrxyz:
> Nice work on this!
> 
> Don't forget to update the version, there also seems to be an error in the subgraph.

Thanks for catching these. Fixed.

dmelotik (approved):
approved

PR #1546, opened 2022-12-22
fix(#patch); dashboard; fix queries

dmelotik:
This was an accidental add, and was removed quietly. I am making this change to ensure the dashboard works.

Also, `RevenueDetails` was renamed to `RevenueDetail`


dmelotik:
@MichaelC1999 is making changes in a different PR

PR #1548, opened 2022-12-24
patch(): sdk; Fixed bridge sdk bug from schema change

jaimehgb:
We broke down that field in the schema into two other fields and didn’t get updated in the SDK.


tnkrxyz (approved):
Approved

PR #1552, opened 2022-12-31
fix (#patch); bridge sdk; fixed argument type for token

tnkrxyz:
In tokens.ts, the argument for `getOrCreateToken` and `getTokenParams` is specified as type `Address`, but Token.id is specified as `Bytes` in the schema. This PR changes the argument type to `Bytes`.

### DIFF
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -278,7 +278,9 @@ export class Account {
     updateMetrics: boolean = true
   ): LiquidityDeposit {
     const _pool = pool.pool;
-    const token = this.tokens.getOrCreateToken(_pool.inputToken);
+    const token = this.tokens.getOrCreateToken(
+      Address.fromBytes(_pool.inputToken)
+    );
 
     const deposit = new LiquidityDeposit(idFromEvent(this.event));
     deposit.hash = this.event.transaction.hash;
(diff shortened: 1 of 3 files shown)


jaimehgb:
Doesn’t it work as it is? Since Address inherits from Bytes.
In case it fails, instead of changing the type there I’d suggest changing it in `tokens.ts` where the `Token.load()` and `new Token()` are called, converting the Address into Bytes right where it is needed. 

This way we avoid having the client worrying about converting from one into another, and also prevent them from mistakenly using non-address bytes.

Screenshot 2023-01-02 at 12 13 20

tnkrxyz:
> Doesn’t it work as it is? Since Address inherits from Bytes. In case it fails, instead of changing the type there I’d suggest changing it in `tokens.ts` where the `Token.load()` and `new Token()` are called, converting the Address into Bytes right where it is needed.

It reported an error at building:
image

What you suggest makes sense. I reverted my change and converting Bytes to Address before passing to `getOrCreateToken`

tnkrxyz:
I realized you're already doing the conversion, for example, in [pool.ts](https://github.com/messari/subgraphs/blob/c331e206cdd4804b6c4f8aba1043ea578ac91337/subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts#L91). There are just two occurances in account.ts that you missed.

jaimehgb:
> I realized you're already doing the conversion, for example, in [pool.ts](https://github.com/messari/subgraphs/blob/c331e206cdd4804b6c4f8aba1043ea578ac91337/subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts#L91). There are just a few cases in account.ts that you missed.

Oh sounds good, my bad 👍

Issue #1079, opened 2022-09-27
Create custom format rules


dmelotik:
In an effort to normalize and create efficient workflows across messari/subgraphs. We want to have consistent formatting. This makes it easier to read, refactor, and add new code.

This effort is similar to the work being done on #1022 and can be looped into a similar set of tools.

dmelotik:
Tried implementing with no luck @bye43 . Thinking of canning this bc it is not worth any more effort

jaimehgb:
You almost had it @dmelotik 🙈 

Since we have prettier already installed at the root of the project (it’s on package.json), installing those dependencies and making it a npm script did the trick. When doing `npm install` npm then runs the post install scripts, which in this case would install husky. You can avoid that with `run: npm ci --ignore-scripts`, as we have in other jobs.

For the files changed, since we are already using something to determine which subgraphs to build based on the files that got changed we can reuse that too.


Seems to be working for me here:
https://github.com/messari/subgraphs/pull/1648
Unless I’m missing some edge case. What issues did you find?

PR #1559, opened 2023-01-04
feat(); actions; add format checker

dmelotik:
Check for format using prettier action: https://github.com/marketplace/actions/prettier-action

Currently: It will format on push when the action is run. Is that the behavior we want?

#1079

### DIFF
--- a/.github/workflows/pr-actions.yaml
+++ b/.github/workflows/pr-actions.yaml
@@ -57,11 +57,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Prettify code check
-        - uses: actions/checkout@v2
-        - uses: actionsx/prettier@v2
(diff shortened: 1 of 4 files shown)


jaimehgb:
IMO it is better to just check if it is formatted and fail if not, and let the whomever made the changes be aware of the requirement and then fix it manually. Specially if there are any quirks with assemblyscript that might break the build if some format is used. I think I found some of that with nested ternary operators or complex types or something.

Running `prettier —check` should do.

I also generally dislike getting my code changed automatically without my awareness 😅 But this is a personal thing.

dmelotik:
I agree, I was struggling to get it working the way you are suggesting though

dmelotik:
@jaimehgb  did this better

Issue #1293, opened 2022-10-28
Abracadabra Snapshots with 0 values


dmelotik:
Abracadabra snapshots have 0 values when there is no activity. To fix this we need to ensure that no cumulative values are set to 0 on creation, they should be initialized with the actual cumulative value.

See image:
![image](https://user-images.githubusercontent.com/60583747/198720868-1a7666bd-94b6-474f-be0d-dad08ab71de3.png)

MichaelC1999:
Abracadabra Avalanche marketDailySnapshots do not hold any data in the rate fields, regardless of which market

https://api.thegraph.com/subgraphs/name/messari/abracadabra-money-avalanche/graphql?query=%7B%0A++markets+%7B%0A++%09dailySnapshots+%28first%3A+1%29+%7B%0A++++++id%0A++++%09rates+%7B%0A++++++%09id%0A++++%09%7D%0A++++%7D%0A++++hourlySnapshots+%28first%3A+1%29+%7B%0A++++++rates+%7B%0A++++++++id%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D

bye43:
@dmelotik Is this still an issue?

dmelotik:
I added into the fix for #957

dmelotik:
This does not seem to be a subgraph issue anymore in the latest version of the subgraph: https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fid%2FQmRde24xjTq8NGP4QUX9sUVzFzD3YZ5WF8SV7X8QxQc4YM&tab=pool&poolId=0x4eaed76c3a388f4a841e9c765560bbe7b3e4b3a0&view=marketDailySnapshots-totalDepositBalanceUSD

@yulesa on your DA, with this endpoint (https://thegraph.com/hosted-service/subgraph/messari/abracadabra-money-ethereum) is it still an issue?

yulesa:
It is:
[...]

Issue #957, opened 2022-09-08
Abracadabra negative `openPositionCount`


dmelotik:
![image](https://user-images.githubusercontent.com/60583747/189042300-17f4a236-e78d-4b9a-a3a1-99bbd3bc700c.png)

Abracadabra has a negative open position early on in fantom's sync

bye43:
@dmelotik Is this fixed? Looks like its on [1.2.11](https://subgraphs.messari.io/subgraph?endpoint=messari/abracadabra-money-fantom&tab=pool&poolId=) now.

dmelotik:
No, I don't think it was ever fixed

bye43:
@ishraq8 Can you give this a quick look should be merged in.

ishraq8:
looks good, resolved.

PR #1564, opened 2023-01-04
fix(#patch); abra; fix negative positions

dmelotik:
#957 fix negative position counts
- I think positions were not being moved to closedPositions, yet market.openPositionCount was being decremented

#1293 
- Attempting to fix the spikes in iinput tokens. Not visible in the api though...
- Fixed empty market snapshot rates

Also fixed the abracadabra linting errors to get the actions to pass

Test: 
- https://okgraph.xyz/?q=dmelotik%2Fabracadabra-avalanche
- https://okgraph.xyz/?q=dmelotik%2Fabracadabra-arbitrum
- https://okgraph.xyz/?q=dmelotik%2Fabracadabra-bsc
- https://okgraph.xyz/?q=dmelotik%2Fabracadabra-fantom
- https://okgraph.xyz/?q=dmelotik%2Fabracadabra-ethereum

### DIFF
--- a/subgraphs/abracadabra/schema.graphql
+++ b/subgraphs/abracadabra/schema.graphql
@@ -878,14 +878,8 @@ type Account @entity {
   " All positions that belong to this account "
   positions: [Position!]! @derivedFrom(field: "account")
 
-  " All open positions this account has "
-  openPositions: [Position!]!
-
   " Number of open positions this account has "
   openPositionCount: Int!
-
-  " All closed positions this account has "
-  closedPositions: [Position!]!
 
   " Number of closed positions this account has "
   closedPositionCount: Int!
@@ -1256,6 +1250,14 @@ type LiquidateProxy @entity {
   " { Transaction hash }-{ Log index } "
   id: ID!
   amount: BigInt!
+}
+
+type PositionCounter @entity {
+  " { Account address }-{ Market address }-{ Position Side } "
+  id: ID!
+
+  " Next count "
+  nextCount: Int!
 }
 
 # Helper entity for calculating daily/hourly active users
--- a/subgraphs/abracadabra/src/positions.ts
+++ b/subgraphs/abracadabra/src/positions.ts
@@ -235,7 +235,7 @@ export function getLiquidatePosition(
 
 // A series of side effects on position added
 // They include:
-// * Create a new position when needed or reuse the exisitng position
+// * Create a new position when needed or reuse the existing position
 // * Update position related data in protocol, market, account
 // * Take position snapshot
 function addPosition(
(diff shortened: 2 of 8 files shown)


tnkrxyz (commented):
looks good so far

dmelotik:
@tnkrxyz this should be ready for review now, the pos counts are looking much better

subgraphs/abracadabra/schema.graphql
```diff
@@ -878,15 +878,9 @@ type Account @entity {
   " All positions that belong to this account "
   positions: [Position!]! @derivedFrom(field: "account")
 
-  " All open positions this account has "
-  openPositions: [Position!]!
```
jaimehgb, line 882: Was this part of the schema? Or was it added to debug?

subgraphs/abracadabra/src/positions.ts
```diff
@@ -388,3 +202,234 @@ function getAccountBalance(
...
+    .concat(side);
+  let positionCounter = PositionCounter.load(counterID);
+  if (!positionCounter) {
+    positionCounter = new PositionCounter(counterID);
+    positionCounter.nextCount = 0;
+    positionCounter.save();
+  }
+  const positionID = positionCounter.id
+    .concat("-")
+    .concat(positionCounter.nextCount.toString());
```
jaimehgb, line 263: Not a big deal, don’t want to delay this PR, but for the next one I’d suggest wrapping all this into its own function. Creating both IDs by concatenating values is something where it is easy to mess up. I think it is worth centralising in one or two functions.

subgraphs/abracadabra/src/positions.ts
```diff
@@ -388,3 +202,234 @@ function getAccountBalance(
...
+
+  let position = Position.load(positionID);
+  const openPosition = position == null;
+  if (openPosition) {
```
jaimehgb, line 267: Is there a reason or preference to do this instead of just `if (!position)`?

subgraphs/abracadabra/src/positions.ts
```diff
@@ -388,3 +202,234 @@ function getAccountBalance(
...
+  position = position!;
+  position.balance = newBalance;
+  if (eventType == EventType.DEPOSIT) {
+    position.depositCount += 1;
+  } else if (eventType == EventType.BORROW) {
+    position.borrowCount += 1;
+  }
+  position.save();
+
+  if (openPosition) {
```
jaimehgb, line 295: We’ve had this conversation before :D 

Try not to have the same condition evaluated in two different places. Centralize decision making.

If you think the branch of an `if` is too long because of too many things, then break it down into functions.

But evaluating the same condition multiple times can lead to nasty bugs. If at some point you need to fix something on the flow determined by that condition, it is possible that you need to modify both branches. If they are close to each other, it might be easy to catch, but the further away they are the more unlikely it will be. Not only within the same function or class, but eventually they could be in separate modules or separate services, being super hard to catch. Does this make sense?

Try to avoid it as much as possible :P

subgraphs/abracadabra/schema.graphql
dmelotik: This was an addon the abra dev did

subgraphs/abracadabra/src/positions.ts
dmelotik: Yeahhh,  I copied this logic from compound forks. I believe the one that fixes that issue is in the new lending schema

dmelotik: Fixed

subgraphs/abracadabra/src/positions.ts
jaimehgb: Oh, I didn’t mean to refactor it into the opposite case to do an early return. Was asking about 
```
const openPosition = position == null;
if (openPosition) {
 ...
}

// instead of just:
if (!position) { … }

```

Just curious about the redundancy. Is it more legible? Or feels safer when you know you are using a boolean? or? :D

jaimehgb (approved):

subgraphs/abracadabra/src/positions.ts
dmelotik: On the other instance you suggested we do an early return lmao, so that is what i did in the v3 lending sdk... I realized it was nicer looking to have the double conditions, now there are like 8 lines of duplicate code, but not a big deal

Either way we have to update code in 2 diff places if something happens 😵‍💫🤷‍♂️

PR #1574, opened 2023-01-06
feat(#minor); bridge sdk; add rewards.ts in sdk/util

dhruv-chauhan:


subgraphs/_reference_/src/sdk/util/rewards.ts
```diff
@@ -0,0 +1,292 @@
...
+    circularBuffer.save();
+  }
+
+  return circularBuffer;
+}
+
+function getStartingBlockRate(): BigDecimal {
+  // Block rates pulled from google searches - rough estimates
+
+  const network = dataSource.network().toUpperCase();
```
jaimehgb, line 255: You’ll need to replace the `_` with `-`, for the case of arbitrio-one at least

jaimehgb (approved):
Other than that, approved. Feel free to merge right after the fix

PR #1581, opened 2023-01-09
"fix(#patch); arbitrum network name"

itsjerryokolo:


jaimehgb (commented):
Dang, we are using different formatters. Can you run it through [prettier](https://prettier.io/)?

itsjerryokolo:
Okay, let me see if I can turn off the autoformatter.

subgraphs/_reference_/src/sdk/protocols/bridge/protocol.ts
```diff
@@ -68,7 +68,7 @@ export class Bridge {
     protocol = new BridgeProtocolSchema(id);
     protocol.name = conf.getName();
     protocol.slug = conf.getSlug();
-    protocol.network = dataSource.network().toUpperCase();
+    protocol.network = constants.Network.ARBITRUM_ONE;
```
jaimehgb, line 71: Wait, shouldn’t it be `dataSource.network().toUpperCase().replace(“-“, “_”)`? 🙈

jaimehgb: Oh, re-read the messages now. Understand the confusion 🙈 

Think about the library the same as any other code library you can find around. Right now we have to copy and paste it into each subgraph source code, which allows us to modify it as needed for all subgraphs. But ideally all changes to the lib should remain generic, since the same code will be used in other subgraphs. The contents of the lib shouldn’t change from one subgraph to another, they should all have the same.

So harcoding Arbitrum as the network for the protocol entity … doesn’t work 😅

itsjerryokolo: Okay, that makes sense. Thanks.
I guess if I was more aware, I would have figured out what you meant first time. 😅

jaimehgb (approved):

PR #1583, opened 2023-01-09
fix(#patch); bridge-sdk; update SDK api to also work with ethereum.call 

dhruv-chauhan:


subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -24,13 +24,17 @@ import {
...
+  sortBytesArray,
+  updateArrayAtIndex,
+} from "../../util/arrays";
 import { TokenManager } from "./tokens";
 import { PoolSnapshot } from "./poolSnapshot";
 import { SDK } from ".";
 
 type onCreatePoolCallback<T> = (
-  event: ethereum.Event,
+  event: ethereum.Block,
```
jaimehgb, line 37: The callback might need to use info from the event. Specially because AssemblyScript doesn’t allow closures. Can you make it `CustomEventType` instead so we don’t remove context?

subgraphs/_reference_/src/sdk/protocols/bridge/protocol.ts
```diff
@@ -146,10 +146,10 @@ export class Bridge {
 
   /**
    *
-   * @returns {ethereum.Event} the event currently being handled.
+   * @returns {ethereum.Block} the block currently being handled.
    */
-  getCurrentEvent(): ethereum.Event {
-    return this.event;
+  getCurrentBlock(): ethereum.Block {
```
jaimehgb, line 151: And then the same here, make it return `CustomEventType` for whomever needs to have the raw event.

jaimehgb (commented):
Thanks!

Just a comment on the event. Great otherwise :)

tnkrxyz:
An alternative approach I took is to create an "event" from call:
```
  const value = new ethereum.Value(ethereum.ValueKind.INT, 0 as u64);
  const eventParameters = [new ethereum.EventParam("value", value)];
  const event = new ethereum.Event(
    call.to,
    call.transaction.index,
    call.transaction.index,
    null,
    call.block,
    call.transaction,
    eventParameters,
    null
  );
```

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
dhruv-chauhan: makes a lot of sense, updated!

jaimehgb (approved):
🔥

PR #1595, opened 2023-01-11
feat(); Generic Schema SDK and Version 2.0.0; Add generic schema SDK

steegecs:
Context: 

Added Generic schema SDK using the Bridge SDK as a template. Additionally, I made updates to the schema, and added a new version of the schema to update some things the were growing dated.


subgraphs/_reference_/src/sdk/protocols/generic/index.ts
```diff
@@ -0,0 +1,147 @@
...
+import { ProtocolManager } from "./protocol";
+import { EventsManager } from "./event";
+import { PoolsManager } from "./pool";
+import { AccountsManager } from "./account";
+import { TokensManager, TokenInitializer } from "./token";
+import { TokenPricer } from "../config";
+import { Config } from "./config";
+import { BIGINT_ZERO, ZERO_ADDRESS } from "../../util/constants";
+
+export class CustomEventType {
```
jaimehgb, line 12: Make this type common for all the SDKs so we don’t re-implement it every time. You can then remove it from bridges too

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      BIGINT_ZERO,
+      BIGINT_ZERO,
+      Bytes.empty(),
+      BIGINT_ZERO
+    );
+    this.logIndex = BIGINT_ZERO;
+    this.event = null;
+  }
+
+  static initialize(
```
jaimehgb, line 51: I know you copied this as it was from bridges, but since we are here …

To reduce verbosity or redundant arguments, you can add 2 other methods like `initializeFromEvent` and `initializeFromCall` (or without the “initialize” part), to make it straightforward and simpler to call.

Alternatively, instead of requiring the SDK to pass a CustomEvent to the constructor, you can do the same thing there, having a couple static methods that serve as constructors: `SDK.initFromEvent()` or `SDK.initFromCall`. I prefer to be explicit over having an `initialize<T>` with a generic argument.

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      return new SDK(config, pricer, tokenInitializer, customEvent);
+    }
+    return new SDK(config, pricer, tokenInitializer, new CustomEventType());
+  }
+
+  create_pool() {
+    this.TokensManager;
+  }
+
+  update_balances(
```
jaimehgb, line 127: Do you want to expose this directly at the SDK level instead of at the PoolManager?

What will be the purpose of the PoolManager?

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.AccountsManager = new AccountsManager(this.ProtocolManager);
+    this.PoolsManager = new PoolsManager(this.ProtocolManager);
+    this.EventsManager = new EventsManager(
+      this.PoolsManager,
+      this.TokensManager
+    );
+    this.event = event;
+  }
+
+  create_pool() {
```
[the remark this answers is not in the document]
steegecs: yes!

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
steegecs: Added to utils!

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
steegecs: Okay - I added separate static methods.

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
steegecs: Updated this to be more like the bridges

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
```diff
@@ -85,29 +31,40 @@ export class SDK {
     this.Protocol.sdk = this;
   }
 
-  static initialize<T>(
```
jaimehgb, line 88: Being purists, we should leave this to not break old implementations that depend on it.

If you want to ensure that no one uses it in the future, and to enforce people to upgrade without breaking their build, you can add a deprecated note. The linter should warn you if you still use it:

https://jsdoc.app/tags-deprecated.html

steegecs: Where are you storing the version of the SDK so that I can reference it in the deprecated comment?

schema-generic.graphql
```diff
@@ -256,12 +289,15 @@ type Pool @entity {
...
+  liquidityToken: Token
+
+  " Type of token used to track liquidity "
+  liquidityTokenType: TokenType
+
   " Tokens that need to be deposited to take a position in protocol. e.g. WETH and USDC to deposit into the WETH-USDC pool. Array to account for multi-asset pools like Curve and Balancer "
   inputTokens: [Token!]!
 
-  " Token that is minted to track ownership of position in protocol "
-  outputToken: Token
```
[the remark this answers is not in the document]
steegecs: Messaged you about this.

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
jaimehgb: Nowhere yet. SDK versioning is something we should consider adding.
It is fine if you don’t mention a version number for now, a link to the PR might suffice

subgraphs/_reference_/src/sdk/protocols/generic/config.ts
```diff
@@ -0,0 +1,50 @@
+import { Versions } from "../../../../../../deployment/context/interface";
+
+import { ProtocolConfigurer } from "../config";
+import { BridgePermissionType } from "./enums";
+
+export interface Config extends ProtocolConfigurer {
+  getPermissionType(): BridgePermissionType;
+}
```
jaimehgb, line 8: I think you don’t need this?

subgraphs/_reference_/src/sdk/protocols/generic/config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class Config implements Config {
```
jaimehgb, line 10: Since the generic schema doesn’t require any custom config parameters you can reuse the class on `/sdk/protocols/config.ts`

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
```diff
@@ -0,0 +1,70 @@
+import { ethereum } from "@graphprotocol/graph-ts";
+
+import { AccountManager } from "./account";
+import { ProtocolManager } from "./protocol";
+import { PoolManager } from "./pool";
+import { TokenPricer } from "../config";
+import { Config } from "./config";
```
jaimehgb, line 7: And then here use the interface in that same `/sdk/protocols/config.ts`: `ProtocolConfigurer`

subgraphs/_reference_/src/sdk/protocols/generic/pool.ts
```diff
@@ -0,0 +1,387 @@
...
+  sortArrayByReference,
+  sortBytesArray,
+  updateArrayAtIndex,
+} from "../../util/arrays";
+import { TokenManager } from "./tokens";
+import { PoolSnapshot } from "./poolSnapshot";
+import { SDK } from ".";
+import { CustomEventType } from "../../util/events";
+
+type onCreatePoolCallback<T> = (
```
jaimehgb, line 20: I liked this approach of having a callback that runs only when a pool is created and to which you can pass custom arguments, but it has been very confusing for the people who have worked with it so far.

I’m inclined to remove it if we don’t find a way to make it more straightforward. What do you think?

subgraphs/_reference_/src/sdk/protocols/generic/pool.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  /**
+   * Utility function to convert some amount of input token to USD.
+   *
+   * @param amount the amount of inputToken to convert to USD
+   * @returns The converted amount.
+   */
+  getInputTokenAmountPrice(amount: BigInt): BigDecimal {
+    const token = this.getInputToken();
+    const price = this.protocol.getTokenPricer().getTokenPrice(token);
```
jaimehgb, line 161: Note to self: For some reason I added this same method to the TokenPricer interface. For convenience, to not have to do the calculation every time. And at the same time, I added it to the `Pool` too 😅 

I think we could remove the `getAmountValueUSD()` method from the TokenPricer interface. Since the SDK can calculate it from `getTokenPrice` alone, and this way we save the client from implementing it every time.

subgraphs/_reference_/src/sdk/protocols/generic/pool.ts
```diff
@@ -0,0 +1,387 @@
...
+    const pAmountUSD = pricer.getAmountValueUSD(inputToken, protocolSide);
+    const sAmountUSD = pricer.getAmountValueUSD(inputToken, supplySide);
+    this.addRevenueUSD(pAmountUSD, sAmountUSD);
+  }
+
+  /**
+   * Adds a given amount to the pool's mintSupply. It should only be used for pools of type BURN_MINT.
+   * @param amount {BigInt} The amount to add to the pool's mintSupply. It can be positive or negative.
+   */
+  addMintSupply(amount: BigInt): void {
```
jaimehgb, line 261: Do we have `mintSupply` for generic schema?

subgraphs/_reference_/src/sdk/protocols/generic/poolSnapshot.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    if (snapshotHourID != this.hourID) {
+      this.takeHourlySnapshot(snapshotHourID);
+      this.pool.lastSnapshotHourID = snapshotHourID;
+      this.pool.save();
+    }
+  }
+
+  private isInitialized(): boolean {
+    return !!(
```
jaimehgb, line 47: You don’t need the `!!` since the `&&` already returns a boolean.

(damn, seeing that I did the same on the bridges version, my bad)

Also, you’ll want to use the new fields you’ve added for this purpose? Since `_lastDailySnapshotTimestamp` and `_lastHourlySnapshotTimestamp` are not there

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
```diff
@@ -0,0 +1,250 @@
...
+  Address,
+  Bytes,
+  BigDecimal,
+} from "@graphprotocol/graph-ts";
+import { SDK } from ".";
+import { BIGINT_ZERO } from "../lending/constants";
+import { CustomEventType } from "../../util/events";
+
+/**
+ * Bridge is a wrapper around the BridgeProtocolSchema entity that takes care of
```
jaimehgb, line 20: “Bridge” :D 

You have a few references to Bridge stuff around

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (protocol) {
+      const proto = new ProtocolManager(protocol, pricer, event);
+      proto.setVersions(conf.getVersions());
+      return proto;
+    }
+
+    protocol = new ProtocolSchema(id);
+    protocol.name = conf.getName();
+    protocol.slug = conf.getSlug();
+    protocol.network = dataSource.network().toUpperCase().replace("-", "_");
```
jaimehgb, line 69: Should we move this into its own function? And then add something to also replace GNOSIS with XDAI?

subgraphs/_reference_/src/sdk/protocols/generic/protocolSnapshot.ts
```diff
@@ -0,0 +1,211 @@
...
+    snapshot.hourlyTransactionCount = transactionDelta;
+    snapshot.save();
+
+    activity.hourlyActiveUsers = 0;
+    activity.save();
+  }
+}
+
+function initActivityHelper(): _ActivityHelper {
+  let helper = _ActivityHelper.load(ActivityHelperID);
```
jaimehgb, line 201: You are missing this entity in the schema

jaimehgb (commented):
🔥

schema-generic.graphql
```diff
@@ -303,6 +339,15 @@ type Pool @entity {
...
 
+  " Day ID of the most recent daily snapshot "
+  lastSnapshotDayID: Int!
+
+  " Hour ID of the most recent hourly snapshot "
+  lastSnapshotHourID: Int!
+
+  " Timestamp of the last time this entity was updated "
+  lastUpdateTimestamp: BigInt!
+
```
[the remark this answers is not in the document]
steegecs: We could prepend the `-`. I did not do it, because usually those we use for supplementary fields. However, I think we want this to be a field in the schema for all subgraphs, since it should be used commonly. I'd imagine if a protocol is implementing the new schema, we'd also want them to use the new methodology for snapshots, meanly these will always be used, no? I made them required, because if we want to standardize the snapshotting mechanism, I thought this would encourage it. Also, I thought it might potentially be interesting for users to see these fields on the main entity

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
steegecs: Added link

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
```diff
@@ -85,6 +31,46 @@ export class SDK {
...
+    );
+    return new SDK(config, pricer, tokenInitializer, customEvent);
+  }
+
+  static intializeFromCustomEvent(
+    config: BridgeConfigurer,
+    pricer: TokenPricer,
+    tokenInitializer: TokenInitializer
+  ): SDK {
+    return new SDK(config, pricer, tokenInitializer, new CustomEventType());
```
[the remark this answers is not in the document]
steegecs: Ah my mistake, I believe I put in the proper fix.

subgraphs/_reference_/src/sdk/protocols/generic/config.ts
steegecs: Ah yes, cleaned up.

subgraphs/_reference_/src/sdk/protocols/generic/config.ts
steegecs: Yup - Updated to use the base config class.

subgraphs/_reference_/src/sdk/protocols/generic/index.ts
steegecs: Use the interface or the class?

steegecs: Got it. Updated!

subgraphs/_reference_/src/sdk/protocols/generic/pool.ts
steegecs: We do not. Removed.

subgraphs/_reference_/src/sdk/protocols/generic/poolSnapshot.ts
steegecs: Yes! Updated this. All good. I no catch either.

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
steegecs: Fixed!

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
steegecs: Actually, if you use `gnosis` to specify network in the subgraph.yaml, then `dataSource.network()` will return `gnosis`. Do you think we need a function for it since this is the case?

subgraphs/_reference_/src/sdk/protocols/generic/protocolSnapshot.ts
steegecs: Added this entity and the circular buffer!

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
jaimehgb: Hmm, but then the BE will break if it expects `xdai` and you use `gnosis`. No?

I’m thinking, if at a subgraph level we mess up accidentally we can ensure it still works.

jaimehgb: Even if we don’t replace the gnosis part, I’d still put it in its own function. It does no harm and I’ve seen a few places where we use `datasource.network()` to match with our network enum. I think having it in a central place will work as a reference to refer to when attempting to get the current network. Plus, removing duplicates, which I’ve seen and done myself wrong a few times

schema-generic.graphql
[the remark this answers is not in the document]
jaimehgb: Hmm 🤔 

The encouraging part, to keep snapshot methods consistent, sells me. I would have said that having them immutable already achieves that, but this might help also keep the same intervals and stuff. Sounds good 👍

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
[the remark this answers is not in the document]
jaimehgb: Same thing on `generic/index.ts` :D

subgraphs/_reference_/src/sdk/protocols/generic/protocol.ts
steegecs: Yeah that makes sense. I agree then with putting the transformation in a common function. I think, in a way though, it makes sense not to do the conversion. When I was messing with this, the graph-cli caught that I was setting the network value to the protocol that was not in the enum. If we leave it as is, the usage of `xdai` be caught and fixed.

jaimehgb: Ah right, that makes sense too. Unless someone fixes it by adding it back to the schema :D 
But we might catch that too on the review if anything

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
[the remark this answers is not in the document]
steegecs: Screen Shot 2023-01-26 at 1 57 02 AM

Oh wait, I did mean to put that there. I was following your lead with this in the Bridge schema. I guess its for chains that do not have a normal ethereum.Event/Call data type?

jaimehgb (approved):

jaimehgb (approved):

subgraphs/_reference_/src/sdk/protocols/bridge/index.ts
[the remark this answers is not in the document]
jaimehgb: Interesting. In this case we can panic, do some `log.critical` if that type is not considered, since it is a misconfiguration and not something that will come up in the middle of the sync process

PR #1610, opened 2023-01-16
feat (); bridge sdk; Added account methods messageOut and messageIn

tnkrxyz:
Added `messageOut` and `messageIn` method for Account and necessary helper functions to handle messages sent/received across the bridge managed by the bridge protocol.

Note that there are protocol.cumulativeUniqueMessageSenders, UsageMetric.cumulativeUniqueMessageSenders and daily/hourlyActiveMessageSenders, but not protocol.cumulativeUniqueMessageReceivers or UsageMetric.daily/hourlyActiveMessageReceivers in the schema. As [jaimehgb explains](https://discord.com/channels/953684103012683796/961658223444787220/1063578350985150545), that is because receiving a message is not done by an account per se, but by a contract.

### DIFF
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -285,7 +285,7 @@ export class Account {
       srcChainId,
       this.protocol.getCurrentChainID(),
       source,
-      this.account.id,
+      Address.fromBytes(this.account.id),
       false,
       data
     );
@@ -310,7 +310,7 @@ export class Account {
     return this.message(
       this.protocol.getCurrentChainID(),
       dstChainId,
-      this.account.id,
+      Address.fromBytes(this.account.id),
       destination,
       true,
       data
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -7,6 +7,7 @@ import {
   LiquidityWithdraw,
   PoolRoute,
   ActiveAccount,
+  BridgeMessage,
 } from "../../../../generated/schema";
 import { Pool } from "./pool";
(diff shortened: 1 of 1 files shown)


subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
```diff
@@ -267,6 +268,83 @@ export class Account {
     return transfer;
   }
 
+  /**
+   * Creates a BridgeMessage entity for a message arriving to this chain
+   *
+   * @param srcChainId the source chain id
+   * @param source The account sending the message.
```
jaimehgb, line 275: I think the `source` should be `Address | null`, since it might not be possible to always know where it is coming from? WDYT?

jaimehgb (approved):
Other than that, good!

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
tnkrxyz: Since both `from` and `to` are required fields for [BridgeMessage entity](https://github.com/messari/subgraphs/blob/9be3f84272faf4e0d397e4e36bb25267ecd06be2/schema-bridge.graphql#L1224), I think it makes sense to let the subgraph developer choose a value for `source` (zero address or receiver, or else), instead of the sdk making assumptions. What do you think? @jaimehgb

jaimehgb: Oh, we can change the schema too :D 

But zero address for whenever is unknown might work too.

jaimehgb (approved):

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
tnkrxyz: I slightly prefer to keep the current implementation and let subgraph developer choose the value for unavailable information.

PR #1615, opened 2023-01-17
fix(#patch); truefi; fix versions

dmelotik:
- Fix truefi schema version to prevent "protocol out of sync" error on PM


steegecs (approved):
LGTM

Issue #1617, opened 2023-01-17
Abracadabra Ethereum: Indexing Error


bye43:
https://okgraph.xyz/?q=messari%2Fabracadabra-money-ethereum

bye43:
@dmelotik Is this issue closed?

PR #1618, opened 2023-01-17
fix(#patch); abra; fix holey array

dmelotik:
Fixing holey array issue in abra: #1617

- Build using different structure
- Testing: https://okgraph.xyz/?q=dmelotik%2Fabracadabra-ethereum

### DIFF
--- a/subgraphs/abracadabra/src/common/tokens.ts
+++ b/subgraphs/abracadabra/src/common/tokens.ts
@@ -24,6 +24,12 @@ export function fetchTokenSymbol(tokenAddress: Address): string {
     // for broken pairs that have no symbol function exposed
     if (!isNullEthValue(symbolResultBytes.value.toHexString())) {
       symbolValue = symbolResultBytes.value.toString();
+    } else {
+      // try with the static definition
+      let staticTokenDefinition = fromAddress(tokenAddress);
+      if (staticTokenDefinition != null) {
+        symbolValue = staticTokenDefinition.symbol;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


subgraphs/abracadabra/src/common/tokens.ts
```diff
@@ -76,108 +76,103 @@ export function fetchTokenDecimals(tokenAddress: Address): i32 {
...
-  } else {
-    return INVALID_TOKEN_DECIMALS as i32;
   }
+
+  return INVALID_TOKEN_DECIMALS as i32;
 }
 
 export function isNullEthValue(value: string): boolean {
-  return value == "0x0000000000000000000000000000000000000000000000000000000000000001";
+  if (value == "0x0000000000000000000000000000000000000001" || value == "0x0000000000000000000000000000000000000000")
```
jaimehgb, line 88: Why `….0001` is considered null?

jaimehgb (commented):
What’s exactly the fix? 😅

subgraphs/abracadabra/src/common/tokens.ts
dmelotik: It's a burn address: https://etherscan.io/address/0x0000000000000000000000000000000000000001

dmelotik:
@jaimehgb 

> What’s exactly the fix?
redoing the static class, it seems to be causing a holey array issue, and I added an extra check for burn addresses

jaimehgb (approved):

dmelotik:
@jaimehgb we should probably get down to the bottom of this. I have never known what is actually wrong when I get the holey array error. Are you aware?

dmelotik:
Aha, thanks @jaimehgb

PR #1622, opened 2023-01-18
feat(); Gnosis Decentralized Network; Add decentralized network deployments to the deployment.json

steegecs:


dmelotik:
Why do we need the decentralized network part if it is not deployed?

dmelotik:
I imagine the BE will try and query that endpoint `UNKOWN` too

dmelotik:
If we want this to be a common pattern we prolly want to make `UNKNOWN` something the BE recognizes as no deployment, skip

steegecs:
> If we want this to be a common pattern we prolly want to make `UNKNOWN` something the BE recognizes as no deployment, skip

My bad. I meant to have this PR on draft. Just had not deployed and pushed the ids yet, but they are updated now.

dmelotik (approved):
I will forgive you for not marking this a draft @steegecs . lfgtm

PR #1626, opened 2023-01-19
chore(); remove comp v3 decentralized network

dmelotik:
Since our id's use Bytes and the Byte < > comparisons are only avail on the hosted service right now we need to avoid trying to query the decentralized network.

We will add this back in once the changes are persisted on the decentralized network.


lowjiansheng (approved):
LGTM :)

Issue #1598, opened 2023-01-12
#bug; incorrect Pangolin V2 Avalanche fees


TKYAngus:
Link to subgraph: https://api.thegraph.com/subgraphs/name/messari/pangolin-avalanche
Example with the liqudityPool id=0xbd918ed441767fe7924e99f6a0e0b568ac1970d9
```json
        "fees": [
          {
            "id": "0xbd918ed441767fe7924e99f6a0e0b568ac1970d9-lp-fee",
            "feePercentage": "0.3",
            "feeType": "FIXED_LP_FEE"
          },
          {
            "id": "0xbd918ed441767fe7924e99f6a0e0b568ac1970d9-protocol-fee",
            "feePercentage": "0",
            "feeType": "FIXED_PROTOCOL_FEE"
          },
          {
            "id": "0xbd918ed441767fe7924e99f6a0e0b568ac1970d9-trading-fee",
            "feePercentage": "0.3",
            "feeType": "FIXED_TRADING_FEE"
          }
        ]
```

However from the doc (https://docs.pangolin.exchange/multichain/avalanche-network/png-token/litepaper#protocol-fees-and-token-buybacks), on Avalanche trader pays a 0.3% fee that is paid out to 
- Liquidity Pools: 0.25% (in accordance with their share in the particular pool)
- PNG staking Pool: 0.0425%
- Pangolin DAO’s treasury: 0.0075%

bye43:
Addressing here on PR: https://github.com/messari/subgraphs/pull/1632

PR #1632, opened 2023-01-20
fix (#patch); pangolin-avalanche; fixed fees

tnkrxyz:
This PR fixes fee issue reported in #1598 by editing the relevant value in configurations.ts

- Test deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/pangolin-avalanche
- data validation: https://subgraphs.messari.io/subgraph?endpoint=https://thegraph.com/hosted-service/subgraph/tnkrxyz/pangolin-avalanche&tab=protocol
- relevant query: https://api.thegraph.com/subgraphs/name/tnkrxyz/pangolin-avalanche/graphql?query=query+MyQuery+%7B%0A++liquidityPools%28where%3A+%7Bid%3A+%220xbd918ed441767fe7924e99f6a0e0b568ac1970d9%22%7D%29+%7B%0A++++fees+%7B%0A++++++feePercentage%0A++++++feeType%0A++++++id%0A++++%7D%0A++%7D%0A%7D

### DIFF
--- a/subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
+++ b/subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
@@ -1,8 +1,9 @@
 ////////////////////
 //////Versions//////
 ////////////////////
+import { Versions } from "../../../../src/versions";
 
-export const PROTOCOL_SUBGRAPH_VERSION = "1.0.4";
-export const PROTOCOL_METHODOLOGY_VERSION = "1.0.0";
+export const PROTOCOL_SUBGRAPH_VERSION = Versions.getSubgraphVersion();
+export const PROTOCOL_METHODOLOGY_VERSION = Versions.getMethodologyVersion();
 export const PROTOCOL_NAME = "Pangolin";
 export const PROTOCOL_SLUG = "pangolin";
--- a/subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
+++ b/subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
@@ -2,7 +2,7 @@
 //////Versions//////
 ////////////////////
 
-export const PROTOCOL_SUBGRAPH_VERSION = "1.0.0";
+export const PROTOCOL_SUBGRAPH_VERSION = "1.0.4";
 export const PROTOCOL_METHODOLOGY_VERSION = "1.0.0";
 export const PROTOCOL_NAME = "Pangolin";
 export const PROTOCOL_SLUG = "pangolin";
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -5859,7 +5859,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.0.3",
+          "subgraph": "1.0.4",
           "methodology": "1.0.0"
         },
         "services": {
--- a/subgraphs/uniswap-forks/abis/pangolin/Factory.json
+++ b/subgraphs/uniswap-forks/abis/pangolin/Factory.json
@@ -189,5 +189,20 @@
     "payable": false,
     "stateMutability": "nonpayable",
     "type": "function"
+  },
+  {
+    "constant": true,
+    "inputs": [],
+    "name": "feeTo",
+    "outputs": [
+      {
+        "internalType": "address",
+        "name": "",
+        "type": "address"
+      }
+    ],
+    "payable": false,
+    "stateMutability": "view",
+    "type": "function"
   }
 ]
--- a/subgraphs/uniswap-forks/configurations/configure.ts
+++ b/subgraphs/uniswap-forks/configurations/configure.ts
@@ -3,7 +3,7 @@ import { getNetworkConfigurations } from "./configurations/configurations";
 import { Deploy } from "./configurations/deploy";
 
 // Select the deployment protocol and network
-let deployment = Deploy.HONEYSWAP_POLYGON;
+const deployment = Deploy.PANGOLIN_AVALANCHE;
 
 // export const NetworkConfigs = configurationsMap.get(deployment)!
 export const NetworkConfigs = getNetworkConfigurations(deployment);
--- a/subgraphs/uniswap-forks/protocols/apeswap/src/common/constants.ts
+++ b/subgraphs/uniswap-forks/protocols/apeswap/src/common/constants.ts
@@ -2,7 +2,9 @@
 //////Versions//////
 ////////////////////
 
-export const PROTOCOL_SUBGRAPH_VERSION = "1.1.7";
-export const PROTOCOL_METHODOLOGY_VERSION = "1.0.0";
+import { Versions } from "../../../../src/versions";
+
+export const PROTOCOL_SUBGRAPH_VERSION = Versions.getSubgraphVersion();
+export const PROTOCOL_METHODOLOGY_VERSION = Versions.getMethodologyVersion();
 export const PROTOCOL_NAME = "ApeSwap";
 export const PROTOCOL_SLUG = "apeswap";
--- a/subgraphs/uniswap-forks/protocols/biswap/src/common/constants.ts
+++ b/subgraphs/uniswap-forks/protocols/biswap/src/common/constants.ts
@@ -1,8 +1,9 @@
 ////////////////////
 //////Versions//////
 ////////////////////
+import { Versions } from "../../../../src/versions";
 
-export const PROTOCOL_SUBGRAPH_VERSION = "1.0.0";
-export const PROTOCOL_METHODOLOGY_VERSION = "1.0.0";
+export const PROTOCOL_SUBGRAPH_VERSION = Versions.getSubgraphVersion();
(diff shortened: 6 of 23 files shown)


bye43:
@tnkrxyz Took a quick look at this [query ](https://api.thegraph.com/subgraphs/id/QmZX5nMJWDN1jCFBohD7tZTcDVotGLvYtYzppmVamf4nxu/graphql?query=query+MyQuery+%7B%0A++liquidityPools+%7B%0A++++fees+%7B%0A++++++feePercentage%0A++++++feeType%0A++++++id%0A++++%7D%0A++%7D%0A%7D) and I think the protocol fee should be 0.05% since we count the fee that goes to PNG stakers (0.0425%), as a part of the protocol fees.

tnkrxyz:
> 



> @tnkrxyz Took a quick look at this [query ](https://api.thegraph.com/subgraphs/id/QmZX5nMJWDN1jCFBohD7tZTcDVotGLvYtYzppmVamf4nxu/graphql?query=query+MyQuery+%7B%0A++liquidityPools+%7B%0A++++fees+%7B%0A++++++feePercentage%0A++++++feeType%0A++++++id%0A++++%7D%0A++%7D%0A%7D) and I think the protocol fee should be 0.05% since we count the fee that goes to PNG stakers (0.0425%), as a part of the protocol fees. Also, I think both these fees can get updated dynamically, so just want to make sure that we are tracking them in the contract and not hard coding them since I think thats why we missed this change last time. Thanks!

Good question! This is something I struggled a bit myself - it seems to me the 0.0425% doesn't go to the protocols, but stakers of PNG (can be anyone staking PNG: swap users, LP etc). Maybe I should create a new fee type FIXED_STAKER_FEE?

You're right the distribution of the fee can be updated in the contract. The current implementation doesn't consider this and will have to be added.

bye43:
> > 
> 
> > @tnkrxyz Took a quick look at this [query ](https://api.thegraph.com/subgraphs/id/QmZX5nMJWDN1jCFBohD7tZTcDVotGLvYtYzppmVamf4nxu/graphql?query=query+MyQuery+%7B%0A++liquidityPools+%7B%0A++++fees+%7B%0A++++++feePercentage%0A++++++feeType%0A++++++id%0A++++%7D%0A++%7D%0A%7D) and I think the protocol fee should be 0.05% since we count the fee that goes to PNG stakers (0.0425%), as a part of the protocol fees. Also, I think both these fees can get updated dynamically, so just want to make sure that we are tracking them in the contract and not hard coding them since I think thats why we missed this change last time. Thanks!
> 
> Good question! This is something I struggled a bit myself - it seems to me the 0.0425% doesn't go to the protocols, but stakers of PNG (can be anyone staking PNG: swap users, LP etc). Maybe I should create a new fee type FIXED_STAKER_FEE?
> 
> You're right the distribution of the fee can be updated in the contract. The current implementation doesn't consider this and will have to be added.

Hmm - I do think that it would be nice to have a separate `FIXED_STAKER_FEE`, but would we have to change the schema? IIRC we have been counting the staker portion of the fee as part of the protocol fee for other protocols (I think an example is MKR), since token holders are similar to traditional share holders.

tnkrxyz:
> > > @tnkrxyz Took a quick look at this [query ](https://api.thegraph.com/subgraphs/id/QmZX5nMJWDN1jCFBohD7tZTcDVotGLvYtYzppmVamf4nxu/graphql?query=query+MyQuery+%7B%0A++liquidityPools+%7B%0A++++fees+%7B%0A++++++feePercentage%0A++++++feeType%0A++++++id%0A++++%7D%0A++%7D%0A%7D) and I think the protocol fee should be 0.05% since we count the fee that goes to PNG stakers (0.0425%), as a part of the protocol fees. Also, I think both these fees can get updated dynamically, so just want to make sure that we are tracking them in the contract and not hard coding them since I think thats why we missed this change last time. Thanks!
> > 
> > 
> > Good question! This is something I struggled a bit myself - it seems to me the 0.0425% doesn't go to the protocols, but stakers of PNG (can be anyone staking PNG: swap users, LP etc). Maybe I should create a new fee type FIXED_STAKER_FEE?
> > You're right the distribution of the fee can be updated in the contract. The current implementation doesn't consider this and will have to be added.
> 
> Hmm - I do think that it would be nice to have a separate `FIXED_STAKER_FEE`, but would we have to change the schema? IIRC we have been counting the staker portion of the fee as part of the protocol fee for other protocols (I think an example is MKR), since token holders are similar to traditional share holders.

Yes, this would require changing the schema. It's a good point that token holders are like share holders; I will count the 0.0425% as the protocol share for now.

tnkrxyz:
@bye43 , OK, I fixed allocation of the 0.0425%. I also make the Fee On/Off switch function depending on feeTo value. Early on, the protocol fee is set to 0 (feeTo = 0x0). 

However, so far I haven't found a good way to determine the fee and split using on-chain data. The 0.3% fee seem to be hard coded (https://github.com/pangolindex/exchange-contracts/blob/ca7e9d971ea31d745cfc1250a6120ce16bd1538b/contracts/pangolin-core/PangolinPair.sol#L88). The split of fees (between protocol and supplier) is only coded in [this PangolinRouterSupportingFees contract](https://github.com/pangolindex/exchange-contracts/blob/main/contracts/pangolin-periphery/PangolinRouterSupportingFees.sol), but it doesn't seem to be the contract deployed: https://docs.pangolin.exchange/multichain/avalanche-network/contracts . I am not sure whether it is worthwhile to keep digging.

bye43:
> @bye43 , OK, I fixed allocation of the 0.0425%. I also make the Fee On/Off switch function depending on feeTo value. Early on, the protocol fee is set to 0 (feeTo = 0x0).
> 
> However, so far I haven't found a good way to determine the fee and split using on-chain data. The 0.3% fee seem to be hard coded (https://github.com/pangolindex/exchange-contracts/blob/ca7e9d971ea31d745cfc1250a6120ce16bd1538b/contracts/pangolin-core/PangolinPair.sol#L88). The split of fees (between protocol and supplier) is only coded in [this PangolinRouterSupportingFees contract](https://github.com/pangolindex/exchange-contracts/blob/main/contracts/pangolin-periphery/PangolinRouterSupportingFees.sol), but it doesn't seem to be the contract deployed: https://docs.pangolin.exchange/multichain/avalanche-network/contracts . I am not sure whether it is worthwhile to keep digging.

Thanks! And gotcha, if its not easy to implement, then no worries ... it is probably not worth the time as long as we have the correct block for when the new fees were added in then we should be good to go!

jaimehgb:
> The split of fees (between protocol and supplier) is only coded in [this PangolinRouterSupportingFees contract](https://github.com/pangolindex/exchange-contracts/blob/main/contracts/pangolin-periphery/PangolinRouterSupportingFees.sol), but it doesn't seem to be the contract deployed:

I’ve been taking a quick look to try to find their contract where they do the fee distribution but with no luck. Let’s assume that it is the same distribution as in the GH contract

subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
```diff
@@ -2,7 +2,7 @@
 //////Versions//////
 ////////////////////
 
-export const PROTOCOL_SUBGRAPH_VERSION = "1.0.0";
+export const PROTOCOL_SUBGRAPH_VERSION = "1.0.4";
```
jaimehgb, line 5: Since you are at it, do you mind removing this constant and using instead versions.ts?

jaimehgb (approved):

subgraphs/uniswap-forks/protocols/pangolin/src/common/constants.ts
tnkrxyz: Good idea! done.

tnkrxyz:
> > The split of fees (between protocol and supplier) is only coded in [this PangolinRouterSupportingFees contract](https://github.com/pangolindex/exchange-contracts/blob/main/contracts/pangolin-periphery/PangolinRouterSupportingFees.sol), but it doesn't seem to be the contract deployed:
> 
> I’ve been taking a quick look to try to find their contract where they do the fee distribution but with no luck. Let’s assume that it is the same distribution as in the GH contract

That's my conclusion too. Good that you confirm it.

jaimehgb (approved):

bye43:
@dmelotik Can you deploy this to the messari hosted? Thanks!

dmelotik:
Deployed @bye43 https://okgraph.xyz/?q=messari%2Fpangolin-avalanche.

@tnkrxyz is this something that we can graft? It will probably take quite a while to sync. Or do you have a deployment that is further ahead that I can graft to messari's hosted service?

tnkrxyz:
> Deployed @bye43 https://okgraph.xyz/?q=messari%2Fpangolin-avalanche.
> 
> @tnkrxyz is this something that we can graft? It will probably take quite a while to sync. Or do you have a deployment that is further ahead that I can graft to messari's hosted service?

@dmelotik Unfortunately, no :( . My deployment (https://thegraph.com/hosted-service/subgraph/tnkrxyz/pangolin-avalanche) is still syncing and since this change affects the revenue split from the very beginning, we cannot draft from the current version messari/pangolin-avalanch, which has incorrect split.

PR #1636, opened 2023-01-23
fix(#patch); Uniswap V3; Fix account tracking in events

steegecs:
**Context**
Needed to switch the `account` field in swaps, deposits, and withdraws to use event.transaction.from, so that the field can properly reference account entity.


subgraphs/uniswap-v3/src/common/entities/token.ts
```diff
@@ -54,8 +59,13 @@ export function getOrCreateToken(
     token.save();
   }
 
-  if (token.lastPriceBlockNumber != event.block.number && getNewPrice) {
-    token.lastPriceUSD = findUSDPricePerToken(event, token);
+  if (
+    event.block.number.minus(token.lastPriceBlockNumber!).gt(BIGINT_TEN) &&
```
jaimehgb, line 63: Not against this change, but could it happen that by leaving a 10 block margin between price updates we might be skipping blocks where we had price issues? Potentially making it look like the price controls you added work, when in reality we just skipped that block?

Have you validated that they work without this? :P

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,108 @@ export function findUSDPricePerToken(
...
-              NetworkConfigs.getMinimumLiquidityThreshold()
-            )
-          ) {
-            largestWhitelistTokenValue = whitelistTokenValueLocked;
-            // token0 per our token * whitelist token per token0
-            priceSoFar = poolAmounts.tokenPrices[0].times(
-              token0.lastPriceUSD as BigDecimal
-            );
+
+        // log.warning("HELLO", [])
```
jaimehgb, line 162: Hi 👋

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,108 @@ export function findUSDPricePerToken(
...
+
+        // whitelist token is token1
```
jaimehgb, line 164: This comment is not accurate anymore since the index can be either 1 or 0?

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,108 @@ export function findUSDPricePerToken(
...
+        ].times(whitelistToken.lastPriceUSD!);
+
+        // Check if it meets criteria to update prices.
+        if (
+          whitelistTokenValueLocked.gt(largestWhitelistTokenValue) &&
+          whitelistTokenValueLocked.gt(
+            NetworkConfigs.getMinimumLiquidityThreshold()
+          )
+        ) {
+          // Calculate new price of a token and TVL of token in this pool.
```
jaimehgb, line 182: A rule I usually follow to avoid functions to become too long and to hopefully increase readability: when I find myself commenting what something does in the middle of the body of a function, I wrap that next block of code into its own well named function. Even if that is the only place where it is going to be used.

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: lol

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: Correct! Removed.

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-arbitrum/configurations.ts
```diff
@@ -67,7 +67,19 @@ export class UniswapV3ArbitrumConfigurations implements Configurations {
...
+      // "0x8b794cdd5a89b05f65886d4afddc7152d0a1633a", // Fake WETH
+      // "0x2807cdee12db6b043cc7ac4ddaec88c89871d299", // Core DAO
+      // "0x0e017687ce1971fc7672246de6675112de137730", // Flare
+      // "0x9799b59e133d95a8f145be345f25e8ee97815ce0", // Poison.Finance Poison
+      // "0x1ca6e7eb4ba052ac6cb53e2516061f88821c1ec2", // Canto
+      // "0x59767fda87eaf896caf7b149d12f3c7287697a9e", // VitaDAO
+      // "0x7f91531fc25dd262aebf57e8ebe9a6a6df372e96", // WoofWork1
+      // "0x90959fc17635f6b62bf5a9bdaee849f0f54ec2c4", // WoofWork2
+      // "0x6f2446d795e0807e828d5707871590c4abe84627", // Rocket Pool
+    ]);
```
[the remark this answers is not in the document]
steegecs: Removing these.

subgraphs/uniswap-v3/src/common/entities/token.ts
steegecs: Not as thoroughly as I have with it. I can try that out. This definitely could happen, but on this, I think more than likely it'll still catch those instances. This is because there are usually multiple blocks or many blocks between pricings on these smaller pools.

steegecs: I changed this to try to improve performance, not sure its helping too much though.

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,108 @@ export function findUSDPricePerToken(
...
+          const newPriceSoFar = poolAmounts.tokenPrices[
+            whitelistTokenIndex
+          ].times(whitelistToken.lastPriceUSD as BigDecimal);
```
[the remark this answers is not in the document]
steegecs: Yes this is correct. Its because it is deriving this price from the sqrtx96 price. It's a bit of a misleading name I see, but it essentially describes the price of a token in terms of the other.

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-ethereum/configurations.ts
```diff
@@ -81,42 +81,42 @@ export class UniswapV3MainnetConfigurations implements Configurations {
   }
   getUntrackedPairs(): Bytes[] {
     return stringToBytesList([
-      "0x8fe8d9bb8eeba3ed688069c3d6b556c9ca258248", // Mulan/USDT
-      "0xfd9715a9f03678975b2e224c938fae8d481b09b2", // Aurora/wETH
-      "0x9663f2ca0454accad3e094448ea6f77443880454", // LUSD/wETH
-      "0x785d7ff551c9b62562c6767551ffa3e3e562b6ed", // SiFum/USDT
+      // "0x8fe8d9bb8eeba3ed688069c3d6b556c9ca258248", // Mulan/USDT
```
dmelotik, line 84: same deal 👀

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-optimism/configurations.ts
```diff
@@ -69,8 +69,8 @@ export class UniswapV3OptimismConfigurations implements Configurations {
   }
   getUntrackedTokens(): Bytes[] {
     return stringToBytesList([
-      "0x296f55f8fb28e498b858d0bcda06d955b2cb3f97", // stargatetoken
-      "0x2e3d870790dc77a83dd1d18184acc7439a53f475", // frax
+      // "0x296f55f8fb28e498b858d0bcda06d955b2cb3f97", // stargatetoken
```
dmelotik, line 72: 👀

subgraphs/uniswap-v3/src/common/dex_event_handler.ts
```diff
@@ -582,6 +576,20 @@ export class DexEventHandler {
     this.pool.save();
   }
 
+  updateAndSavePoolTokenEntities(): void {
+    for (let i = 0; i < this.poolTokens.length; i++) {
+      const poolToken = this.poolTokens[i];
```
dmelotik, line 581: I think this array will automatically get sorted since is an array of entities: https://github.com/messari/subgraphs/blob/master/docs/ERRORS.md#array-sorting-when-querying

Do we know poolTokens and inputTokenBalanceDeltas are parallel?

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-ethereum/configurations.ts
dmelotik: oops, is this the buffer?

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,105 @@ export function findUSDPricePerToken(
...
+  );
+  const protocolTVLPercentageDelta = absBigDecimal(
+    safeDiv(tokenTVLDelta, protocol.totalValueLockedUSD)
+  );
+  if (protocolTVLPercentageDelta.gt(BIGDECIMAL_FIVE_PERCENT)) {
+    log.warning("Price too high for token: {} from pool: {}", [
+      token.id.toHexString(),
+      priceSoFar.toString(),
+    ]);
+    token._largeTVLImpactBuffer += 1;
```
dmelotik, line 221: I like this, should we also add a count for the number of times a price is skipped bc it is over 10bb?

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,105 @@ export function findUSDPricePerToken(
...
+    return token.lastPriceUSD!;
+  }
+
+  // Uncomment this after the current production version is fully synced and re-deploy to make the oracle more robust
```
dmelotik, line 225: nothing is commented?

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -139,46 +144,105 @@ export function findUSDPricePerToken(
...
+      token._largePriceChangeBuffer = token._largePriceChangeBuffer + 1;
+      return token.lastPriceUSD!;
+    }
+  }
+
+  token._largePriceChangeBuffer = 0;
+  token._largeTVLImpactBuffer = 0;
+
+  token.save();
   return priceSoFar; // nothing was found return 0
```
dmelotik, line 246: couldn't this be >0 now?

subgraphs/uniswap-v3/src/common/utils/utils.ts
```diff
@@ -286,3 +287,17 @@ export function stringToBytesList(list: string[]): Bytes[] {
   }
   return result;
 }
+
+export function absBigInt(value: BigInt): BigInt {
```
dmelotik, line 291: there is actually a built in abs() func for BigInt, it doesn't look like that is the case for BD though

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-ethereum/configurations.ts
steegecs: This was the blacklist. I went ahead and removed these tokens.

subgraphs/uniswap-v3/protocols/uniswap-v3/config/deployments/uniswap-v3-optimism/configurations.ts
steegecs: rm

subgraphs/uniswap-v3/src/common/dex_event_handler.ts
steegecs: Yes, this will be the case, I always and handling the tokens in the object with the same ordering on the pool, which matches the ordering from the contracts (amount0, amount1)

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: I decided not to do that, because there is never an instance where a token TVL has been close, and if it does go over, its most certainly wrong.

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: Removed!

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: Yeah - that was a dumb comment, it should always be > 0

subgraphs/uniswap-v3/src/common/utils/utils.ts
steegecs: Oh right. I removed it.

subgraphs/uniswap-v3/src/common/price/price.ts
steegecs: Yeah, that is a good point. I broke it down a little bit with two helper functions.

subgraphs/uniswap-v3/src/common/price/price.ts
```diff
@@ -144,27 +144,16 @@ export function findUSDPricePerToken(
...
-          whitelistTokenIndex = 1;
-          tokenIndex = 0;
-        } else if (pool.inputTokens[1] == token.id) {
-          whitelistTokenIndex = 0;
-          tokenIndex = 1;
-        }
-
-        // Return if token is not in pool
-        if (tokenIndex == -1 || whitelistTokenIndex == -1) {
+        const token_index = get_token_index(pool, token);
```
jaimehgb, line 147: ❤️

jaimehgb (approved):

Issue #1379, opened 2022-11-07
Rethink subgraph versioning


dmelotik:
Something @jaimehgb brought up was to rethink our  versioning process:

tl;dr
- it should be based on deployment expectations
- a #patch doesn't need to be redeployed
- a #minor could benefit from redeployment, but not critical
- a #major is critical that redeployment occurs
- This will be easier to know the status of current subgraphs and priority

Personally i am for this. Use this issue as a discussion board

steegecs:
Nice! Thanks for putting up the issue.

steegecs:
I like the structure as this is.

bye43:
Also like the structure. To @steegecs point, I think maybe it would be good to think of minor and major in terms of protocol metrics/backfilling. For example, Uniswap V3 I would say `cumulativeUniqueUsers` is a minor change since we are not ingesting user metrics, but it would major when we do start displaying them. Obviously, this classification will not work all the time, but might be useful for initial framing. My other question is how is versioning going to relate to the versions of subgraphs we see (i.e. 1.2.1, etc). I think I generally am pretty lost when those versions change and what they signify.

jaimehgb:
>  For example, Uniswap V3 I would say cumulativeUniqueUsers is a minor change since we are not ingesting user metrics, but it would major when we do start displaying them.

This is a good point. How do we know if a past minor change should be promoted to major? Feels like every minor change is a future major. If not, it should be a patch.

Maybe we can have some rule to not allow minor mismatches to persist for longer than X amount of time?

bye43:
This looks good to me! Thanks @jaimehgb

tnkrxyz:
Looks good to me too. I want to point out there may still be some subjective judgement involved in deciding between major and minor version, because we may not want to apply the 20% threshold on a few data points (e.g. in a hourly snapshot), on less critical/stable metrics (hypothetically difference in hourly active users, if any, will likely be large).

dmelotik:
Second @tnkrxyz - I wouldn't put a number on it, but it is more of a judgement thing. TBH a major is probably something bad enough that we want to remove the protocol from PM until it is fixed

dmelotik:
https://github.com/messari/subgraphs/pull/1659

PR #1645, opened 2023-01-27
fix(#minor); comp v3; fix addPosition

dmelotik:
New positions do not update the balance or position event count. Fixed this

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1841,7 +1841,7 @@
         "status": "prod",
         "versions": {
           "schema": "3.0.0",
-          "subgraph": "1.0.5",
+          "subgraph": "1.1.0",
           "methodology": "1.0.0"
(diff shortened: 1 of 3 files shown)


jaimehgb (approved):

dmelotik:
Do you think this is a minor change @jaimehgb and when should we redeploy to the decentralized network?

jaimehgb:
> Do you think this is a minor change @jaimehgb and when should we redeploy to the decentralized network?

If we follow what we said [here](https://github.com/messari/subgraphs/issues/1379#issuecomment-1375623592) I think a minor is correct. If it wasn’t in the context of subgraphs I’d have made it a patch.

For minors and patches, unless they are requested to be sooner, we can maybe leave a buffer and batch them together with other deployments. Say once a week or so?

PR #1648, opened 2023-01-27
feat; actions: Validate formatting

jaimehgb:
This PR adds an additional action to the CI to validate that subgraph files are formatted with `prettier`. If they are not, it will fail. Enforcing contributors to push the code already formatted.


dmelotik:
Does it work?! @jaimehgb

dmelotik (approved):
nice

PR #1651, opened 2023-01-27
fix(#major); comp v3; ETH pricing

dmelotik:
With the addition of the new ethereum market the prices are returned in ETH, unfortunately there is nothing on chain identifying this so we need to hardcode in logic to get the ETH price to get the real TVL.

Test: https://okgraph.xyz/?q=dmelotik%2Fcompound-v3-ethereum


dmelotik:
wrong branch

PR #1654, opened 2023-01-29
Add Event Interface to Generic Schema

Data-Nexus:
Added an Event Interface to the Generic Schema.

Protocols typically have a set of events that would useful to display to users to see an activity feed for the protocol, a market or their account. Each protocol may have varying events, but they can be standardly queried using an Event interface and using __typename to return the event type.

Below is a sample query would accomplish this: 
```
query UserEventLog($skip: Int = 0, $first: Int = 10, $account: String = "0x0xabc123") {
  events(first: $first, skip: $skip, where: {account: $account}) {
    hash
    timestamp
    account {
      id
    }
    amount
    market {
      id
      name
    }
    __typename
  }
}
```

### DIFF
--- a/schema-generic.graphql
+++ b/schema-generic.graphql
@@ -479,6 +479,59 @@ type PoolHourlySnapshot @entity {
   blockNumber: BigInt!
 }
 
+    ##### Event Data #####
+    ##### Different Protocols will emit varying events #####
+    ##### The Event interface can be used for querying an activity feed #####
+
+interface Event {
+  " { Transaction hash }-{ Log index } "
+  id: ID!
+
+  " Transaction hash of the transaction that emitted this event "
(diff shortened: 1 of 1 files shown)


dmelotik:
@Data-Nexus is this an example or are you proposing this gets added to the generic schema?

Data-Nexus:
I am proposing that we add this as a standard feature in the schema. Each protocol will vary in the events they emit, but as developers start planning how they're going to tackle their mapping they should incorporate this in their building plan.

dmelotik:
@Data-Nexus, this will be our best place to keep track of this as we go through issues regularly https://github.com/messari/subgraphs/issues/1658

dmelotik:
I am going to close with the previous comment in mind.

Issue #1519, opened 2022-12-14
Upgrade AAVE to 3.0.0 Schema


bye43:
Issues to address when we do this: 

### Tasks
- [x] Upgrade aave forks to latest lending schema/SDK
- [x] Add liquidation Rev to Aave forks: https://github.com/messari/subgraphs/issues/1831
- [x] Ensure all helper fields are prepended with `_` before deployment: https://github.com/messari/subgraphs/issues/1084
- [x] Ensure no immutable entity IDs are changed (protocol.id, market.id) (note: this is case sensitive)
- [x] Simplify Aave V3 (and forks) (see spark lend manifest for what to do)
- [x] add flashloan data
- [x] Ensure positions are properly labeled with isolation type in aave v3: https://github.com/messari/subgraphs/issues/1491
- [x] Will need to include this during upgrade: https://github.com/messari/subgraphs/pull/1669 
    - [need clarification](https://github.com/messari/subgraphs/pull/1669#issuecomment-1510016903)
- [ ] ~~(optional) treasury data: https://github.com/messari/subgraphs/issues/163~~
     - [need more specifics](https://github.com/messari/subgraphs/issues/163#issuecomment-1510002233)
     - [decided to not do this atm](https://discord.com/channels/@me/1091205155921338448/1097901589102080071)

dmelotik:
Will also allow us to get aave flashloan data

dmelotik:
Another note: aave v3 and forks can be refactored to be simpler (see spark lend). We might as well do that change in this PR as well

ishraq8:
@dmelotik just to confirm, this will not effect endpoints for protocol metrics right?

dmelotik:
Good point @ishraq8. We need to discuss how we can go about this from a consumption standpoint. 

To answer the question, this affects endpoints. The upgrade creates breaking changes on the endpoint. But, our service already supports the schema we will be moving to in this issue.

The question is, what work needs to be done in order to support this update, if any?

tnkrxyz:
I am almost done with the refactoring. @dmelotik , @bye43 , one quesiton for you is whether the upgraded subgraphs need to be QAed? Since the upgrade/refactoring adds new logics and changes existing logic (transfers), I feel more comfortable if any protocol specialist can help take a look.

ishraq8:
@tnkrxyz happy to take a look for QA, I think it is best practice to QA these refactoring. Give me a heads up when they are ready.

dmelotik:
[...]

PR #1669, opened 2023-02-01
fix(#patch); aave v3; handle change to BalanceTransfer event

defispartan:
The `value` parameter of the `BalanceTransfer` event was updated in the [Aave 3.0.1 contract update](https://github.com/aave/aave-v3-core/pull/682) to be the scaled balance in place of current balance. 

This PR interprets `value` based on current block compared to V3 -> V3.0.1 contract upgrade for that market.

Once Aave V3 markets are upgraded to V3.0.1 via Governance, the update blocks will need to be added to this helper function and subgraphs re-deployed.

### DIFF
--- a/subgraphs/aave-v3/protocols/aave-v3/src/utils/constants.ts
+++ b/subgraphs/aave-v3/protocols/aave-v3/src/utils/constants.ts
@@ -5,3 +5,14 @@ export namespace PositionSide {
   export const LENDER = "LENDER";
   export const BORROWER = "BORROWER";
 }
+
+// returns block of market update to v3.0.1, or null if no upgrade
+export function getUpdateBlock(network: string): u32 {
+  let updateBlock = -1;
+  if (network === 'mainnet') {
+    updateBlock = 0;
+  }
+  // TODO: Add upgrade block numbers as they occur
+
+  return updateBlock;
+}
--- a/subgraphs/aave-v3/protocols/aave-v3/src/mappings/AToken.ts
+++ b/subgraphs/aave-v3/protocols/aave-v3/src/mappings/AToken.ts
@@ -1,4 +1,4 @@
-import { Address } from "@graphprotocol/graph-ts";
+import { Address, dataSource } from "@graphprotocol/graph-ts";
 import {
   AToken,
   BalanceTransfer,
@@ -17,6 +17,7 @@ import { updateUserLenderPosition } from "../entities/position";
 import { amountInUSD } from "../entities/price";
 import { getReserve, updateReserveATokenSupply } from "../entities/reserve";
 import { getTokenById } from "../entities/token";
+import { getUpdateBlock } from "../utils/constants";
 
 export function handleBurn(event: Burn): void {
   const contract = AToken.bind(event.address);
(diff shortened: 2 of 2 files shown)


dmelotik:
Hey @defispartan thanks for this! Do you know when we can expect this to be deployed to mainnet so we can update the blocknumber accordingly?

defispartan:
Mainnet is already live, the block number is just set to 0 since this market will only have the updated events and the old logic will never be used. This subgraph should good to deploy with no other changes.

All of the other V3 markets will have a block number entry once they're upgraded.

subgraphs/aave-v3/protocols/aave-v3/src/utils/constants.ts
```diff
@@ -5,3 +5,14 @@ export namespace PositionSide {
   export const LENDER = "LENDER";
   export const BORROWER = "BORROWER";
 }
+
+// returns block of market update to v3.0.1, or null if no upgrade
+export function getUpdateBlock(network: string): u32 {
+  let updateBlock = -1;
+  if (network === 'mainnet') {
```
dmelotik, line 12: Can you use the global constant from constants.ts for the network name

dmelotik (commented):
We had a change to aave v3 that we made recently so you will need to rebase.

One of the changes is that we do not actually calculate the account balances from `transfer.balance` event param anymore. We make a contract call to get the new balances of the sender/receiver. My question is: does the `balanceOf()` function in the AToken contract use this logic? From that PR https://github.com/aave/aave-v3-core/pull/682 it looks like no

dmelotik:
This will be necessary to do once we upgrade `aave-forks` to the 3.0.0 schema and we are keeping track of Transfers. #1519

tnkrxyz:
@defispartan , can you help clarify about the aToken upgrade?

It seems to me that the aToken contract has identical signature for BalanceTransfer since January, this is [the first aToken](https://etherscan.io/address/0x7effd7b47bfd17e52fb7559d3f924201b9dbff3d#code#F17#L227) created on Jan 27 according to ReserveInitialized events emitted by [the PoolConfigurator contract](https://etherscan.io/address/0x64b761D848206f447Fe2dd461b0c635Ec39EbB27#events), while [this one](https://etherscan.io/address/0x21714092d90c7265f52fdfdae068ec11a23c6248#code#F18#L227) the last one on Apr 10th. I see that the value argument is divided by index before BalanceTransfer event emitting on both contracts.

Are you saying that we should start to multiplying `event.params.index` with ``event.params.value` on later aToken deployment, but not on earlier ones? Which block did the switch/upgrade happen? and how can we tell? thx!

tnkrxyz:
@dmelotik , appreciate your input too!

dmelotik:
@tnkrxyz to me it seems like this change only affected the value from `BalanceTransfer` event. And the current aave v3 code doesn't need that event.

I don't think we even need to account for this then. Do you agree? And is there a reason we would need to add `BalanceTransfer` back in??

tnkrxyz:
> @tnkrxyz to me it seems like this change only affected the value from `BalanceTransfer` event. And the current aave v3 code doesn't need that event.
> 
> I don't think we even need to account for this then. Do you agree? And is there a reason we would need to add `BalanceTransfer` back in??

@dmelotik From what I found so far, we should switch to use BalanceTransfer for aave-v2 [because the Transfer event gives transfer amount that has been discounted by liquidity index](https://github.com/aave/protocol-v2/blob/ce53c4a8c8620125063168620eba0a8a92854eb8/contracts/protocol/tokenization/AToken.sol#L384-L390). It is trickier for v3, because they flipped the amount argument between Transfer event and BalanceTransfer in middle of aave-v3 deployment (see my description above), so my current implementation is get amount from both BalanceTransfer and Transfer event, and use the large number of the two. Let me know if this makes sense.

tnkrxyz:
#1993 implements the BalanceTransfer/Transfer logic for aave-v3. We can close this issue once #1993 is merged.

dmelotik:
Closing, this was tackled in #1993 Thanks all!

PR #1677, opened 2023-02-02
Updates to qidao readme docs

ishraq8:
Updated revenue on qidao docs to reflect subgraphs. Also removed rewards calculation as that is done offchain.


dmelotik (approved):
lgtm, merging

PR #1682, opened 2023-02-03
fix(#minor); aave forks; fix liquidations

dmelotik:
Liquidations were not properly mapped. Fixed:
- `liquidate.asset` is now the asset that is borrowed, not the collateral asset
- `liquidate.profitUSD` is now the difference between the amount seized and the amount of debt paid by the liquidator.

Test deployment: https://okgraph.xyz/?q=dmelotik%2Faave-v2-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -130,7 +130,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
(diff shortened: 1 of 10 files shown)


tnkrxyz (approved):
Looks good. approved.

PR #1683, opened 2023-02-03
feat(#minor): Fix Perp schema problem happening during compilation and deployment

corerouter:
Some small changes to Perp schema in order to compile and deploy related GMX subgraph.
1. Some changes are needed as the following errors happening when compiling, deploying and running subgraphs otherwise.
error2
error4

2. Change some id from Bytes to String as it contains some string such as Deposit
3. Change some entries from immutable:true to immutable:false as the mapping codes may need to update the snapshot more than once.

### DIFF
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -1591,7 +1591,7 @@ type Position @entity {
 }
 
 type PositionSnapshot @entity(immutable: true) {
-  id: String!
+  id: Bytes!
 
   "Position this snapshot refers to"
   position: Position!
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -1518,7 +1518,7 @@ type Account @entity {
 # Helper entity for calculating daily/hourly active users
 type ActiveAccount @entity(immutable: true) {
   " { daily/hourly }-{ Address of the account }-{ Days/hours since Unix epoch } "
-  id: String!
+  id: Bytes!
 }
 
 enum PositionSide {
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -69,8 +69,8 @@ enum RewardTokenType {
 }
 
 type RewardToken @entity(immutable: true) {
-  " { Reward token type }-{ Smart contract address of the reward token } "
-  id: String!
+  " Bytes.fromI32(0|1).concat(rewardTokenAddress), where 0 is for deposit and 1 is for borrow "
+  id: Bytes!
 
   " Reference to the actual token "
   token: Token!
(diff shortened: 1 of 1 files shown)


schema-derivatives-perpfutures.graphql
```diff
@@ -1440,32 +1508,37 @@ type Account @entity{
...
+
+  " Position opened as short "
+  SHORT
 }
 
 # A position is defined as who has what position in the protocol and distribution of longs and shorts at a positional level
 type Position @entity {
   " { Account address }-{ Market address }-{ Position Side }-{ Optional: Interest Rate Type}-{ Counter } "
-  id: Bytes!
+  id: String!
```
jaimehgb, line 1535: Keep IDs as `Bytes`. If we need to concat various values to get the ID you can still do so with Bytes. We don’t need the dashes, so it is all addresses and numbers. (Interes Rate Type doesn’t apply in this case, I think we can also remove from the comment)

schema-derivatives-perpfutures.graphql
```diff
@@ -487,13 +492,11 @@ type UsageMetricsDailySnapshot @entity (immutable: true) {
...
   " Cumulative number of collateral Out from traders "
   cumulativeCollateralOut: Int!
-
   " Total number of pools "
   totalPoolCount: Int!
-
 }
 
-type UsageMetricsHourlySnapshot @entity (immutable: true) {
+type UsageMetricsHourlySnapshot @entity(immutable: false) {
```
jaimehgb, line 499: Instead of making them mutable, you can have an auxiliary entity that holds these counters. And then when taking the snapshot, copy it over from there and reset it.

And same for the daily ones

schema-derivatives-perpfutures.graphql
```diff
@@ -63,14 +63,14 @@ enum RewardTokenType {
...
+
   " For reward tokens awarded to stakers "
   STAKE
 }
 
-type RewardToken @entity (immutable: true) {
+type RewardToken @entity(immutable: true) {
   " { Reward token type }-{ Smart contract address of the reward token } "
-  id: Bytes!
+  id: String!
```
jaimehgb, line 73: Take a look at this one: https://github.com/messari/subgraphs/blob/master/schema-bridge.graphql#L172

We can keep it as bytes, and consistent with bridges

jaimehgb:
Thanks for the fixes! Left some comments on the immutable ones and IDs as strings. All the rest looks correct to me! 💪

corerouter:
> Thanks for the fixes! Left some comments on the immutable ones and IDs as strings. All the rest looks correct to me! 💪

All fixed at 3864986b411c56a979d1bde345c6a0abe72c0402 Thanks!

schema-derivatives-perpfutures.graphql
corerouter: fixed at 3864986b411c56a979d1bde345c6a0abe72c0402 Thanks!

schema-derivatives-perpfutures.graphql
corerouter: fixed at 3864986b411c56a979d1bde345c6a0abe72c0402 Thanks!

schema-derivatives-perpfutures.graphql
corerouter: fixed at 3864986b411c56a979d1bde345c6a0abe72c0402 Thanks!

jaimehgb (approved):

PR #1684, opened 2023-02-03
feat(#minor): Price Lib; Added support for Celo, Fuse & Cronos + Optimizations

harsh9200:
- Added Return Oracle Type (helpful in debugging wrong token prices)
- Oracle doesn't run if the contract is not deployed/built yet. (Optimizations)
- Added support for Cronos, Celo & Fuse chains.


steegecs (approved):
LGTM

PR #1687, opened 2023-02-03
fix(#patch): multichain; add missing pool destination tokens

jaimehgb:
https://okgraph.xyz/?q=QmSoawPaM5fWFiSMH7yaGDwNprXXTwGoV3cv5JsebojWhn

`Pool.destinationTokens` is showing empty for multichain. Since destination tokens are linked to pool routes, whenever a new pool route is added to a pool, we’ll add the token too if not there already.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -6160,7 +6160,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.0.0",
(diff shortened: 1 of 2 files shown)


jaimehgb:
Build might fail because of too many ABIs to generate code from

dmelotik (approved):
lgtm!

PR #1691, opened 2023-02-04
feat(#major); arbitrum-one-bridge; add arbitrum-one-bridge subgraph

fortysevenlabs:
Deployment Links

- https://thegraph.com/hosted-service/subgraph/fortysevenlabs/arbitrum-one-bridge-ethereum
- https://thegraph.com/hosted-service/subgraph/fortysevenlabs/arbitrum-one-bridge-arbitrum


fortysevenlabs:
@this-username-is-taken @jaimehgb ready for review, can you assign a reviewer?

will do a final cleanup/reorganize after the review is complete.

fortysevenlabs:
Majority of the errors seem to be `Ethereum addresses should be lowercase  rulesdir/no-checksum-addresses` coming from `prices` and `sdk`. Has this been fixed in prices lib as well?

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
```diff
@@ -0,0 +1,83 @@
...
+  const crossToken = sdk.Tokens.getOrCreateCrosschainToken(
+    networkToChainID(Network.ARBITRUM_ONE),
+    crossTokenAddress!,
+    CrosschainTokenType.CANONICAL,
+    crossTokenAddress!
+  );
+
+  // -- POOL
+
+  const poolId = event.address;
```
jaimehgb, line 59: Does this address handle multiple ERC20 tokens? Or there is a contract per each token?

In case a single address manages multiple tokens, you’ll need to have them as separate pools in the schema. Since each Pool in the schema only handles one token. You would have to break down the real pool into multiple “virtual” ones.

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
```diff
@@ -0,0 +1,76 @@
...
+  const acc = sdk.Accounts.loadAccount(event.params.sender);
+
+  // -- HANDLE ETH DEPOSIT & MESSAGES
+
+  // Nitro
+  // Message Types - https://github.com/OffchainLabs/nitro/blob/master/contracts/src/libraries/MessageTypes.sol#L10
+  // L1MessageType_ethDeposit = 12
+  // L2_MSG = 3
+
+  if (event.params.kind == 12) {
```
jaimehgb, line 31: Can you put this `12` into a named constant? Same for the `3` in the next if branch.

Also, are the rest of message types safe to ignore for our purpose?

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
```diff
@@ -0,0 +1,76 @@
...
+      event.params.sender,
+      event.transaction.value,
+      event.transaction.hash
+    );
+  } else if (event.params.kind == 3) {
+    acc.messageIn(
+      networkToChainID(Network.ARBITRUM_ONE),
+      event.params.sender,
+      event.params.messageDataHash
+      // event.transaction.hash   // bug: no optional transactionID param
```
jaimehgb, line 73: What do you mean?

jaimehgb:
Do you have a deployment link?

jaimehgb:
> Majority of the errors seem to be `Ethereum addresses should be lowercase rulesdir/no-checksum-addresses` coming from `prices` and `sdk`. Has this been fixed in prices lib as well?

Fixed that here: https://github.com/messari/subgraphs/pull/1692

After merge, feel free to copy over the SDK and price lib files again

fortysevenlabs:
Added deployment links so you can go ahead with further review. Will address your comments tomorrow.

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
fortysevenlabs: Ah, didn’t realize we want to approach it that way but makes sense. I’ll break the pools down. Is there a naming convention for the pools in such case? (`address-token` work?).

@jaimehgb  Slightly related, how should the protocols be setup? Right now, I’m create a single protocol (or rather the sdk is) on arbitrum and ethereum side. The address I’m using for the protocol is a generic address (because there isn’t a single factory contract or equivalent that we can use). Do these make sense?

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
fortysevenlabs: Will do for the named constants.

> Also, are the rest of message types safe to ignore for our purpose?
I think so but I’ll review this again. Here’s the full list of [L2 messages](https://github.com/OffchainLabs/nitro/blob/master/contracts/src/libraries/MessageTypes.sol#L10).

@jaimehgb  Can you define again what kind of message types should we consider for tracking under the bridge? all types including ones required for L2 operations vs generic types vs deposit money, etc.

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
fortysevenlabs: transactionID param was documented for`messageIn` but it’s actually not part of the function. we need to remove this @param documentation. See @params and method paramus.

https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts#L274-L292

fortysevenlabs:
Resolved most comments. Should be able to close out soon.

Pending tasks:
- [ ]  Revisit message types and evaluate if we want to add anything
- [ ]  Pull newer prices-lib and sdk (holding off because of a few bugs)

@jaimehgb Created issues for the bugs to provide visibility for the issue (so it can be fixed everywhere).
- https://github.com/messari/subgraphs/issues/1694
- https://github.com/messari/subgraphs/issues/1696

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
fortysevenlabs: @jaimehgb I believe we should ignore the other message types because they are related to protocol operations or sending transactions vs messages (which are tracked using `L2_MSG` event and are sent by invoking `sendL2Message()`, `sendL2MessageFromOrigin()`). 

See [Arbitrum’s Inbox]( https://github.com/OffchainLabs/nitro/blob/master/contracts/src/bridge/Inbox.sol) for more details.

fortysevenlabs:
@jaimehgb I’m aware of the solution you’ve outlined [https://github.com/messari/subgraphs/issues/1694#issuecomment-1420705290]. I created the bug because (I didn’t have the bandwidth and) wanted to ensure this fix would ensure others don’t have to debug/deal with this. Are we going to go ahead with the changes in sdk?

I’ll go ahead and resolve the issue for this subgraph though (again to save time related to PR/merge).

fortysevenlabs:
@jaimehgb
* good to go from your side?
* can we assign somebody for QAing the deployments?

bye43:
> @jaimehgb
> 
> * good to go from your side?
> * can we assign somebody for QAing the deployments?

Ya - will send that off. @fortysevenlabs You can ping me for all the QA related stuff.

fortysevenlabs:
@bye43 sounds good. what's the QA status on this?

@jaimehgb good to go from PR perspective?

bye43:
> @bye43 sounds good. what's the QA status on this?
Hopefully, should be up by EOW

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
jaimehgb: Answering this, although already outdated, but for context:

As long as each pool ID is unique, it is fine. We don’t rely on IDs for anything else, since it is not always possible to make them unique from the contract address alone.

Same for the protocol. If you have a factory contract from which all the rest come, you can use that one. If not, you can use the deployer address. But same, the ID is not used for anything other than uniquely identifying entities, so anything goes.

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
jaimehgb: Makes sense 👍

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/messagesAndEthDeposits.ts
jaimehgb: Ah dang, good catch. Will look into that

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export function handleTransferIn3pGateway(event: DepositFinalized): void {
+  log.error("[3p Gateway] We are in transferIn3pGateway", []);
+
+  const l1Token = new ethereum.EventParam("l1token", event.parameters[0].value);
+  const _from = new ethereum.EventParam("_from", event.parameters[1].value);
+  const _to = new ethereum.EventParam("_to", event.parameters[2].value);
+  const _amount = new ethereum.EventParam("_amount", event.parameters[3].value);
+
+  const params: ethereum.EventParam[] = [l1Token, _from, _to, _amount];
```
jaimehgb, line 23: Oh snap, didn’t it work passing the params as they came? :(

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
```diff
@@ -0,0 +1,97 @@
...
+    log.info("calculate cross token address call reverted", []);
+  } else {
+    crossTokenAddress = crossTokenAddressResult.value;
+  }
+
+  const crossToken = sdk.Tokens.getOrCreateCrosschainToken(
+    networkToChainID(Network.ARBITRUM_ONE),
+    crossTokenAddress!,
+    CrosschainTokenType.CANONICAL,
+    crossTokenAddress!
```
jaimehgb, line 68: Just noticing that this is not well documented, but the first address is the one of the crosschain token, and the second one is the address of the token on the current chain, or what you call `inputTokenAddress`.

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/transferOut.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  // -- TOKENS
+
+  const crossTokenAddress: Address = event.params.l1Token;
+
+  const crossToken = sdk.Tokens.getOrCreateCrosschainToken(
+    networkToChainID(Network.ARBITRUM_ONE),
+    crossTokenAddress!,
+    CrosschainTokenType.CANONICAL,
+    crossTokenAddress!
```
jaimehgb, line 67: Same here. Unless you don’t have the other one available

jaimehgb (approved):
I left two tiny things. I’m sorry I took this long to review 😓 

Also! I’m assuming that no ad-hoc changes were made to the SDK, right? If we upgraded it and just replaced it, would everything still work?

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transferIn.ts
fortysevenlabs: Ah, I was wondering about that but forgot diving into the underlying method. We should rename the params to be intuitive in that case (eg. inputTokenAddress, crossTokenAddress).

I'll go push the change out to pass the inputTokenAddress for the second one.

subgraphs/arbitrum-one-bridge/src/common/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  fetchTokenDecimals(tokenAddress: Address): number {
+    const tokenContract = _ERC20.bind(tokenAddress);
+    const call = tokenContract.try_decimals();
+    if (call.reverted) {
+      return 0;
+    } else {
+      return call.value.toI32();
+    }
+  }
```
jaimehgb, line 72: Since you are also potentially dealing with native ETH, you'll want to handle that particular case.

ETH is not an ERC20, so to get the symbol, name, and decimals you cannot make a contract call. This is probably the reason why you have amounts and prices off by orders of magnitude.

fortysevenlabs: @jaimehgb  How do you recommend handling the pricing? Shouldn't native eth pricing be handled by `prices` lib given we can expect this to be needed in most mainnet based subgraphs?

I can think of two ways to manage pricing here:

1. Set token name (to ETH_ADDRESS), symbol (to ETH_SYMBOL), decimals (to 18) if the address is eth_address. Expect `prices` lib to return accurate price. Can `prices` lib return a price when the token info is set? Didn't find reference to code for `prices` lib handling this.

2. Update the `getTokenPrice` and `getAmountValueUSD` functions in the Pricer to get ETH price using WETH price if token address == ETH_ADDRESS (0xee...eee).

dmelotik: I think what @jaimehgb is referring to is that we just need to handle the case when we get the eth address (generally it is (0xEeee... OR 0x0000...). In those cases we need to hardcode the id (0xEeee...eeee), name, symbol, and decimals. Without proper decimals it will make the prices off by A LOT.

As far as pricing goes, once the proper values for native ETH are hardcoded I'd say we can expect the price lib to work properly to get the eth price.

fortysevenlabs: Cool, that aligns with the approach 1 I outlined. Pushed this change to my target, will review and confirm.

dmelotik:
@fortysevenlabs could you update to use the latest bridge schema on this? You should be able to just add the latest SDK and it should build

dmelotik:
Nice work on this @fortysevenlabs ! 🚀

One small thing, could you update to use the latest bridge schema on this? You should be able to just add the latest SDK and it should build

fortysevenlabs:
> One small thing, could you update to use the latest bridge schema on this? You should be able to just add the latest SDK and it should build

@dmelotik done

dmelotik (commented):
Looking good, on ethereum it looks like there is a crazy price in one of the pools that is propogating to all the other USD fields, though.

Also a small linting issue to fix

fortysevenlabs:
Will look into that one pool and try to close this out over the next day

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transfers.ts
```diff
@@ -0,0 +1,197 @@
...
+  const inputTokenBalanceResult = erc20.try_balanceOf(event.address);
+  if (inputTokenBalanceResult.reverted) {
+    log.info("calculate token balance owned by bridge contract reverted", []);
+  } else {
+    inputTokenBalance = inputTokenBalanceResult.value;
+  }
+  pool.setInputTokenBalance(inputTokenBalance!);
+}
+
+export function handleTransferOut3pGateway(event: DepositFinalized): void {
```
fortysevenlabs, line 108: bug: Should be `WithdrawalInitiated` instead of `DepositFinalized`

fortysevenlabs: here and in multiple places below

fortysevenlabs:
Looks like the following are the causing TVL issues due to inaccurate pricing

| # | ID                                    | Input Token Name                   | Input Token Last Price USD | Input Token Balance             | Total Value Locked          |
|---|---------------------------------------|-----------------------------------|---------------------------|--------------------------------|-----------------------------|
| 1 | 0xa3a7b6f88361f48403514059f1f16c8e78d60eec454c4f4e | Dogelon                           | 175500.800666             | 2781721735788799849954828201  | 488194391860949.6807422930132079764 |
| 2 | 0xa3a7b6f88361f48403514059f1f16c8e78d60eec47545053 | Global Transaction Payment Solution | 0.012321              | 365666616666666800000009264229046 | 4505378383950.001642800114144566076 |
| 3 | 0xa3a7b6f88361f48403514059f1f16c8e78d60eec464349 | FUTUREGOLDCOIN                    | 0.000233                  | 579909782200000000000000000000000 | 135118979252.6                 |
| 4 | 0xa3a7b6f88361f48403514059f1f16c8e78d60eec4558504f | Exponential Capital                | 14.430236                  | 3994917233302904654778261682 | 57647598477.02797365394884374101695 |
| 5 | 0xa3a7b6f88361f48403514059f1f16c8e78d60eec20 | 0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2 | 678.90189        | 15975632571876328574283513     | 10845887146.99240031534208237153957 |


Observations
* Pricing for the Dogelon [seems to be way off](https://etherscan.io/dex/uniswapv2/0x7b73644935b8e68019ac6356c40661e1bc315860).
* Pricing for Exponential Capital is [also off](https://etherscan.io/dex/uniswapv2/0x8e1b5164d4059fdec87ec5d0b9c64e4ff727b1ed)
* The numbers being reported and calculations (decimals, input token balance, price) look fine for GTPS and FUTUREGOLDCOIN. 
* The last one is MKR. The inputTokenBalance for the pool.id (event.address) is 95397874689328253235, unsure how its getting 15975632571876328574283513. `(678.90189*95397874689328253235)/(10^18) would have translated to ~$65k`.

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transfers.ts
dmelotik: Is this an action item to fix this?

fortysevenlabs: yes, action item for myself, will resolve

fortysevenlabs:
Aligned, was taking the same approach. TVL for all 4 of those is bonkers :)

subgraphs/arbitrum-one-bridge/src/handlers/arbitrum/transfers.ts
fortysevenlabs: done

fortysevenlabs:
@dmelotik @jaimehgb Can we move this to final QA?

bye43:
> @dmelotik @jaimehgb Can we move this to final QA by messari team?

@fortysevenlabs Doing a quick scan before I send over, but should the TVL on the mainnet side look this erratic? I believe on L[2 beats](https://l2beat.com/scaling/projects/arbitrum) uses the same methodology as us for TVL and looks different. 
arbitrum-one-daily-totalValueLockedUSD-041423

fortysevenlabs:
@bye43 It looks like this is happening because of [MKR pool](https://subgraphs.messari.io/subgraph?endpoint=https%3A//api.thegraph.com/subgraphs/name/fortysevenlabs/arbitrum-one-bridge-ethereum&tab=pool&poolId=0xa3a7b6f88361f48403514059f1f16c8e78d60eec20).

We observed [issues with MKR TVL earlier](https://github.com/messari/subgraphs/pull/1691#issuecomment-1501166855) which had gone away in a future deployment but looks like it came back again in the latest deployment. This seems like an issue with incorrect inputTokenBalances in some deployments. Working on identifying the root cause and fix.

fortysevenlabs:
@bye43 Pushed a fix for the odd TVL issues. Lets review after 24 hours when syncing would be 90% complete.

bye43:
@fortysevenlabs Ya this looks better. Let me quickly check this EOD when its fully synced and will send after. Thanks

subgraphs/arbitrum-one-bridge/src/handlers/ethereum/transfers.ts
```diff
@@ -228,3 +240,20 @@ export function handleTransferOut(event: DepositInitiated): void {
   }
   pool.setInputTokenBalance(inputTokenBalance!);
 }
+
+function ignoredToken(tokenAddress: string): bool {
+  const IGNORED_TOKENS: Array<string> = [
+    "0x761d38e5ddf6ccf6cf7c55759d5210750b5d60f3",
```
dmelotik, line 246: Can you comment the token names inline?

dmelotik:
@fortysevenlabs 

- I am actually thinking a better option for the 4 ignored tokens is to just force their price to be 0. That way we still have those pool routes with values in their native currency. Then in the readme we can make a note saying that we did that for those tokens bc we couldn't find good pricing on chain.
- Can you fix the pool.name here as well. I think it is safe to make the name the same as the input token name

fortysevenlabs:
@bye43 we were talking about the same thing but I realized I mixed up the loadPool vs initialize functions. I reverted the pool initialize to use inputToken.name. The pool names are now as expected ([reference](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmepqmDvtKS6AwzY6vvKeMUMGrVRdZWwcCZZ32RhqGjQCf&tab=poolOverview)).

dmelotik (approved):
Looking real nice @fortysevenlabs !

I am approving, and we will merge once we get the green light on the QA cc @bye43

PR #1692, opened 2023-02-05
chore(): lint price-lib and sdk

jaimehgb:
Made all addresses in price-lib and SDK lowercase so that when using them in our subgraphs they don’t break the linter.


steegecs (approved):
Good stuff!

harsh9200 (approved):
LGTM

PR #1693, opened 2023-02-06
feat(#minor): Bridge Schema; added net value exported and updated SDK

jaimehgb:
Since TVL for bridges is ambiguous we decided to add an additional field `netValueExportedUSD`. This is the value of all assets bridged in or away from a given pool and/or chain.

In other words, every asset currently locked in the bridge because it is now in some other chain counts towards value exported. And the other way around, every asset locked in other chains that is now living in the current chain counts towards value exported negatively.

For LOCK/RELEASE pools, total value exported is the value of all assets locked in the pool.
For BURN/MINT pools, total value exported is negative the value of all assets minted by the pool.
For liquidity based pools, the value exported is the difference between their balance and the total liquidity that has been provided to the pool.

At a pool level, we added `netValueExported`, because for a given pool it can only be positive or negative. At the protocol level we broke it down into `totalValueImportedUSD & totalValueExportedUSD`. This is so we can differentiate the sum of all positive and negative pools separately. Meaning “the value of native assets on this chain that are now abroad” and “the value of native assets from other chains that are now here”.

Find a deployment featuring this changes here: https://github.com/messari/subgraphs/pull/1735


subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
```diff
@@ -252,6 +261,80 @@ export class Pool {
...
+  /**
+   * Adds the given delta to the net value exported for this pool.
+   * Will also update the protocol's net value exported based on the change in this pool's.
+   * Since at the protocol level valueExported is broken down into totalValueExported and totalValueImported,
+   * it is a bit tricky to update those. In order to do so, we deduct the current netValueExported from
+   * the protocols, recalculate the pool's netValueExported, and then add it back again to the protocol.
+   *
+   * @param delta The change in net value exported for this pool.
+   */
+  addNetValueExportedUSD(delta: BigDecimal): void {
```
tnkrxyz, line 284: Will the logic be more clear if we track totalValueImported/totalValueExported - adding/subtracting when transferIn/Out - and derive netValueExported from them?

tnkrxyz (approved):
Looks good. I have only two comments.

schema-bridge.graphql
```diff
@@ -277,9 +277,15 @@ type BridgeProtocol implements Protocol @entity {
 
   ##### Quantitative Data #####
 
-  " Current TVL (Total Value Locked) of the entire protocol on this chain. "
+  " Current TVL (Total Value Locked) of the entire protocol on this chain. This is the sum of totalValueExported + liquidity deposits "
   totalValueLockedUSD: BigDecimal!
 
+  " USD value of all assets bridged away from this chain. In other words, value of assets locked in contracts in this chain because they are liquid in some other chain. "
+  totalValueExportedUSD: BigDecimal!
```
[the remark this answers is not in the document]
jaimehgb: Although I agree that it would be more consistent, I think it would be more ambiguous. It could be mistaken with volume.

Whereas value imported/exported, while it can still be mistaken with volume, it is not related to value transferred, but to value living somewhere else, if that makes sense 😅 

Naming being naming

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
jaimehgb: I think we can’t do that. Because value imported/exported is not exactly volume. It’s the value of all assets currently living in the bridge  (or away) for a given pool.

For lock/release pools, it will be the value of all assets locked. And this value can change both on transferin/out but also if the price of the asset changes.

Same for mint/burn pools, it will be the value of all supply minted.

For lock/release, totalValueImported will always be 0. And totalValueExported will be the value of locked assets (goes up and down with transfers and with price changes).
For mint/burn, totalValueExported will always be 0. And totalValueImported will be the value of supplies minted (goes up and down with transfers and price changes).

schema-bridge.graphql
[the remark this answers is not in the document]
tnkrxyz: makes sense. thanks!

subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts
tnkrxyz: makes sense.

PR #1697, opened 2023-02-06
fix(#patch): fix SDK broken build

### LINKED ISSUE #1696: #bug; event property doesn’t exist error (in sdk)
**Describe the bug**
`CustomEventType` was  moved from https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/protocols/bridge/index.ts to https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/util/events.ts but wasn’t updated in all SDK classes (eg Protocol, Pool, Account, etc)

**Reproduction**
* Go to any bridge subgraph that uses bridge sdk
* Ensure you are using the latest sdk and run `messari build`, you should see many `ERROR TS2339: Property 'event' does not exist on type 'src/sdk/protocols/bridge/….. '` errors

**Expected behavior**
Update the the classes to import CustomEventType from src/sdk/util/events.ts and ensure build passes for subgraphs using sdk

jaimehgb:
The build of the SDK broke after moving a few things around. This fixes it.
Fixes #1696

### DIFF
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -14,7 +14,7 @@ import { Bridge } from "./protocol";
 import { TokenManager } from "./tokens";
 import { BridgePoolType, TransactionType, TransferType } from "./enums";
 import { getUnixDays, getUnixHours } from "../../util/events";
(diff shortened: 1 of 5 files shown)


tnkrxyz (approved):
Looks good. approved.

PR #1698, opened 2023-02-06
chore(); Update deployments.json; Stargate deployments

bye43:
Change stargate deployments (avax, bsc, arb, matic, op) to `dev`


dmelotik (approved):
lgtm, from your comment sounds like you are keeping fantom prod?

PR #1704, opened 2023-02-07
fix(#minor); Network; Remove values array

dmelotik:
Removing the array `Stat.values` from the schema. This is slowing down syncing speed as some networks have arrays with >100k entries. The challenge is that this requires us to calculate variance and standard deviation in a stream w/o going back.

The variance algo has a lot of literature, but I am still skeptical if quartiles are possible to accurately do this way.

Test:
- With arrays: https://okgraph.xyz/?q=messari%2Fnetwork-optimism
- Without arrays: https://okgraph.xyz/?q=dmelotik%2Fnetwork-optimism


jaimehgb (approved):
Math checks out

dmelotik:
thanks chatgpt

PR #1708, opened 2023-02-08
feat (#new); satellite bridge/axelar network; new subgraph

tnkrxyz:
This PR adds the satellite subgraph (WIP)

### TODOs
- [x] no source chain & account for mintToken and burnToken
- [x] deployments for other chains
- [x] initial QA 

#### ethereum
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-ethereum
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-ethereum&tab=protocol
#### Arbitrum
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-arbitrum
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-arbitrum&tab=protocol
#### Avalanche
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-avalanche
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-avalanche&tab=protocol
#### BSC
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-bsc
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-bsc&tab=protocol
#### celo
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-celo
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-celo&tab=protocol
#### fantom
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-fantom
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-fantom&tab=protocol
#### moonbeam
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-moonbeam
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-moonbeam&tab=protocol
#### polygon
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-polygon
- validation: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-polygon&tab=protocol


tnkrxyz:
Like cbrdige subgraph, the axelar/satellite subgraph uses callHandlers, so it cannot be deployed on many of the chains (because callHandlers is not supported on hosted services).

bye43:
@tnkrxyz Does this need a full QA? I believe that @jaimehgb and @dhruv-chauhan may have a work around for the `callHandler` problem

tnkrxyz:
> @tnkrxyz Does this need a full QA? I believe that @jaimehgb and @dhruv-chauhan may have a work around for the `callHandler` problem

I have a solution that I implemented for the cbridge subgraph, and plan to do the same for the satellite subgraph. I am very interested what solution @jaimehgb and @dhruv-chauhan come up.

dhruv-chauhan:
Here's an idea,

`function handleDeployToken(call: DeployTokenCall)`
can be removed, instead, we can just use already implemented
`function handleTokenDeployed(event: TokenDeployed)`
to get 
 - (symbol, address) {from params}, and 
  - (tokenType) {maybe by checking if [owner](https://etherscan.io/address/0x31dab3430f3081dff3ccd80f17ad98583437b213#readContract#F7) is AxelarGatewayProxyMultisig}
  we don't really need other `_TokenSymbol` fields as far as I can tell

You would need to add `WrappedTokenTemplate` as data source template and [abi](https://etherscan.io/address/0x31dab3430f3081dff3ccd80f17ad98583437b213#code) `WrappedToken` in our template.yaml

in `handleTokenDeployed` then,
```
if (tokenType != EXTERNAL)
  WrappedTokenTemplate.create()
```
 <hr>

`function handleMintToken(call: MintTokenCall)` and 
`function handleBurnTokenCall(call: BurnTokenCall)`
can be changed to 
`function handleWrappedToken(event: WrappedTokenTransfer)`
 - (from, to, amount) {from params}

```
if (from == Address(0)) // https://etherscan.io/address/0x31dab3430f3081dff3ccd80f17ad98583437b213#code#L404
  _handleMintToken
if (to == Address(0)) // https://etherscan.io/address/0x31dab3430f3081dff3ccd80f17ad98583437b213#code#L425
  _handleBurnToken
```
<hr>

`function handleFeeRefund(call: RefundCall)`
I did not find an obvious substitute, but to my understanding, we typically do not consider gas fees in protocol's revenue. And if that's agreed on, we could simply ignore this.

@tnkrxyz

tnkrxyz:
I switched to use eventHandlers for all except for fee refund, which I keep for chains supporting callHandlers.

subgraphs/axelar/src/mappings.ts
```diff
@@ -0,0 +1,762 @@
...
+    customEvent
+  );
+}
+
+export function handleContractCallApprovedWithMint(
+  event: ContractCallApprovedWithMint
+): void {
+  // contract call
+  const srcChainId = networkToChainID(event.params.sourceChain.toUpperCase());
+  // this is needed to support transferIn from non-EVM chains
```
dmelotik, line 298: I don't think any of the chains are non evm. Do you mean non-ethereum chains?

subgraphs/axelar/src/mappings.ts
```diff
@@ -0,0 +1,762 @@
...
+        [
+          fromAddress.toHexString(),
+          toAddress.toHexString(),
+          thisLog.data.toHexString(),
+          thisLog.transactionHash.toHexString(),
+          thisLog.logIndex.toString(),
+        ]
+      );
+
+      log.info("[handleCommandExecuted]to==0x0 {} tx={}-{}", [
```
dmelotik, line 399: why do we need this log? It seems to have the same informationas the one above but the to address is 0, which you could see in the above log

dmelotik (commented):
Commenting @tnkrxyz ! Nice work 💪

- There is negative revenue in ethereum / fantom
- @bye43 I think we may have to decide which axelar deployments we want to be prod and which ones not. There are some limitations based on the contract structure.
- There is no TVL on any of the subgraphs. In the readme you say it is what is stored in the gateway contracts. That makes sense to me.
    - One way you could do this would just be to call `ERC20.balanceOf`. Or using the input/output from the chain?

subgraphs/axelar/src/mappings.ts
tnkrxyz: There are transfers in/out with cosmos chain, and they have non-evm addresses.

subgraphs/axelar/src/mappings.ts
tnkrxyz: good point. it is remain from debugging. deleted.

subgraphs/cbridge/README.md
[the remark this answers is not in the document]
tnkrxyz: would you mind clarfying what you mean?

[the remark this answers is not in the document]
dmelotik: There are changes to the cBridge readme in this PR

subgraphs/axelar/src/mappings.ts
dmelotik: Ah, gotcha

tnkrxyz:
> Commenting @tnkrxyz ! Nice work 💪
 
thx!

>     * There is negative revenue in ethereum / fantom

This can happen because there is a mechanism to refund fees - which may happen in different day the fee payment and lead to negative fees.
 
>     * @bye43 I think we may have to decide which axelar deployments we want to be prod and which ones not. There are some limitations based on the contract structure.

>     * There is no TVL on any of the subgraphs. In the readme you say it is what is stored in the gateway contracts. That makes sense to me.
>     
>       * One way you could do this would just be to call `ERC20.balanceOf`. Or using the input/output from the chain?

I believe for the bridge protocol, we only count liquidity on a pool-based bridge in TVL, since axelar doesn't have pool-based bridge, the TVL is 0.

subgraphs/cbridge/README.md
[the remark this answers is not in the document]
tnkrxyz: oops, that's certainly unintended. fixed.

dmelotik (dismissed):
Thanks for the context! Approved, we will do a QA now

tnkrxyz:
> Edit: @tnkrxyz we do need to add TVL for lock & mint style bridges. The TVL is only what is locked though on the origin chain. So really just the balances of tokens in the Gateway contract for Axelar!

@dmelotik , sure we can do that. Although I think if that is what we want to do, it makes more sense to change it at the bridge sdk than at the specific protocol/subgraph. @jaimehgb  what do you think?

jaimehgb:
> > Edit: @tnkrxyz we do need to add TVL for lock & mint style bridges. The TVL is only what is locked though on the origin chain. So really just the balances of tokens in the Gateway contract for Axelar!
> 
> @dmelotik , sure we can do that. Although I think if that is what we want to do, it makes more sense to change it at the bridge sdk than at the specific protocol/subgraph. @jaimehgb  what do you think?
> 

The SDK should do that for you already! If you properly mark each pool as Lock/Release, Mint/Burn or Liquidity, it is smart enough to add to the TVL or not.

Lock/release and liquidity pools have TVL, mint/burn don't. If you expect TVL to be non-zero for a given pool, it is likely set as the wrong type

tnkrxyz:
> The SDK should do that for you already! If you properly mark each pool as Lock/Release, Mint/Burn or Liquidity, it is smart enough to add to the TVL or not.
> 
> Lock/release and liquidity pools have TVL, mint/burn don't. If you expect TVL to be non-zero for a given pool, it is likely set as the wrong type

You're absolutely right. I treated all tranfers as BURN_MINT. fixed & re-deployed.

dmelotik:
@tnkrxyz I think the deployments are still missing volume in/out, and TVL/inputTokenBalance from what I see

tnkrxyz:
> @tnkrxyz I think the deployments are still missing volume in/out, and TVL/inputTokenBalance from what I see

Are you looking at the pending version: 

[image](https://subgraphs.messari.io/subgraph?endpoint=https%3A%2F%2Fapi.thegraph.com%2Fsubgraphs%2Fid%2FQmZ7wBLUMDcvNRiHsAhdudVtYVDTpBNQmYNk6PMSVBe3RS&tab=protocol&version=pending&name=tnkrxyz%2Faxelar-ethereum&view=financialsDailySnapshots-totalValueLockedUSD)


Some of the subgraph hasn't fully synced yet.

dmelotik (approved):
LGTM @tnkrxyz ! Nice work on this 🚀

Issue #1695, opened 2023-02-06
MakerDao Ethereum: ETH-A and ETH-B TVLs are wrong


ishraq8:
**Describe the bug**
Eth-A and Eth-B TVL are off, looks like `inputtokenbalances` are off.

**Reproduction**

Following Vaults are the issue:
ETH-A: 0x2f0b23f53734252bda2277357e97e1517d6b042a
ETH-B: 0x08638ef1a205be6762a8b935f5da9b700cf7322c
![Screen Shot 2023-02-06 at 11 18 46 AM](https://user-images.githubusercontent.com/47088402/217025977-e01b0c2b-9bf4-4de2-90c2-be09f4f9f494.png)

![Screen Shot 2023-02-06 at 11 20 59 AM](https://user-images.githubusercontent.com/47088402/217026632-6400079c-4558-49d2-8d9e-42617e41b94f.png)

![Screen Shot 2023-02-06 at 11 21 56 AM](https://user-images.githubusercontent.com/47088402/217026644-65e1ecb5-f70f-44ac-bae6-d66f502be033.png)

**Expected behavior**
This may be a input balance issue or pricing issue on weth. double check. Doesn't match onchain.

**Additional context**
Everything in MakerDAO is accurate except this.

bye43:
@tnkrxyz This is weird. [Looks](https://api.thegraph.com/subgraphs/name/messari/makerdao-ethereum/graphql?query=%7B%0A++markets+%28where%3A+%7Bid_in%3A+%5B%220x2f0b23f53734252bda2277357e97e1517d6b042a%22%2C%220x08638ef1a205be6762a8b935f5da9b700cf7322c%22%5D%7D%29+%7B%0A++++name%0A++++id%0A++++inputTokenBalance%0A++++inputTokenPriceUSD%0A++%7D%0A%7D) like the `inputTokenBalance` is off for both ETH A and B.[ ETH A etherscan](https://etherscan.io/address/0x2f0b23f53734252bda2277357e97e1517d6b042a) and [ETH B etherscan](https://etherscan.io/address/0x08638ef1a205be6762a8b935f5da9b700cf7322c)

tnkrxyz:
that's odd. looking into this.

tnkrxyz:
@bye43 , @ishraq8 , sorry this take a long while to figure out/fix. I believe I figured out the issue and have a fix #1717 if verified working.

ishraq8:
Fix is working, will close once merged and backfilled

PR #1717, opened 2023-02-09
fix (#minor); makerdao; fixed inaccurate TVL

tnkrxyz:
This PR fixes inaccurate TVL/inputTokenBalance in maker markets, esp. ETH-A and ETH-B, as reported in #1695 . Basically, a refactoring I did to consolidate handlers causes liquidated collateral not removed from TVL. To fix this, I moved liquidatePosition call up to `Cat.Bite()` and `Dog.Bark()` when liquidation auction was first set up, from when liquidation auction ends (`handleFlipEndAuction()` and `handleClipTakeBid()`).

### Deployment & validation (sync in progress)
- test deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/makerdao-ethereum?version=pending&selected=logs
- validation dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmdWkF7LsHEq88hrMAMHdoCeNxLyzSqNZdd2qyHiCCpCXQ&tab=protocol

### DIFF
--- a/subgraphs/makerdao/src/common/helpers.ts
+++ b/subgraphs/makerdao/src/common/helpers.ts
@@ -7,7 +7,6 @@ import {
   Withdraw,
   Borrow,
   Repay,
-  Liquidate,
   Token,
   Position,
   PositionSnapshot,
--- a/subgraphs/makerdao/src/common/helpers.ts
+++ b/subgraphs/makerdao/src/common/helpers.ts
@@ -723,13 +723,11 @@ export function liquidatePosition(
   ilk: Bytes,
   collateral: BigInt, // net collateral liquidated
   debt: BigInt, // debt repaid
-): void {
+): string[] {
   const protocol = getOrCreateLendingProtocol();
   const market: Market = getMarketFromIlk(ilk)!;
   const accountAddress = getOwnerAddress(urn);
   const account = getOrCreateAccount(accountAddress);
-
-  const liquidate = Liquidate.load(createEventID(event))!;
 
   log.info("[liquidatePosition]urn={}, ilk={}, collateral={}, debt={}", [
     urn,
@@ -774,10 +772,6 @@ export function liquidatePosition(
   lenderPosition.balance = lenderPosition.balance.minus(collateral);
   lenderPosition.liquidationCount += INT_ONE;
 
-  assert(
-    lenderPosition.balance.ge(BIGINT_ZERO),
-    `[liquidatePosition]balance of position ${lenderPosition.id} ${lenderPosition.balance} < 0`,
-  );
   if (lenderPosition.balance == BIGINT_ZERO) {
     // lender side is closed
     lenderPosition.blockNumberClosed = event.block.number;
@@ -794,15 +788,12 @@ export function liquidatePosition(
   lenderPosition.save();
   snapshotPosition(event, lenderPosition);
 
-  //TODO: this should be an array/list including borrowerPosition and lenderPosition
-  liquidate.position = lenderPosition.id;
-  liquidate.save();
-
   protocol.save();
   market.save();
   account.save();
 
-  assert(account.openPositionCount >= 0, `Account ${account.id} openPositionCount=${account.openPositionCount}`);
+  //assert(account.openPositionCount >= 0, `Account ${account.id} openPositionCount=${account.openPositionCount}`);
+  return [lenderPosition.id, borrowerPosition.id];
 }
 
 export function snapshotPosition(event: ethereum.Event, position: Position): void {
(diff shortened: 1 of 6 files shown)


bye43:
@tnkrxyz Was going to take a look at this, but looks like the deployment failed.

tnkrxyz:
> @tnkrxyz Was going to take a look at this, but looks like the deployment failed.

Fixed (hopefully) and redeployed.

ishraq8:
looks like the syncing is failing

tnkrxyz:
> looks like the syncing is failing

Yup, a different issue. Fixed & restarted.

ishraq8:
Issue resolved, thanks @tnkrxyz 
Screen Shot 2023-02-12 at 2 41 09 PM
Screen Shot 2023-02-12 at 2 41 27 PM

subgraphs/makerdao/src/common/helpers.ts
```diff
@@ -438,10 +437,17 @@ export function updatePosition(
...
-    assert(
-      lenderPosition.balance.ge(BIGINT_ZERO),
-      `[updatePosition]balance for position ${lenderPosition.id} ${lenderPosition.balance} < 0`,
-    );
+    if (lenderPosition.balance.le(BIGINT_ZERO)) {
+      if (lenderPosition.balance.ge(BIGINT_NEG_HUNDRED)) {
+        // a small negative lender position, likely due to rounding
+        lenderPosition.balance = BIGINT_ZERO;
+      } else {
+        log.error("[updatePosition]balance for position {} = {} < 0", [
```
dmelotik, line 445: should anything be done if pos balances are "very" negative?

subgraphs/makerdao/src/mapping.ts
```diff
@@ -445,20 +447,28 @@ export function handleDogBark(event: BarkEvent): void {
   const due = event.params.due; //including interest, but not penalty
 
   const market = getMarketFromIlk(ilk)!;
-  const storeID = clip.toHexString().concat("-").concat(id.toString());
-
-  // remove borrowed amount from borrowed balance
-  // collateral/tvl update is taken care of when it exits vat
-  // via the slip() function/event
+  const token = getOrCreateToken(market.inputToken);
```
dmelotik, line 450: Some duplicate logic in `handleDogBark()` and `handleCatBite()` but it is not too much and there would be a lot of variables if you wanted to make a helper function so I will leave it.

Also, the naming schema in makerdao is so strange

subgraphs/makerdao/src/common/helpers.ts
tnkrxyz: I cannot think of a case where this would happen and is reasonable/correct data, which is why I log an error here. I can of course do `log.critical()` to stop the subgraph here. Do you prefer the latter?

subgraphs/makerdao/src/mapping.ts
tnkrxyz: Yep, I am on the same boat. There are just too many variables and difference. The temporary data store entities are also different, FlipBidStore for cat and ClipBidStore for dog. Because it is two stage auction for cat (the old collateral auction contract, one stage for dog). I lean toward leaving it as it is.

The strange naming scheme is a hallmark of makerdao code. 😁

dmelotik: maker code seems like a rabbit hole that will take me a while to get through. You do a good job 🙏

subgraphs/makerdao/src/common/helpers.ts
dmelotik: tbh I really prefer to not add/subtract position balances and TVL stuff bc it doesn't account for interest. But if there are no contract calls that would give us the position balance of an account, this is okay.

tnkrxyz: That's a good point. I can switch to get position balance using contract call.

ishraq8:
@tnkrxyz deployment failed

subgraphs/makerdao/src/common/helpers.ts
tnkrxyz: This doesn't work. It seems there are cases of multiple transactions in the same block & getting position balance from contract call gets the final balance. I am reverting. I think we should fix the this current issue & come back to do the refectory later, e.g. when upgrading schema to 3.0. What do you think, @dmelotik ?

BTW, I plan to re-do the prettier formating (I used an old config) once you're done reviewing, because I don't want to show thousands of formatting changes.

dmelotik: Okay, let's leave as is @tnkrxyz. Feel free to do the prettier update in a separate commit. I will approve and you can merge when the checks pass

dmelotik (approved):
lgtm. contingent on undoing the `position.balance` refactor and doing prettier updates

ishraq8:
thnks @tnkrxyz

dmelotik:
@tnkrxyz @ishraq8 the protocol side revenue looks like there was an issue

PR #1720, opened 2023-02-10
fix(#minor); comp v2; fix div by 0 runtime error

dmelotik:
Fixing div by 0 runtime error compound V2.

Testing graft: https://okgraph.xyz/?q=dmelotik%2Fcompound-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1318,7 +1318,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.7.8",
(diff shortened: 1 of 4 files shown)


tnkrxyz:
Maybe I missed sth, how is the change/fix related to "div by 0 runtime error"?

jaimehgb (approved):

jaimehgb:
> Maybe I missed sth, how is the change/fix related to "div by 0 runtime error"?

If the network name is wrong it won’t match any of the `if` branches, and the default value that function returns is a 0 :D

Issue #1694, opened 2023-02-06
#bug; introduction of block param to price getters in prices-lib breaks subgraph sdk


fortysevenlabs:
**Describe the bug**
The addition of `block` param breaks (in this https://github.com/messari/subgraphs/commit/9b4b172cd2ea3990f2247e54a23734f504303d41) breaks `Pricer`, `TokenPricer` in the sdk.

**Reproduction**
Run `messari build` in any of the protocols that use the `sdk`.

**Expected behavior**
Update the `TokenPricer` interface and add documentation for how to use `Pricer` for bridge sdk.

fortysevenlabs:
@harsh9200 @jaimehgb fyi

bye43:
Issue related: https://github.com/messari/subgraphs/issues/1696

jaimehgb:
@harsh9200 is it possible to make the block optional so it doesn’t break old subgraphs when upgrading the price lib version?

harsh9200:
@jaimehgb yeah, that is feasible. I will do that.

PR #1738, opened 2023-02-13
feat(#minor); Price lib; make block param optional

harsh9200:
Fixes: #1694


subgraphs/_reference_/src/prices/calculations/CalculationsCurve.ts
```diff
@@ -6,10 +6,10 @@ import { CalculationsCurve as CalculationsCurveContract } from "../../../generat
 
 export function getCalculationsCurveContract(
   contract: OracleContract,
-  block: ethereum.Block
+  block?: ethereum.Block
```
jaimehgb, line 9: In AS this won’t work. Instead do:
`block: ethereal.Block | null = null`

jaimehgb (changes requested):
Tiny comment on making parameters optional on AssemblyScript. Other than that looks good to me!

jaimehgb (approved):

PR #1739, opened 2023-02-13
feat(); Perps Schema SDK; Add perps schema SDK

harsh9200:


dmelotik:
@harsh9200 there are a few missing fields that I am adding in rn. I will add you as a reviwer on that schema change so you are familiar. It should not be difficult to update the SDK to support it. ✅

harsh9200:
@dmelotik Sounds Good.

dmelotik:
https://github.com/messari/subgraphs/pull/1748

ishraq8:
@dmelotik Just checking if other changes need to be made.

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
```diff
@@ -0,0 +1,554 @@
...
+    deposit.logIndex = this.event.logIndex.toI32();
+    deposit.protocol = this.protocol.getBytesID();
+    deposit.to = pool.getBytesID();
+    deposit.from = this.account.id;
+    deposit.account = this.account.id;
+    deposit.blockNumber = this.event.block.number;
+    deposit.timestamp = this.event.block.timestamp;
+    deposit.inputTokens = pool.getInputTokens();
+    deposit.outputToken = pool.getOutputToken();
+    deposit.inputTokenAmounts = amounts;
```
jaimehgb, line 182: Should we validate that `amounts.length == pool.inputTokens.length`?

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
```diff
@@ -0,0 +1,554 @@
...
+    deposit.outputTokenAmount = sharesMinted;
+    deposit.amountUSD = this.getAmountUSD(pool.getInputTokens(), amounts);
+    deposit.pool = pool.getBytesID();
+    deposit.save();
+
+    if (updateMetrics) this.protocol.addTransaction(TransactionType.DEPOSIT);
```
jaimehgb, line 188: This doesn't match with what the comment of the `updateMetrics` param states. It says it will update the pool and protocol TVL values. We should either do that, or change the comment.

Same for withdraw, swap, borrow, etc

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
```diff
@@ -0,0 +1,554 @@
...
+   * @param collateralToken Token which was the collateral
+   * @param amountLiquidated Amount of collateral liquidated in native units
+   * @param liquidator Account that carried out the liquidation
+   * @param liquidatee Account that got liquidated
+   * @param position The position this Liquidate belongs to
+   * @param profitUSD Amount of profit from liquidation in USD
+   * @param updateMetrics Optional, defaults to true. If true it will update the pool and protocol TVL and inputTokenBalance.
+   * @returns Liquidate
+   */
+  liquidate(
```
jaimehgb, line 419: Should we call `countLiquidate` and `countLiquidatee` on the accounts involved from this method? Or do you want to leave that up to the caller?

dmelotik:
@harsh9200 

- There were a few more fields that we snuck in this week: https://github.com/messari/subgraphs/pull/1784
- Also what is the 2.0.0 version you mention?

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+
+    this.pool.rewardTokens = tokens;
+    this.pool.rewardTokenEmissionsAmount = amounts;
+    this.pool.rewardTokenEmissionsUSD = amountsUSD;
+    this.save();
+  }
+
+  /**
+   * Adds some value to the cumulativeUniqueBorrowers counter. If the value is omitted it will default to 1.
+   * If you are loading accounts with the AccountManager you won't need to use this method.
```
jaimehgb, line 571: Is this accurate?

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+   * @param count {u8} The value to add to the counter.
+   */
+  addLiquidatee(count: u8 = 1): void {
+    this.pool.cumulativeUniqueLiquidatees += count;
+    this.save();
+  }
+
+  /**
+   * Adds 1 to the cumulativePositionCount counter and adds 1 to the counter corresponding the given position type.
+   * If you are creating transaction entities from the Account class you won't need to use this method.
```
jaimehgb, line 601: Is this accurate?

subgraphs/_reference_/src/sdk/protocols/perpfutures/poolSnapshot.ts
```diff
@@ -0,0 +1,481 @@
...
+    return getOrCreateVolumeHourlyTracker(hourID, this.pool);
+  }
+
+  getVolumeDailyTracker(): _VolumeDailyTracker {
+    const dayID =
+      this.pool._lastUpdateTimestamp!.toI32() / constants.SECONDS_PER_DAY;
+
+    return getOrCreateVolumeDailyTracker(dayID, this.pool);
+  }
+}
```
jaimehgb, line 397: Why `dailyVolumeUSD` is calculated comparing cumulatives, but `dailyVolumeByToken` is done via a tracker? Is there a reason that prevents us from doing it in a unique way for both?

subgraphs/_reference_/src/sdk/protocols/perpfutures/position.ts
```diff
@@ -0,0 +1,239 @@
...
+      .getAmountValueUSD(this.tokens.getOrCreateToken(token), amount);
+    this.save();
+  }
+
+  /**
+   * Sets the position's collateralBalance value.
+   * @param collateralToken
+   * @param amount
+   */
+  setcollateralBalance(collateralToken: Address, amount: BigInt): void {
```
jaimehgb, line 148:
```suggestion
  setCollateralBalance(collateralToken: Address, amount: BigInt): void {
```

subgraphs/_reference_/src/sdk/protocols/perpfutures/position.ts
```diff
@@ -0,0 +1,239 @@
...
+    this.position.hashClosed = event.transaction.hash;
+    this.position.blockNumberClosed = event.block.number;
+    this.position.timestampClosed = event.block.timestamp;
+    this.save();
+
+    if (this.account) this.account.closePosition(this.position.side);
+    if (this.pool) this.pool.closePosition(this.position.side);
+  }
+
+  takePositionSnapshot(): void {
```
jaimehgb, line 201: Do we want this to be called by the caller manually? Or can we abstract them from doing so?

jaimehgb (commented):
I like it 🚀 

Left a few comments, but good overall. It is hard to predict exactly how we'll be using these until we start building subgraphs with it, so I expect some additions to be made, but it sets a good foundation 💪

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+    this.protocol.addSupplySideRevenueUSD(rev);
+  }
+
+  /**
+   * Adds a given USD value to the pool and protocol protocolSideRevenue. It can be a positive or negative amount.
+   * Same as for the rest of setters, this is mostly to be called internally by the library.
+   * But you can use it if you need to. It will also update the protocol's snapshots.
+   * @param rev {BigDecimal} The value to add to the protocol's protocolSideRevenue.
+   */
+  addProtocolSideRevenueUSD(rev: BigDecimal): void {
```
dmelotik, line 270: The comment says it updates the snapshot, but I don't see the code for that

dmelotik: Same question for most of the following functions

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+  }
+
+  /**
+   * Adds a given USD value to the pool and protocol's supplySideRevenue and protocolSideRevenue. It can be a positive or negative amount.
+   * Same as for the rest of setters, this is mostly to be called internally by the library.
+   * But you can use it if you need to. It will also update the protocol's snapshots.
+   * @param protocolSide {BigDecimal} The value to add to the protocol's protocolSideRevenue.
+   * @param supplySide {BigDecimal} The value to add to the protocol's supplySideRevenue.
+   */
+  addRevenueUSD(protocolSide: BigDecimal, supplySide: BigDecimal): void {
```
dmelotik, line 287: Do we want this to be the function exposed to the user, then the functions that it calls are private and only handled by the sdk? In an attempt to reduce the number of options for the dev and make it simpler

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+
+  /**
+   * Sets the rewardTokenEmissions and its USD value for a given reward token.
+   * It will also create the RewardToken entity and add it to the pool rewardTokens array
+   * if not already present.
+   * @param type The type of reward token
+   * @param token The actual token being rewarded
+   * @param amount The daily amount of reward tokens emitted to this pool
+   */
+  setRewardEmissions(
```
dmelotik, line 524: what if there is more than 1 reward token?

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+    }
+
+    this.pool.openPositionCount -= 1;
+    this.pool.closedPositionCount += 1;
+    this.save();
+
+    this.protocol.closePosition(positionSide);
+  }
+
+  addVolumeByToken(tokenAddress: Address, amount: BigInt): void {
```
dmelotik, line 641: The following 4 functions basically all do the same logic, you could make a generalizable function to handle these updates

subgraphs/_reference_/src/sdk/protocols/perpfutures/protocol.ts
```diff
@@ -0,0 +1,581 @@
...
+    this.save();
+  }
+
+  /**
+   * Adds a given USD value to the protocol supplySideRevenue. It can be a positive or negative amount.
+   * Same as for the rest of setters, this is mostly to be called internally by the library.
+   * But you can use it if you need to. It will also update the protocol's snapshots.
+   * @param rev {BigDecimal} The value to add to the protocol's supplySideRevenue.
+   */
+  addSupplySideRevenueUSD(rev: BigDecimal): void {
```
dmelotik, line 276: similar thing I mentioned on the pools applies here

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
dmelotik: Ah I see, the snapshots are literally taken by calling `takeSnapshot()`. Are there functions in the SDK that we can make those calls from so the user doesn't need to touch it?

dmelotik (commented):
similar thoughts as jaime, we will learn more as we dev a subgraph with this. For now, those are my comments

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
harsh9200: Yeah, I will update the comment

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+        inputTokenBalance
+      );
+      totalValueLockedUSD = totalValueLockedUSD.plus(amountUSD);
+    }
+
+    this.setTotalValueLocked(totalValueLockedUSD);
+  }
+
+  /**
+   * Updates the total value locked for this pool to the given value.
```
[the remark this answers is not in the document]
harsh9200: Done

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -0,0 +1,810 @@
...
+   * Will also update the protocol's total value locked based on the change in this pool's.
+   */
+  setVolume(newVolume: BigDecimal): void {
+    const delta = newVolume.minus(this.pool.cumulativeVolumeUSD);
+    this.addVolume(delta);
+    this.save();
+  }
+
+  /**
+   * Adds the given delta to the total value locked for this pool.
```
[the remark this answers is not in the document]
harsh9200: Done

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
harsh9200: yeah, that makes sense. I will make the other two functions private. Let's say the user only wants to update the `protocolSideRevenue`, he can pass the other variable (`supplySideRevenue`) as zero. So exposing all three can be avoided which will make it simpler.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
harsh9200: This function can be used within a for loop when calculating the daily reward emission amount by the user.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
harsh9200: How do you suggest generalizing these 4 functions?

subgraphs/_reference_/src/sdk/protocols/perpfutures/protocol.ts
harsh9200: I cannot make these functions private in the protocol because they are being used in other places.

subgraphs/_reference_/src/sdk/protocols/perpfutures/poolSnapshot.ts
harsh9200: `dailyVolumeByToken` is the total volume of a specific token in a pool on a daily basis. We cannot use the cumulative here because it is calculated using the sum of the volume of all tokens in a pool, that's why I have created a separate tracker for this.

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
```diff
@@ -0,0 +1,554 @@
...
+   * @param pool The pool where the liquidity was swapped.
+   * @param tokenIn The token deposited into the pool.
+   * @param amountIn The token amount deposited into the pool.
+   * @param tokenIn The token withdrawn from the pool.
+   * @param amountIn The token amount withdrawn from the pool.
+   * @param tradingPair  The contract address for the trading pair or pool.
+   * @param updateMetrics Optional, defaults to true. If true it will update the pool and protocol TVL and inputTokenBalance.
+   * @returns Borrow
+   */
+  borrow(
```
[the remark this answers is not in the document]
harsh9200: Done

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
harsh9200: yeah, nice catch. This should be handled by the SDK.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
harsh9200: No, I will fi this.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
harsh9200: No, I will fi this.

subgraphs/_reference_/src/sdk/protocols/perpfutures/account.ts
harsh9200: I think it's better that the function handles this automatically. I have made changes to handle this inside the liquidate function, same for other similar functions.

subgraphs/_reference_/src/sdk/protocols/perpfutures/position.ts
harsh9200: We can abstract this easily by adding `takeSnapshot` function call at the end of every method in the position class, but I think that's not a good practice, saving an entity many times in a single block can sometimes lead to unknown issues. So I think it's better for a user to call it manually. The same goes for protocol and position snapshots. What are your and @dmelotik thoughts on this?

subgraphs/_reference_/src/sdk/protocols/perpfutures/poolSnapshot.ts
jaimehgb: You could keep an array of cumulatives :D

subgraphs/_reference_/src/sdk/protocols/perpfutures/position.ts
jaimehgb: >  saving an entity many times in a single block can sometimes lead to unknown issues

I think that as long as we are always saving the same loaded entity (and not two or more copies loaded at different times) it should be fine. And since we are streamlining the load, update, and update of snapshots of all entities through the SDK I think it is less likely to have the same entity being loaded in multiple places when handling the same event.

We could potentially enforce this at the SDK level. Make the `Protocol` a singleton, so it can only be loaded once, or make it panic if attempted to load multiple times. Same for Pools and accounts. But we can leave that for a future PR if it actually is a problem

dmelotik: I don't think we should call `takePositionSnapshot()` on every single function in the position class. I think we only want to take a snapshot when position.balance changes.

I am for what Jaime said though. I prefer making the dev worry about less function calls.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
dmelotik: You have a helper function that takes the tokenAddress and amount, then also pass in the type of add. And you have an if block or switch statement for the actual add block. 

The rest is the same for each function

ishraq8:
I updated the schema, would it be possible to add the subdivision in this pull request. If it is a lot of work we can push it to another issue. let me know what you think.

https://github.com/messari/subgraphs/pull/1825

harsh9200:
@ishraq8 I will add the new changes in this PR only.

subgraphs/_reference_/src/sdk/protocols/perpfutures/pool.ts
dmelotik: Feel free to resolve if you think this is unnecessary @harsh9200

dhruv-chauhan:
Hey @harsh9200 it'll be great if you could add a README indicating methods that are to be externally set / internally used

dmelotik:
> Hey @harsh9200 it'll be great if you could add a README indicating methods that are to be externally set / internally used

[Here](https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/protocols/bridge/README.md) is a comprehensive example

dmelotik:
@harsh9200 can you version the files in the SDK like something along the lines of this: https://github.com/dmelotik/subgraphs/blob/292e545821134cdf959e582746705a49bfd6f8fa/subgraphs/_reference_/src/sdk/protocols/lending/manager.ts#L54

harsh9200:
Hello @dmelotik @jaimehgb, this PR is ready for review.

schema-derivatives-perpfutures.graphql
```diff
@@ -1803,3 +1902,39 @@ type PositionSnapshot @entity(immutable: true) {
   " Timestamp of this snapshot "
   timestamp: BigInt!
 }
+
+type _ActivityHelper @entity {
```
dmelotik, line 1906: I don't think it is necessary to have different fields for daily/hourly. Couldn't we just have `activeUsers` for example and then that will either be for daily or hourly depending on the id being used?

schema-derivatives-perpfutures.graphql
dmelotik, line 1: If we are going to update the schema let's update the version too

schema-derivatives-perpfutures.graphql
```diff
@@ -1803,3 +1902,39 @@ type PositionSnapshot @entity(immutable: true) {
   " Timestamp of this snapshot "
   timestamp: BigInt!
 }
+
```
dmelotik, line 1905: Make a `####` divider in order to separate the helper fields

schema-derivatives-perpfutures.graphql
dmelotik: also make sure the field comments follow the same standard (ie, capital letter to start)

schema-derivatives-perpfutures.graphql
harsh9200: yeah, that makes sense. I will implement that.

PR #1740, opened 2023-02-13
chore(); comp v2; remove bad deployment ID

dmelotik:
Until Compound V2 is fully synced on the decentralized network we want to ingest data from the hosted service subgraph to reduce downtime.


steegecs (approved):
LGTM

PR #1741, opened 2023-02-13
Feat (#major); DePin Schema

ishraq8:
Generalized DePin schema based on information from The Graph and Livepeer


ishraq8:
@jaimehgb & @steegecs This is the pull request for a generalized Depin schema.

Question: For TVL, Revenue and Reward there is USD and Token balance. Chris brought up what if there is a non native token. Generally DePin use their own token for payments but may have an option to have non native token. I have made the token balances optional then. Jaime, wanted to see your thoughts? Should token balances be taken out all together (at protocol level).

PR #1742, opened 2023-02-14
chore(); docs; immutable entities

dmelotik:
Add some documentation to describe issues with changing immutable entities on production level subgraphs.

### DIFF
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -56,6 +56,8 @@ There are two use cases for `schemaVersion`:
 
 The field `subgraphVersion` this is mainly used for the specific subgraph developer to keep track the implementation. For example, if there is a major refactor, we should bump this version, but this has nothing to do with the schema and will no impact on downstream consumers. There may also be repository-wide implementation upgrades. For example, we might need to reimplement everything in Rust somewhere down the road (for substream upgrades), they it'll be a major upgrade on the implementation (major version bump) but again, no impact on schema or downstream consumer.
 
+> Read more about how to upgrade schema version in [`./CONTIBUTING.md`](CONTRIBUTING.md)
+
 ### Methodology Version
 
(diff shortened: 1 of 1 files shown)


docs/SCHEMA.md
```diff
@@ -209,3 +211,15 @@ In general, feel free to add extra entities that tracks the internal state of th
...
 
 When adding entities, make sure you document them with comments in your schema. User-facing comments should be done in double quotes and developer-facing comments should be done with hash tags.
+
+## Immutable Field Assumptions
+
+There are some fields that we (and subgraph consumers) are going to expect to be immutable. There are technical limitations as to why we cannot force this on a subgraph level, but we want to be very cautious when we change these fields in `prod` subgraphs especially.
+
+Scenario: We expect the `market.id` field to not change on deployments, and a DB ingesting our subgraph will expect this to be the case. But then we change `market.id` and the DB thinks there is a new `market` when in actuality, we just updated the identifier.
+
+In general you should not change the following fields across deployments before consulting with the downstream users:
```
lowjiansheng, line 221: I think the actual fields that are mutable/immutable can be fairly subjective.
Do you think we should add a label or additional comment on the subgraph schema for each field that's immutable?

dmelotik: Yeah we could do that, is there anything you think I missed in the list?

dmelotik: It seems like overkill to say that deposit.id should be immutable (they are now immutable entities, but it could still have diff IDs across versions, technically)

lowjiansheng: Yep agreed, therefore I think it's better to define and note it down on a per-field-per-entity level, rather than blanket declare that ids are all immutable.

dmelotik: I have put this on hold for a while, but it doesn't feel right to add this to every immutable field in the schemas imo.

Feels unnecessary since we have notes here. Really the IDs are the most important field to remain immutable for our downstream ingestion.

harsh9200 (approved):

PR #1754, opened 2023-02-16
chore(); rpl; remove decentralized slug

dmelotik:
Remove decentralized slug. We will add it back in once deployed by @steegecs


dmelotik:
unnecessary

PR #1768, opened 2023-02-17
feat (); bridge sdk; support non-evm addresses for transfer and token

tnkrxyz:
Changed the type from Address to Bytes in transferIn, transferOut, and transfer, and token addresses, so non-EVM address can be supported.

### DIFF
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -158,7 +158,7 @@ export class Account {
   transferOut(
     pool: Pool,
     route: PoolRoute,
-    destination: Address,
+    destination: Bytes,
     amount: BigInt,
     transactionID: Bytes | null = null,
     updateMetrics: boolean = true
@@ -190,7 +190,7 @@ export class Account {
   transferIn(
     pool: Pool,
     route: PoolRoute,
-    source: Address,
+    source: Bytes,
     amount: BigInt,
     transactionID: Bytes | null = null,
     updateMetrics: boolean = true
@@ -210,7 +210,7 @@ export class Account {
   private transfer(
     pool: Pool,
     route: PoolRoute,
-    counterparty: Address,
+    counterparty: Bytes,
     amount: BigInt,
     isOutgoing: boolean,
     transactionID: Bytes | null,
--- a/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
@@ -145,17 +145,17 @@ export class Account {
 
   /**
    * Creates a BridgeTransfer entity for a transfer away from this chain
-   * and updates the volumes at PoolRoute, Pool and Protocol.
+   * to a non-EVM chain and updates the volumes at PoolRoute, Pool and Protocol.
    *
    * @param pool The pool the transfer was made on.
    * @param route The route the transfer went through.
-   * @param destination The account receiving the funds.
+   * @param destination The non-EVM account receiving the funds.
    * @param amount The amount of tokens transferred.
    * @param transactionID Optional transaction ID on the source chain.
    * @param updateMetrics Optional, defaults to true. If true, volumes will be updated at PoolRoute, Pool and Protocol. Make it false if you want to update these manually for some reason. Activity counts and transaction counts will still be updated.
    * @returns BridgeTransfer
    */
-  transferOut(
+  nonEVMTransferOut(
     pool: Pool,
     route: PoolRoute,
     destination: Bytes,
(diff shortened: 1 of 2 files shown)


subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
```diff
@@ -37,7 +37,7 @@ export class TokenManager {
     this.presaver = presaver;
   }
 
-  getOrCreateToken(address: Address): Token {
+  getOrCreateToken(address: Bytes): Token {
```
jaimehgb, line 40: I think token can be left as Address always, because it will always refer to a token in the chain where the subgraph is running, which will always be EVM. Right?

Same thing on the initializer interface.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
```diff
@@ -10,7 +10,7 @@ import { Bridge } from "./protocol";
 import { RewardTokenType } from "../../util/constants";
 
 export interface TokenInitializer {
-  getTokenParams(address: Address): TokenParams;
+  getTokenParams(address: Bytes): TokenParams;
```
jaimehgb, line 13: We shouldn't be changing the interfaces that subgraphs need to implement. Because if now we wanted to upgrade the SDK version a given subgraph is using it will break its implementation

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
```diff
@@ -82,9 +82,9 @@ export class TokenManager {
 
   getOrCreateCrosschainToken(
     chainID: BigInt,
-    address: Address,
+    address: Bytes,
     type: string,
-    token: Address
+    token: Bytes
```
jaimehgb, line 87: Similar thing here. This is a breaking change to the API. There is a chance it works because Address inherits from Bytes, so AS might cast automatically and everything might work.

But instead of changing public contracts, I'd suggest to add new functionality on top instead.

In this case, a new function like:
`getOrCreateNonEVMCrosschainToken(chainID: BigInt, address: Bytes, type: string, token: Address): CrosschainToken`

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
```diff
@@ -190,7 +190,7 @@ export class Account {
   transferIn(
     pool: Pool,
     route: PoolRoute,
-    source: Address,
+    source: Bytes,
```
jaimehgb, line 193: Same comment about changing the API as above. Make a new method instead, for transfer in and out for non evm chains.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: I changed this because I think `pool.getDestinationTokenRoute(crossToken)` passed to `account.transferOut()` needs to support non-EVM addresses.

```
  const crossToken = sdk.Tokens.getOrCreateCrosschainToken(
    dstChainId,
    dstPoolId,
    crosschainTokenType,
    token
  );
  pool.addDestinationToken(crossToken);
  const txHash = event ? event.transaction.hash : call!.transaction.hash;
  const acc = sdk.Accounts.loadAccount(sender);
  const transfer = acc.transferOut(
    pool,
    pool.getDestinationTokenRoute(crossToken)!,
    receiver,
    amount,
    txHash
  );
```

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: I suppose in this case it is better to use the more general `Bytes`? Since Address inherits from Bytes, Address type should still work.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: Exactly, I want to take advantage of the fact Address inheriting from Bytes, and it should not break current code. I think the current change has the advantage that the same interfaces/methods will work for both EVM and non-EVM chains. If we use different methods for EVM and non-EVM transfer, we will need a way to lookup non-EVM chains or detect non-EVM address.

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
tnkrxyz: same as above.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: OK, I realized I need to treat non-EVM address differently anyway with `Address.fromUTF8(non-EVM Address)` instead of `Address.fromString(EVM address)`, but I do not fully grasp why it is better to have two different methods if one can do.

subgraphs/_reference_/src/sdk/protocols/bridge/account.ts
jaimehgb: Resolving to keep the conversation on the other comment

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
jaimehgb: I'd have to test it, but can you do `class Class implements Interface` on a class in which one of its methods doesn't exactly match the signature of the interface? Even if the difference lays in a type that inherits from another? :P 

Does `getTokenParams(address: Address): TokenParams` implement `getTokenParams(address: Bytes): TokenParams`?

I have the opposite opinion here 😅 , I think it is better to use "less general" constraints whenever possible. It removes ambiguity and potential source of errors. If we wanted to be as generic as possible we wouldn't be using types. If there are no scenarios in which we need to pass Bytes, I don't think we should support it

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
jaimehgb: > If we use different methods for EVM and non-EVM transfer, we will need a way to lookup non-EVM chains or detect non-EVM address.

Fair point. The question then is if we want the SDK to handle that magically or to pass that onto the client, since they are closer to the given protocol and might know better (but gives them additional overhead). I have my preference, but no strong feelings :D 

> but I do not fully grasp why it is better to have two different methods if one can do.

I personally prefer to be explicit about the purpose of each method. Specially if the alternative involves modifying its public signature. Also by always defaulting to break everything down that much it makes it harder to end up with a big function/class with too many responsibilities.

If it will be backwards compatible, and being that the purpose of the function is essentially the same, it might not be that big of a deal, though. I'll leave it up to you, your call :)

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
jaimehgb: How does `getOrCreateToken(Address)` affect `getDestinationTokenRoute` or `acc.transferOut`?

tnkrxyz: You're right. token doesn't need to be Bytes type in this case. I was thinking that the token argument in the `sdk.Tokens.getOrCreateCrosschainToken` might be a non-EVM token address, which should not happen.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: Your point makes more sense. Do you think it is needed to create a separate `getOrCreateNonEVMCrosschainToken` method for token? The only difference is to change argument to `address: Bytes` from `address: Address`.

subgraphs/_reference_/src/sdk/protocols/bridge/tokens.ts
tnkrxyz: My test is that I can define the interface using Bytes type, but implements it using Address type. And it doesn't work the other way around.

tnkrxyz: But I don't think I need the tokenParams to be Bytes anyway (see https://github.com/messari/subgraphs/pull/1768#discussion_r1112519056), I will revert this change.

jaimehgb (approved):

PR #1773, opened 2023-02-18
fix(#patch): Fix strange behavior with nullable interfaces on bridge sdk

jaimehgb:
Having `Interface | null` doesn't seem to work well when left null. Instead, I added a default implementation for the TokenManager TokenPresaver that does nothing, and can be replaced with a custom one instead.


tnkrxyz (approved):
looks good. approved.

Issue #1711, opened 2023-02-08
Build Across V2 Subgraph


bye43:
```[tasklist]
### Deployments QA Status
- [ ] https://github.com/messari/subgraphs/issues/2165
- [x] Arb: https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-arbitrum
- [x] OP: https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-optimism
- [x] Polygon: https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-polygon
- [x] Boba: https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-boba
```

bye43:
@fortysevenlabs

bhor123:
Hello @fortysevenlabs,This is Bhor, I am a web3 enthusiast and Dapp developer and i wanted to work with you for developing Across V2 subgraph. I have prior experience developing some test subgraphs and would love to contribute and learn with the amazing team at Messari. Please feel free to reach out to me , I have already studied the schema and would like to be as much help as possible. My email is bhorsharma560@gmail.com and discord is Bhor sharma#7534.

pietydespises:
@fortysevenlabs how is this going?

pietydespises:
Arbitrum: good
Optimism: good
Matic: good
Boba: probably correct, but looks weird due to extremely low use. I suggest we hide this.

PR #1777, opened 2023-02-21
feat(#major); across-v2; add across-v2 subgraph

fortysevenlabs:
Deployment Links for QA
* https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-ethereum
* https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-arbitrum
* https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-optimism
* https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-polygon
* https://thegraph.com/hosted-service/subgraph/fortysevenlabs/across-v2-boba


fortysevenlabs:
@jaimehgb can you review the code?
@bye43 can you assign a QA?
@this-username-is-taken FYI

dmelotik:
@fortysevenlabs is this ready to be code reviewed?

subgraphs/across-v2/subgraph.old.yaml
dmelotik, line 1: can probably remove this file

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -0,0 +1,72 @@
...
+} from "../sdk/protocols/bridge/enums";
+import { Versions } from "../versions";
+import {
+  ACROSS_HUB_POOL_CONTRACT,
+  ACROSS_PROTOCOL_NAME,
+  Pricer,
+  TokenInit,
+} from "../util";
+
+const conf = new BridgeConfig(
```
dmelotik, line 20: These contract address constants will be different across different networks, we need a way to select that

subgraphs/across-v2/src/availableRoutesApi.ts
```diff
@@ -0,0 +1,929 @@
...
+    originChainId: 42161,
+    originToken: "0x82af49447d8a07e3bd95bd0d56f35241523fbab1",
+    destinationChainId: 1,
+    destinationToken: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+  },
+  {
+    originChainId: 42161,
+    originToken: "0x82af49447d8a07e3bd95bd0d56f35241523fbab1",
+    destinationChainId: 1,
+    destinationToken: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
```
dmelotik, line 738: duplicates

dmelotik: Is this object necessary? From what i see the piece we are missing is the destination token (on chain), but if any more tokens are added it will break our subgraph...

dmelotik: Looks like the info is easily avail with a contract call on mainnet, and then the avail in the respective chains relayer contract as seen in this transaction: https://optimistic.etherscan.io/tx/0x6b6bdae2b28d9f7fb58b9ea59de11a2df79a9f4bb41d0d941da3134df1752d68#eventlog

dmelotik:
A few things to fix before I dive in for a deeper review:

- Can you pull out all of the other SDKs and only keep the bridge SDK in here. There is also some changes to the bridge SDK since you made this PR.
- Could you fix the build error, looks to just be a deployment.json quirk and some import errors.
- I see the deployments have 0 values in the cumulative snapshots and negative TVL/revenue. Those will need to be fixed. And then can you redeploy all the networks?
- It seems like mainnet has an LP token factory where people can add liquidity to the protocol. In return I would think it gives out an LP token (`outputToken`) to track position.

fortysevenlabs:
- [ ] Can you pull out all of the other SDKs and only keep the bridge SDK in here. There is also some changes to the bridge SDK since you made this PR.
- [ ] Could you fix the build error, looks to just be a deployment.json quirk and some import errors.
- [ ] I see the deployments have 0 values in the cumulative snapshots and negative TVL/revenue. Those will need to be fixed. And then can you redeploy all the networks?
- [ ] It seems like mainnet has an LP token factory where people can add liquidity to the protocol. In return I would think it gives out an LP token (outputToken) to track position.

subgraphs/across-v2/subgraph.old.yaml
fortysevenlabs: done

subgraphs/across-v2/src/availableRoutesApi.ts
fortysevenlabs: > Looks like the info is easily avail with a contract call on mainnet, and then the avail in the respective chains relayer contract as seen in this transaction: https://optimistic.etherscan.io/tx/0x6b6bdae2b28d9f7fb58b9ea59de11a2df79a9f4bb41d0d941da3134df1752d68#eventlog

1) Let's take an example to discuss. Assume, I'm handling a filled relay on optimism subgraph and I get access to originChainId, destinationChainId, destinationToken from the event log. How do I find the originToken in this case?

2) Which contract call are you referring to mainnet? I didn't find any and based on my discussion in across discord, it wasn't possible to get originToken. For convenience, contracts on mainnet: [hub pool contract](https://etherscan.io/address/0xc186fA914353c44b2E33eBE05f21846F1048bEda#readContract), [spoke pool](https://etherscan.io/address/0x4D9079Bb4165aeb4084c526a32695dCfd2F77381#readProxyContract), [adaptor](https://etherscan.io/address/0x527E872a5c3f0C7c24Fe33F2593cFB890a285084#code)

fortysevenlabs: >Is this object necessary? From what i see the piece we are missing is the destination token (on chain), but if any more tokens are added it will break our subgraph...

@jaimehgb and I had discussed this. we decided to go ahead with it but document it because it wasn't possible. if it is possible to fetch this on-chain, we can move away from this approach.

fortysevenlabs: > duplicates
I intentionally didn't do any transformation/modifications of the API response so we anybody could easily update this object when required. The API response does return duplicates.

Here's the command I use for creating the object value: `curl https://across.to/api/available-routes | jq '.[]' | pbcopy`

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: The liquidity is managed on the hub pool (deployed only on the mainnet). This will be constant for that reason. However, the bridge config will have different id for the transfers as done here https://github.com/fortysevenlabs/messari-subgraphs/blob/across-v2/subgraphs/across-v2/src/handlers/transfers.ts#L34

fortysevenlabs:
IIRC, the build error was because of a deployment.json issue from the upstream. That may have been resolved but looks like we have a new deployment.json issue with the recent addition `Invalid network in deployment.json for protocol: evm deployment: evm-polygon-zkevm network: polygon-zkevm`. When thats fixed upstream, I'll sync my branch and the build error should be resolved.

subgraphs/across-v2/src/availableRoutesApi.ts
dmelotik: - Here is the [contract call](https://etherscan.io/address/0xc186fA914353c44b2E33eBE05f21846F1048bEda#readContract#F12) on mainnet. You need the L1 token address and destination ID, and it returns the destinationToken.
- And unfortunately it doesn't look like you can get the originToken on L2s :/ 

The current solution is good! Thanks for entertaining my idea

dmelotik: Feel free to resolve once you read this

dmelotik:
@fortysevenlabs that issue should be fixed locally when you upgrade the `messari-subgraph-cli` to the latest version. Github actions will use the latest version of that npm package, when I build locally I see small build issues with import paths.

lmk if that doesn't work for you locally

fortysevenlabs:
Makes sense on the SDK upgrade, didn't realize that happened. Updated and pushed the deployment.json changes.

> It seems like mainnet has an LP token factory where people can add liquidity to the protocol. In return I would think it gives out an LP token (outputToken) to track position.

@dmelotik  Unclear, are you suggesting adding something here? We already track the liquidity positions (see handler https://github.com/messari/subgraphs/blob/d4bf72d9034dc4c4b5af5ae1cb3337d16df48c11/subgraphs/across-v2/src/handlers/liquidity.ts). The hub pools stores all liquidity provider deposits.

dmelotik:
@fortysevenlabs 

> Unclear, are you suggesting adding something here?

Yeah, I am suggesting we utilize the outputToken fields in the `Pool` entity. See this transaction of liquidity being added [here](https://etherscan.io/tx/0x4ee6873e64fedccc0a3d32a7e64c3d929dd0fdb2a5fb986867247d9eb46341da).

You can see the LP token `Across V2 USDC LP Token`, from my perspective, that is the outputToken

fortysevenlabs:
> You can see the LP token `Across V2 USDC LP Token`, from my perspective, that is the outputToken

@dmelotik Yes, that is the LP token but we don't have access to the LP token in this (add/remove liquidity event). We could do this though:
1. We could get the output token through a [contract call to pooledTokens()](https://etherscan.io/address/0xc186fa914353c44b2e33ebe05f21846f1048beda#readContract) with l1Token as the input, lpToken as output
2. Set the outputTokenSupply:
    2.a Call `pool.addOutputTokenSupply(lpTokensMinted)` to add to the outputTokenSupply. This doesn't have an option to removeOutputTokenSupply when removing liquidity. Curious, why did we only add `addOutputTokenSupply()` and not `reduceOutputTokenSupply()`?
   2.b Make another contract call to the LP token address to get the totalSupply and then call `pool.setOutputTokenSupply(lpTokensMinted)` to update to the outputTokenSupply.

Best path seems to be 1 + 2.a and this would introduce two contract calls, good to go ahead with this approach?

dmelotik:
@fortysevenlabs For sure (1), this should only introduce a few new contract calls for each pool. Does not need to be repeated (so negligible performance).

- I would say we try 2a first (bc less contract calls), but if the resulting data is incorrect bc of interest (which I am unsure how they do that), then we need to use the contract call (2b) to ensure we have accurate data

fortysevenlabs:
@dmelotik 1 + 2.b is the ideal path (there was a typo, i said 2.a in previous message).

Irrespective of the interest, 2.a would result in inaccuracies because there isn't a way to reduce the output token supply on removing liquidity in bridge sdk. We have to use 2.b for that reason, right?

fortysevenlabs:
> I see the deployments have 0 values in the cumulative snapshots and negative TVL/revenue. Those will need to be fixed. And then can you redeploy all the networks?

Current set of todos
- [ ] Only the Boba deployment has 0 values because the prices lib doesn't support this chain.
- [ ] No TVL in Billions
- [ ] No Negative TVLs
- [ ] No negative TX counts
- [ ] Cumulatives shouldn't drop

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -61,8 +52,21 @@ export function handleLiquidityAdded(event: LiquidityAdded): void {
...
+  // pool
+  const poolId = event.address
+    .toHexString()
+    .concat(event.params.l1Token.toHexString());
+  const pool = sdk.Pools.loadPool<string>(Bytes.fromUTF8(poolId));
+
+  if (!pool.isInitialized) {
+    pool.initialize(token.name, token.symbol, BridgePoolType.LIQUIDITY, token);
+    pool.pool.outputToken = outputToken.id;
+    pool.pool.save();
```
jaimehgb, line 66: It doesn't make a difference in this case, but preferable do `pool.save()`.

The wrapper could potentially have some additional pre/post save logic that you'd be skipping :P

dmelotik: As a note assemblyscript doesn't respect access modifiers, so it will still compile and run.

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -72,6 +78,20 @@ export function handleLiquidityAdded(event: LiquidityAdded): void {
   const amount = event.params.amount;
   const account = sdk.Accounts.loadAccount(event.params.liquidityProvider);
   account.liquidityDeposit(pool, amount);
+
+  // tvl
+  let inputTokenBalance: BigInt;
+  const erc20 = _ERC20.bind(event.params.l1Token);
+  const inputTokenBalanceResult = erc20.try_balanceOf(event.address);
+  if (inputTokenBalanceResult.reverted) {
+    log.info(
```
dmelotik, line 87: To me this makes more sense as an error or critical log.

I am leaning towards critical bc it could throw off TVL A LOT if it reverts

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -72,6 +78,20 @@ export function handleLiquidityAdded(event: LiquidityAdded): void {
...
+      "[ERC20:balanceOf()]calculate token balance owned by bridge contract reverted",
+      []
+    );
+  } else {
+    inputTokenBalance = inputTokenBalanceResult.value;
+  }
+  pool.setInputTokenBalance(inputTokenBalance!);
```
dmelotik, line 94: We can't assume the inputTokenBalance here. In the reverted block we should make the default value 0, or initialize the variable to 0 to start

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -119,4 +144,18 @@ export function handleLiquidityRemoved(event: LiquidityRemoved): void {
   const amount = event.params.amount;
   const account = sdk.Accounts.loadAccount(event.params.liquidityProvider);
   account.liquidityWithdraw(pool, amount);
+
+  // tvl
```
dmelotik, line 148: since this code block is copy/pasted we should make it it's own function

dmelotik:
A few notes:
- We should fix the non-human readable `pool.name`s
- We can derive outputTokenPriceUSD
- You can also stake your LP tokens so we should implement the logic to follow that field
- Rewards doesn't look like it is working

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -0,0 +1,161 @@
...
+  const poolId = event.address
+    .concat(event.params.l1Token)
+    .concat(Bytes.fromUTF8("liquidity"));
+  const pool = sdk.Pools.loadPool<string>(poolId);
+
+  if (!pool.isInitialized) {
+    pool.initialize(
+      token.name,
+      token.symbol,
+      BridgePoolType.LOCK_RELEASE,
```
dmelotik, line 134: Should this be liquidity?

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -0,0 +1,161 @@
...
```
dmelotik, line 68: same here?

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: Ah, didn't know. Why do we use access modifiers in SDK then? For readability?

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: fair point. will update to critical

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: agreed

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: accidental. liquidity should have liquidity pool type and transfers should have lock/release pool type

bye43:
Ahh yes @fortysevenlabs we do not count staking of the protocol's native token as part of `TVL`, but we need to track the `stakedOutputToken` or staked lp tokens, as we use that to derive the `rewardAPR`

fortysevenlabs:
Related

https://github.com/messari/subgraphs/issues/1711

https://github.com/messari/subgraphs/issues/2165

fortysevenlabs:
@bye43 Should we consider across-v2 done?

fortysevenlabs:
@dmelotik can we mark across-v2 done?

dmelotik:
I will finish up my code review next week @fortysevenlabs

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -0,0 +1,167 @@
...
+      "[_ERC20:tokenSupply()] retrieve outputTokenSupply for LP pool call reverted",
+      []
+    );
+  } else {
+    outputTokenSupply = outputTokenSupplyResult.value;
+    pool.setOutputTokenSupply(outputTokenSupply);
+  }
+
+  // output token price
+  pool.pool.outputTokenPriceUSD = token.lastPriceUSD;
```
dmelotik, line 71: Shouldn't output token price be TVL / output token supply?

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -0,0 +1,167 @@
...
```
dmelotik, line 146: same

subgraphs/across-v2/src/handlers/liquidity.ts
fortysevenlabs: Prices for output (LP) tokens aren't available. Since across's LP token minted/burned is 1:1 with the input token deposited/withdrawn, we can use the input token price as a substitute for output token price.

fortysevenlabs: Prices for output (LP) tokens aren't available and there were no contract read methods for fetching the exchangeRate. Since across's LP token minted/burned is almost 1:1 with the input token deposited/withdrawn, I went ahead with the input token price as a substitute for output token price.

fortysevenlabs: Thinking further, we could derive the exchangeRate (`exchangeRate = l1Tokens / lpTokensMinted`) and then multiple input token price to derive the output token price.

fortysevenlabs: I had ignored exchangeRateCurrent() method earlier because this was a write contractual method. I didn't realize we could call write methods (this one is non-payable type) in a subgraph. So, we do have two options now: a) `exchangeRate = l1Tokens / lpTokensMinted` b) call `exchangeRateCurrent()`. 

Verified they return same amounts upto 15 decimals. I'll go ahead with (a) to avoid contract calls.

fortysevenlabs: What do you think [about this change](https://github.com/messari/subgraphs/pull/1777/commits/aa46df830473bec3653effadad063f5c05e6e80e)?

subgraphs/across-v2/src/handlers/liquidity.ts
```diff
@@ -68,7 +70,12 @@ export function handleLiquidityAdded(event: LiquidityAdded): void {
   }
 
   // output token price
-  pool.pool.outputTokenPriceUSD = token.lastPriceUSD;
+  const exchangeRate = bigIntToBigDecimal(event.params.amount).div(
+    bigIntToBigDecimal(event.params.lpTokensMinted)
+  );
+  pool.pool.outputTokenPriceUSD = tokenPricer
+    .getTokenPrice(token)
```
fortysevenlabs, line 77: Had to do use tokenPricer instead of `token.lastPriceUSD` because it wasn't available (null wasm errors). Looking thru the SDK, seems like token price should be set and saved when adding volume or refreshing TVL but I don't see that happening here.

subgraphs/across-v2/src/handlers/liquidity.ts
dmelotik: This might also work. I was imagining 

```typescript
market.outputTokenPriceUSD = market.inputTokenBalanceUSD / market.outputTokenBalance
```

That should be the same thing, and no need to make a contract call for the exchange rate.

fortysevenlabs: I see. We should be good to go with the current implementation. Doesn't make contract call and avoids relying on data calculated by sdk (observed null wasm issues with the input token price).

dmelotik (approved):
Awesome work @fortysevenlabs ! Code looks good on my end

PR #1786, opened 2023-02-22
feat(#major); spark lend; new subgraph

dmelotik:
This PR is for Spark lend. Spark is an aave fork made by makerdao. Right now they only have testnet deployments, so we mocked it up, and will fill in the mainnet deployments when it is complete.

For now you can see the test subgraph here: https://okgraph.xyz/?q=dmelotik%2Fspark-lend-georli

### DIFF
--- a/subgraphs/aave-forks/protocols/spark-lend/src/mapping.ts
+++ b/subgraphs/aave-forks/protocols/spark-lend/src/mapping.ts
@@ -1,0 +1,422 @@
+import {
+  Address,
+  BigDecimal,
+  BigInt,
+  log,
+} from "@graphprotocol/graph-ts";
+import { PriceOracleUpdated } from "../../../generated/LendingPoolAddressesProvider/LendingPoolAddressesProvider";
+import {
+  getNetworkSpecificConstant,
+  Protocol,
+} from "./constants";
+import {
+  BorrowingDisabledOnReserve,
+  BorrowingEnabledOnReserve,
+  CollateralConfigurationChanged,
+  ReserveActivated,
+  ReserveDeactivated,
+  ReserveFactorChanged,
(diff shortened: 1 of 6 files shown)


subgraphs/aave-forks/protocols/spark-lend/src/mapping.ts
```diff
@@ -0,0 +1,659 @@
...
+  market.save();
+}
+
+function multiArraySort(
+  ref: Array<string>,
+  arr1: Array<BigInt>,
+  arr2: Array<BigDecimal>
+): void {
+  if (ref.length != arr1.length || ref.length != arr2.length) {
+    // cannot sort
```
tnkrxyz, line 638: maybe log an error instead of return silently?

tnkrxyz (approved):
looks good! approved with minor comments.

subgraphs/aave-forks/protocols/spark-lend/src/mapping.ts
```diff
@@ -0,0 +1,659 @@
...
+  const assetToken = Token.load(assetAddress);
+  if (!assetToken || !assetToken._market || !assetToken._type) {
+    log.error(
+      "[handleAssetConfigUpdated]Failed to find token {} or assetToken._market/assetToken._type is null",
+      [assetAddress]
+    );
+    return;
+  }
+
+  const marketId = Address.fromString(assetToken._market!);
```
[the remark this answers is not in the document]
dmelotik: good point!

subgraphs/aave-forks/protocols/spark-lend/src/mapping.ts
dmelotik: good call

PR #1796, opened 2023-02-24
feat(#major); base network; add subgraph

dmelotik:
Add base goerli network subgraph.

Not sybcing with the network slug `base-goerli`

Test: https://okgraph.xyz/?q=QmXfhYAPmAv2XTngJ7x3GioRwuAA6YmRf97HDp4CxwguDN

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2732,7 +2732,7 @@
         }
       },
       "evm-base": {
-        "network": "ethereum",
+        "network": "base-testnet",
         "status": "dev",
         "versions": {
           "schema": "1.1.1",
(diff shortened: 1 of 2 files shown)


deployment/deployment.json
```diff
@@ -2731,6 +2731,28 @@
           }
         }
       },
+      "evm-base": {
```
steegecs, line 2734: Do you want to call this `evm-base-testnet`?

dmelotik: Maybe we just leave base out until it is mainnet. I don't see much of a point to have a testnet deployment rn. I will make it a draft again

dmelotik:
deprecated network subgraphs

PR #1798, opened 2023-02-24
Support for Protocol cumulativeLiquidity Withdraw & Deposit Count

itsjerryokolo:
Without this, the cumulative liquidity deposit and withdraw count isn't automatically updated.


jaimehgb:
Aren't these here: ?https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts#L666
https://github.com/messari/subgraphs/blob/master/subgraphs/_reference_/src/sdk/protocols/bridge/pool.ts#L679

If you want to add them to Account you should also remove them from Pool to not count them twice

itsjerryokolo:
Ah okay. That's good to know. I'll do that
My version didn't have that. 
Screenshot 2023-02-24 at 4 02 48 PM

PR #1804, opened 2023-02-26
chore (); makerdao-ethereum; prettier formated code using default config

tnkrxyz:
This PR re-formatted makerdao subgraph code & abis using the default prettier config (it was formatted using a different config specific to the subgraph).


dmelotik (approved):
looks good

Issue #1415, opened 2022-11-15
Add Bastion Realms


this-username-is-taken:
For example: https://app.bastionprotocol.com/realms/0xe1cf09BDa2e089c63330F0Ffe3F6D6b790835973

dmelotik:
2 approaches: each one is a compound fork OR we add all of them onto the current bastion subgraph.

bye43:
Deployment: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/bastion-protocol-aurora&tab=poolOverview&overlay=https://api.thegraph.com/subgraphs/name/messari/bastion-protocol-aurora

@tnkrxyz Couple QA comments:

- [ ] Looks like `0x08ac1236ae3982ec9463efe10f0f320d9f5a9a4b` is for [staking `BSTN`](https://withdraw.bastionprotocol.com/kingdom/staking), so I believe we can remove this market since we do not usually count staked tokens as part of TVL. @dmelotik double checking this correct?
[...]

PR #1805, opened 2023-02-26
feat (#minor); bastion-protocol-aurora; add bastion realms

tnkrxyz:
This PR adds bastion realms (isolated lending) to the bastion subgraph, as requested in #1415

- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/bastion-protocol-aurora
- dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/bastion-protocol-aurora&tab=protocol


bye43:
@tnkrxyz Just took a brief glance at this and was wondering if we were adding bastion realms to the existing bastion subgraph or if this is a separate subgraph. If its the former, looks like the TVL on the Realms subgraph is lower than the current Bastion subgraph.

tnkrxyz:
> @tnkrxyz Just took a brief glance at this and was wondering if we were adding bastion realms to the existing bastion subgraph or if this is a separate subgraph. If its the former, looks like the TVL on the Realms subgraph is lower than the current Bastion subgraph.

I am adding them to the existing subgraph, as Dylan suggested. Let me look into the TVL.

deployment/deployment.json
```diff
@@ -1220,8 +1220,8 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.1.9",
-          "methodology": "1.0.0"
+          "subgraph": "1.2.0",
```
dmelotik, line 1223: I would say this is major, what do you think? A critical change

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -117,14 +119,23 @@ export function handleMarketListed(event: MarketListed): void {
     cTokenContract.try_reserveFactorMantissa(),
     BIGINT_ZERO
   );
+  let marketNamePrefix = "";
+  if (event.address.equals(AURORA_REALM_ADDRESS)) {
+    marketNamePrefix = "Aurora Ream: ";
```
dmelotik, line 124: typos, should be `Realm`

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -254,7 +254,8 @@ export function _handleMarketEntered(
 
 export function _handleMarketListed(
```
dmelotik, line 255: editing this breaks the build for all other compound forks. let's leave this function alone, but just overwrite the `market.name` in the bastion mapping code

subgraphs/compound-forks/protocols/bastion-protocol/README.md
```diff
@@ -1,6 +1,8 @@
 # Bastion Protocol Subgraph
 
-## Calculation Methodology v1.0.0
+## Calculation Methodology v1.0.1
```
dmelotik, line 3: One of the downsides of this approach with adding the other realms is that the position data is expected to be pooled across the whole protocol, but now a position in a realm is siloed to that realm.

What I am referring to here is that compound positions are pooled together to get a liquidity in the protocol, where this is now siloed to each realm.

deployment/deployment.json
tnkrxyz: Either seems fine to me. Since the realms are all much smaller, one may see them as auxiliary and adding them is "minor"?

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
tnkrxyz: Good catch! Fixed.

subgraphs/compound-forks/src/mapping.ts
tnkrxyz: I set an empty default marketNamePrefix argument for `_handleMarketListed`, so this change will not break other compound forks. But I see your point, and changed to modify market.name afterwards, even though a bit redundant.

subgraphs/compound-forks/protocols/bastion-protocol/README.md
tnkrxyz: Yes, that's true. At position level, it is hard to see which positions are pooled, which are isolated. An alternative is to have a separate subgraph for each realm - not sure that's much better (a lot duplicated code).

tnkrxyz:
> > @tnkrxyz Just took a brief glance at this and was wondering if we were adding bastion realms to the existing bastion subgraph or if this is a separate subgraph. If its the former, looks like the TVL on the Realms subgraph is lower than the current Bastion subgraph.
> 
> I added them to the existing subgraph, as Dylan suggested. It seems that there are a few markets (ETH, USDT, and STNEAR) with 0 TVL, likely due to incorrectly price oracle data. Looking into this.

I can confirm this is due to calls to price oracle get reverted for those tokens. I added an alternative to get price from the prices lib when this happens. However, I am getting an error "Subgraph instance failed to run: A matching Ethereum network with NodeCapabilities { archive: true, traces: true } was not found" when deploying bastion subgraph to hosted service since yesterday (no problem deploying other subgraphs), so I could not test my fix yet. Will keep trying.

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -125,6 +153,9 @@ export function handleMarketListed(event: MarketListed): void {
       cTokenReserveFactorMantissa
     );
     _handleMarketListed(marketListedData, event);
+    const market = Market.load(cTokenAddr.toHexString())!;
```
dmelotik, line 156: It is possible there is an early return in _handleMarketListed() so we should error check this val

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -157,6 +188,10 @@ export function handleMarketListed(event: MarketListed): void {
     ),
     event
   );
+
+  const market = Market.load(cTokenAddr.toHexString())!;
```
dmelotik, line 192: why do we need this 2x?

dmelotik: If we only need it once then we would probably want to put the naming if block in the same if block to reduce the scope of those vars

subgraphs/compound-forks/protocols/bastion-protocol/README.md
dmelotik: Yep, we considered both options. I agree that would be worse as it seems strange to have 4 bastion subgraphs, I am just asking that we make note of the limitation to remain as transparent as possible.

dmelotik:
Which markets were having a hard time with pricing? Each realm has it's own oracle I see from the docs, are the contract calls to them not working?

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
tnkrxyz: I would say we should crash if the market is not successfully created in the line above (early return); if not here, certainly later somewhere else.

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
tnkrxyz: We need both, because the earlier chunk is for native token, and it returns at [line 159](https://github.com/tnkrxyz/subgraphs/blob/abbf59ee3cc00a74c4a74f467b8e62aa9fa2d7db/subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts#L159)

tnkrxyz:
> Which markets were having a hard time with pricing? Each realm has it's own oracle I see from the docs, are the contract calls to them not working?

For both Hub and Realm, oracle calls for ETH, WBTC, USDT, STNear reverted and ended up with prices of 0 (https://api.thegraph.com/subgraphs/name/tnkrxyz/bastion-protocol-aurora/graphql?query=query+MyQuery+%7B%0A++lendingProtocols+%7B%0A++++id%0A++++_priceOracle%0A++%7D%0A++markets+%7B%0A++++name%0A++++inputToken+%7B%0A++++++name%0A++++%7D%0A++++inputTokenBalance%0A++++inputTokenPriceUSD%0A++++%0A++%7D%0A++%0A%7D). 

What is odd is that it seemed to work in earlier messari/hosted service deployment: https://api.thegraph.com/subgraphs/name/messari/bastion-protocol-aurora/graphql?query=query+MyQuery+%7B%0A++lendingProtocols+%7B%0A++++id%0A++++_priceOracle%0A++%7D%0A++markets+%7B%0A++++name%0A++++inputToken+%7B%0A++++++name%0A++++%7D%0A++++inputTokenBalance%0A++++inputTokenPriceUSD%0A++++%0A++%7D%0A++%0A%7D

dmelotik:
Right now it looks like the error you are getting is just because the abis for the price lib are just not added to the manifest

tnkrxyz:
> Right now it looks like the error you are getting is just because the abis for the price lib are just not added to the manifest

Right, but a rather odd one. I already have CurveRegistry abi in my yaml, not sure why it still complains.

dmelotik:
@tnkrxyz I realized this is happening bc the price oracle is different for each realm and it is updating the 1 field in `lendingPool.priceOracle`. We will have to have a price oracle field for each realms so that it knows which one to use. If you look at the 2 queries you are showing the price oracle address is different.

That is why it seems like only 1 realm is priced at a time.

I think it should work if we use the proper price oracles and then we can remove the price lib

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
dmelotik: At both instances in this function can we do an error handling like this: https://github.com/messari/subgraphs/blob/5b3b0c500856a8e79c481ddfb3068ce4413efc80/subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts#L382

we can do a `log.critical` error in order to crash the subgraph

For more context, we are in the process of planning to improve our data quality. One of the pieces of this is not silently failing and not crashing to create downtime. We will begin to raise errors as they happen without silently failing like rn.

Either way, I am going to be strict on proper error handling. `log.critical` is just easier to find in code than a quiet `!`

tnkrxyz:
Ok, the price oracle is now working as expected: https://api.thegraph.com/subgraphs/id/Qmaw6RPYGDY1sZ4yxFBVSwx8CYexycvH2kj6kFsgnBWihM/graphql?query=query+MyQuery+%7B%0A++lendingProtocols+%7B%0A++++id%0A++++_priceOracle%0A++%7D%0A++markets+%7B%0A++++name%0A++++inputToken+%7B%0A++++++name%0A++++%7D%0A++++inputTokenBalance%0A++++inputTokenPriceUSD%0A++%7D%0A++%0A%7D (sync in progress). 

I do need to add a _stNearPriceOracle field to keep oracle for stNear token. It may be a good idea to have the _priceOracle field in the Market entity, but I will keep it as it is for now.

subgraphs/compound-forks/abis/prices/Curve/Pool.json
```diff
@@ -0,0 +1,916 @@
+[
```
dmelotik, line 1: is this file still necessary?

dmelotik:
> Ok, the price oracle is now working as expected

Nice @tnkrxyz , one option I was thinking about is to have a new helper entity (one for each realm) and it stores the price oracle so you can load it and use it each time you get price in a given realm. But I am indifferent to the solution here as long as it works properly.

subgraphs/compound-forks/src/mapping.ts
```diff
@@ -2395,6 +2395,20 @@ export function updateAllMarketPrices(
     }
 
     // update market price
+    const result = priceOracle.try_getUnderlyingPrice(
```
dmelotik, line 2398: is this for debugging?

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -66,21 +66,38 @@ import { Comptroller } from "../../../generated/Comptroller/Comptroller";
...
+    const newPriceOracle = event.params.newPriceOracle;
+    log.info("[handleNewPriceOracle] price oracle {} => {}", [
+      protocol._priceOracle,
+      newPriceOracle.toHexString(),
+    ]);
+    _handleNewPriceOracle(protocol, newPriceOracle);
+    return;
+  }
+
+  if ((event.address = STNEAR_REALM_ADDRESS)) {
```
dmelotik, line 97: `if (event.address == STNEAR_REALM_ADDRESS) {

tnkrxyz:
> Also it looks like the aurora price is still 0...
> 
> Also, don't you need a separate price oracle for each of the 4 realms?

I had thought I can get by without having a separate oracle for each realm, looks like I do have to have it.

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
tnkrxyz: Good catch!

dmelotik (approved):
Looks great! Nice work @tnkrxyz

tnkrxyz:
> @dmelotik and @tnkrxyz , can we merge this in? It's been 2 weeks since approval.

I am hoping @dmelotik can re-review the new commit and comment on a remaining issue in [#1415](https://github.com/messari/subgraphs/issues/1415#issuecomment-1475407997)

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -402,192 +448,206 @@ function getOrCreateProtocol(): LendingProtocol {
...
-  const rewardTokens: string[] = [];
-  const rewardEmissions: BigInt[] = [];
-  const rewardEmissionsUSD: BigDecimal[] = [];
+export function handleRewardAddressChanged(event: RewardAddressChanged): void {
+  const rewardTypeId = event.params.rewardType.toString();
+  let rewardToken = _RewardType.load(rewardTypeId);
+  if (!rewardToken) {
+    rewardToken = new _RewardType(rewardTypeId);
+  }
+  rewardToken.token = event.params.newRewardAddress.toHexString();
```
dmelotik, line 468: we should make sure the token exists before assigning it

subgraphs/compound-forks/protocols/bastion-protocol/src/mapping.ts
```diff
@@ -402,192 +448,206 @@ function getOrCreateProtocol(): LendingProtocol {
...
+  getOrCreateMarketDailySnapshot(
+    market,
+    event.block.timestamp,
+    event.block.number
+  );
+  getOrCreateMarketHourlySnapshot(
+    market,
+    event.block.timestamp,
+    event.block.number
+  );
```
dmelotik, line 637: This function is called from `handleAccrueInterest()` which already creates snapshots. I think this is unnecessary

dmelotik (commented):
Changes look good!

- For the `market.isActive` issue, that is really weird to me since it should be using the same logic as compound, but it is def wrong since there are deposits/borrows actively going on. So, yes, could you override those functions and make a note! Thanks

PR #1814, opened 2023-03-01
fix(#minor); Saddle && Platypus; Fix compatibility with the required fields from graph version update.

steegecs:
**Context:**
Fix subgraph as there was a backward compatibility break from when I upgraded the new CLI version. The fix was very simple. I just had to fix assertions for some non-required fields on the Token entity. There were a few instances where I just had. to check if the field existed because running another condition on the value. 

Second, I noticed in Platypus to null return values from the price lib were given a value of `1`. It makes more sense to set the price to 0 if no price is returned. It is possible that setting these null values to 1 could throw off our pricing if the actual cost of the token is much lower.

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -3530,7 +3530,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.0",
-          "subgraph": "1.1.6",
+          "subgraph": "1.1.7",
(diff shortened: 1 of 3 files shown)


dmelotik (approved):
LGTM, @steegecs could you make another PR to format and lint these 2 directories since we are touching them?

PR #1816, opened 2023-03-01
chore(); Saddle Finance and Playpus Finance; Linting

steegecs:
**Context:**
Recently made a PR for Playpus and Saddle Finance to fix backward compatibility issues with new graph CLI version. I am not sure exactly which version caused those issues, as I have not deployed these subgraphs for a long time --> #1814

In response to this PR, I am linting these subgraphs to fix any of the CI linting errors that popped up.


dmelotik (approved):
nice, approved

PR #1825, opened 2023-03-05
Fix(#minor); Breakdown open interest into Long and short open interest in Perp schema

ishraq8:
broke down open interest into Long Open interest and short open interest. The granular breakdown of the metrics makes it more useful for research. 

Open to discussion, will tag on GMX pull request and SDK pull request.

### DIFF
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -1,5 +1,5 @@
 # Subgraph Schema: Derivatives Perpetual Futures
-# Version: 1.1.0
+# Version: 1.2.0
 # See https://github.com/messari/subgraphs/blob/master/docs/SCHEMA.md for details
 
 enum Network {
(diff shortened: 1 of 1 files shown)


ishraq8:
re added PCV as removing it may cause errors in deployment

schema-derivatives-perpfutures.graphql
```diff
@@ -345,8 +351,14 @@ type DerivPerpProtocol implements Protocol @entity {
   " Number of cumulative liquidatees (accounts that got liquidated) "
   cumulativeUniqueLiquidatees: Int!
 
+" Notional value of all open long positions "
```
dmelotik, line 354: watch the formatting here

dmelotik:
> re added PCV as removing it may cause errors in deployment

I don't see PCV added in this PR. I don't necessarily think we should add it if it is just to prevent errors. Do perp protocols have PCV and do we want to track it?

schema-derivatives-perpfutures.graphql
ishraq8: fixed good catch

ishraq8:
> > re added PCV as removing it may cause errors in deployment
> 
> I don't see PCV added in this PR. I don't necessarily think we should add it if it is just to prevent errors. Do perp protocols have PCV and do we want to track it?

they don't i removed it from the schema in this PR. hopefully no issues, but if anything comes up we can make short fixes. @dmelotik feel free to suggest changes, and once done i can merge. Thanks for the input!

dmelotik (approved):
lgtm

PR #1828, opened 2023-03-06
feat(#minor): Fix Options schema problem happening during compilation and deployment

corerouter:
Some small changes to Options schema in order to compile and deploy related dopex subgraph.

1. Some changes are needed as the errors happening when compiling, deploying and running subgraphs otherwise.
2. Add PCV as the subgraph dashboard will complain it does not comply with generic schema otherwise.
3. Add two extra kinds of fee type to be same with Perp schema.
4. Modify some entry names to be compliant with naming convention.

### DIFF
--- a/schema-derivatives-options.graphql
+++ b/schema-derivatives-options.graphql
@@ -87,25 +87,31 @@ enum LiquidityPoolFeeType {
   " Some protocols use tiered fees instead of fixed fee (e.g. DYDX). Set `feePercentage` as 0 but handle the tiered fees in the mapping code. "
   TIERED_TRADING_FEE
 
-  " Some protocols use dynamic fees instead of fixed fee. Set `feePercentage` as 0 but handle the dynamic fees in the mapping code. "
+  " Some protocols use dynamic fees instead of fixed fee (e.g. Balancer v2). Set `feePercentage` as 0 but handle the dynamic fees in the mapping code. "
   DYNAMIC_TRADING_FEE
 
-  " Fixed fee that's paid to the LP, as a percentage of the traded amount."
+  " Fixed fee that's paid to the LP, as a percentage of the traded amount. e.g. 10% of fees to GMX LP providers (GLP holders). "
   FIXED_LP_FEE
 
-  " Some protocols use dynamic LP fees. Set `feePercentage` as 0 but handle the dynamic fees in the mapping code. "
+  " Some protocols use dynamic LP fees . Set `feePercentage` as 0 but handle the dynamic fees in the mapping code. "
   DYNAMIC_LP_FEE
 
-  " Fixed fee that's paid to the protocol, as a percentage of the traded amount. e.g. 0.05% for Sushiswap, 0.02% for Curve v1. "
+  " Fixed fee that's paid to the protocol, as a percentage of the traded amount. e.g. 10% of GMX fees to protocol. "
   FIXED_PROTOCOL_FEE
 
   " Some protocols use dynamic protocol fees. Set `feePercentage` as 0 but handle the dynamic fees in the mapping code. "
   DYNAMIC_PROTOCOL_FEE
 
-  " One-time fee charged by the protocol during deposit, in percentages of the deposit token "
+  " Fixed fee that's paid to the stakers, as a percentage of the traded amount. e.g. 15% of all fees on GMX to stakers. "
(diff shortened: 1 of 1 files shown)


ishraq8:
good additions, thank you

schema-derivatives-options.graphql
```diff
@@ -1015,6 +1030,12 @@ type LiquidityPoolHourlySnapshot @entity(immutable: true) {
   " Hourly Deposit trade notional volume occurred in this pool for a specific input token, in USD. The ordering should be the same as the pool's `inputTokens` field. "
   hourlyDepositVolumeByTokenUSD: [BigDecimal!]!
 
+  " All Inflow trade notional volume occurred in this pool for a specific input token, in native amount. The ordering should be the same as the pool's `inputTokens` field. "
+  _cumulativeDepositVolumeByTokenAmount: [BigInt!]!
```
dmelotik, line 1034: Why do we want to add this as a helper field? If we add this here there should be a counterpart in `LiquidityPool`

schema-derivatives-options.graphql
```diff
@@ -1245,9 +1278,17 @@ type ActiveAccount @entity {
   id: Bytes!
 }
 
+enum OptionType {
```
dmelotik, line 1281: let's place this at the beginning where all the other enums are

dmelotik:
This technically is a #major api change since it will make incompatible changes, but let's keep it as minor since we have not actually implemented this in production yet. Could you also make the proper version change at the top of the file?

schema-derivatives-options.graphql
```diff
@@ -154,6 +160,9 @@ interface Protocol {
   " Current TVL (Total Value Locked) of the entire protocol "
   totalValueLockedUSD: BigDecimal!
 
+  " Current PCV (Protocol Controlled Value). Only relevant for protocols with PCV. "
+  protocolControlledValueUSD: BigDecimal
```
dmelotik, line 164: Is there a reason we want this aside from making it work in subgraphs.xyz?

This is the case bc the options schema is not integrated in subgraphs.xyz yet. It is defaults to look at the perps schema. This is being fixed by Michael C though

corerouter: Without protocolControlledValueUSD, subgraphs.xyz complains the schema does not comply with generic schema. Now with protocolControlledValueUSD, subgraphs.xyz looks like it doesn't support options schema yet. Nevertheless, it is added to just make it works with subgraphs.xyz.

schema-derivatives-options.graphql
corerouter: Fixed at 5b75305b341ea53c57e08fee4dab7eb7420db7e9 I may add them by mistake and they don't need in the schema template.

schema-derivatives-options.graphql
corerouter: Changed at 5b75305b341ea53c57e08fee4dab7eb7420db7e9

corerouter:
> This technically is a #major api change since it will make incompatible changes, but let's keep it as minor since we have not actually implemented this in production yet. Could you also make the proper version change at the top of the file?

@dmelotik Changed the schema contents and versions based on reviews . Thanks!

schema-derivatives-options.graphql
dmelotik: @ishraq8 do we have a need for PCV in the options schema?

@corerouter What I am getting at is that we should not add it reactively just to make subgraphs.xyz happy. Whatever we decide, Michael C said he would work on adding the options schema into subgraphs.xyz, so we don't have to worry about that.

dmelotik:
@corerouter quick note, minor is the second version (https://semver.org/) so it should be `1.2.0`

schema-derivatives-options.graphql
corerouter: I see. Just restore PCV part to previous versions before I made the changes at 7cee2193120a9b72b80495e89f048f6a5bcc7cb3 @dmelotik @ishraq8

corerouter:
> @corerouter quick note, minor is the second version (https://semver.org/) so it should be `1.2.0`
Made the changes at 7cee2193120a9b72b80495e89f048f6a5bcc7cb3 Thanks!

jaimehgb (approved):

Issue #1665, opened 2023-01-31
Build Opyn Subgraph 


bye43:
```[tasklist]
### Deployment Status
- [ ] https://github.com/messari/subgraphs/issues/1981
- [x] Arbitrum: https://okgraph.xyz/?q=spitko/opyn-gamma-arbitrum
- [x] Avalanche: https://okgraph.xyz/?q=spitko%2Fopyn-gamma-avalanche
- [x] Polygon: https://okgraph.xyz/?q=spitko%2Fopyn-gamma-polygon
```

bye43:
@spitko

bye43:
Waiting for: https://github.com/messari/subgraphs/issues/1963 to merge before QA

bye43:
Ethereum QA: https://github.com/messari/subgraphs/issues/1981

bye43:
@ishraq8 Looks like ETH was good! If you can give the rest of these deployments a quick scan, then we can close out Opyn as done/`prod`

ishraq8:
<img width="805" alt="Screen Shot 2023-04-12 at 2 24 11 PM" src="https://user-images.githubusercontent.com/47088402/231550419-2de91e36-0917-45e6-b222-e3a7b930f510.png">

Getting this error in okgraph, looks like related to new entities in schema @spitko @dmelotik

spitko:
> Getting this error in okgraph, looks like related to new entities in schema @spitko @dmelotik

Redeployed, should be fine now.

ishraq8:
Just a couple of questions @spitko from the QA:

1) TVL looks a bit off compared to defillama. 
<img width="477" alt="Screen Shot 2023-04-14 at 9 31 43 AM" src="https://user-images.githubusercontent.com/47088402/232057832-ebb769bf-f4fe-44b4-a982-20
<img width="475" alt="Screen Shot 2023-04-14 at 9 32 11 AM" src="https://user-images.githubusercontent.com/47088402/232057884-58519e1c-f812-470b-a515-cefc97f02564.png">
dbe24ed4c4.png">
<img width="377" alt="Screen Shot 2023-04-14 at 9 32 26 AM" src="https://user-images.githubusercontent.com/47088402/232057900-98cde4c9-faa0-4d9d-ad2d-0b289d00e8e3.png">

2) What is the logic behind null revenue? Is it because all options are cash settled and basically value of underlying is just transferred between parties. Just want to confirm.

[...]

PR #1832, opened 2023-03-07
feat(#major); opyn-gamma; Add Opyn Gamma subgraph

spitko:
#1665


dmelotik:
@spitko apologies to update the schema while building, but we made a few small changes to the schema: https://github.com/messari/subgraphs/pull/1828

subgraphs/opyn-gamma/schema.graphql
```diff
@@ -0,0 +1,1395 @@
...
+  dailyDepositVolumeUSD: BigDecimal!
+
+  " Daily Deposit trade notional volume for a specific input token, in native amount. The ordering should be the same as the pool's `inputTokens` field. "
+  dailyDepositVolumeByTokenAmount: [BigInt!]!
+
+  " Daily Deposit trade notional volume for a specific input token, in USD. The ordering should be the same as the pool's `inputTokens` field. "
+  dailyDepositVolumeByTokenUSD: [BigDecimal!]!
+
+  " Deposit historical trade notional volume occurred in this pool, in USD "
+  cumulativeDepositVolumeUSD: BigDecimal!
```
jaimehgb, line 885: Can you keep the original naming of these fields? Since there is another options protocol in the works, to keep them consistent.

dmelotik:
@spitko can you add your test deployments?

spitko:
> @spitko can you add your test deployments?

Sorry, here you go:

https://okgraph.xyz/?q=spitko%2Fopyn-gamma-ethereum
https://okgraph.xyz/?q=spitko%2Fopyn-gamma-arbitrum
https://okgraph.xyz/?q=spitko%2Fopyn-gamma-avalanche
https://okgraph.xyz/?q=spitko%2Fopyn-gamma-polygon

Looks like the avalanche deployment failed, will check it out. Others seem to have fully indexed 👍

subgraphs/opyn-gamma/src/entities/usage.ts
dmelotik, line 1: The string literals in this file should be constants in constants.ts. Since the constants are used in multiple places. If we ever changed them in the future we would only need to update 1 location in constants.ts

subgraphs/opyn-gamma/src/mappings/OToken.ts
```diff
@@ -0,0 +1,34 @@
...
+} from "../entities/account";
+import { burnOption, mintOption } from "../entities/option";
+import { getOrCreatePool } from "../entities/pool";
+import { updatePosition } from "../entities/position";
+import { takeSnapshots } from "../entities/snapshot";
+
+export function handleTransfer(event: TransferEvent): void {
+  const option = Option.load(event.address)!;
+  const amount = event.params.value;
+  takeSnapshots(event, getOrCreatePool(getOrCreateToken(option.pool)));
```
dmelotik, line 17: take snapshots at the end of this handler?

subgraphs/opyn-gamma/src/price.ts
```diff
@@ -0,0 +1,116 @@
...
+  );
+  const priceResult = oracle.try_getPrice(Address.fromBytes(token.id));
+  if (!priceResult.reverted) {
+    const price = bigIntToBigDecimal(priceResult.value, INT_EIGHT);
+    token.lastPriceBlockNumber = event.block.number;
+    token.lastPriceUSD = price;
+    token.save();
+    return price;
+  }
+  log.error("Failed to get price for asset: {}, trying with price oracle", [
```
dmelotik, line 32: Probably better as a warning: https://github.com/messari/subgraphs/pull/1960

dmelotik, line 32: This makes more sense as a warning to me: https://github.com/messari/subgraphs/pull/1960

subgraphs/opyn-gamma/src/entities/pool.ts
```diff
@@ -0,0 +1,236 @@
...
+  return pool;
+}
+
+export function handlePoolDeposit(
+  event: ethereum.Event,
+  pool: LiquidityPool,
+  deposit: Deposit
+): void {
+  const amount = deposit.inputTokenAmounts[0];
+  pool.inputTokenBalances = [pool.inputTokenBalances[0].plus(amount)];
```
dmelotik, line 99: For inputTokenBalance would it be more accurate to use this contract call (https://etherscan.io/address/0x5934807cC0654d46755eBd2848840b616256C6Ef#readContract#F3) ? There is a difference in the values from what I see

dmelotik (commented):
Nice work! Only a few comments

When this (https://github.com/messari/subgraphs/pull/1967) merges, can you update to use this?

Apologies for taking so long to review

subgraphs/opyn-gamma/src/entities/event.ts
```diff
@@ -0,0 +1,90 @@
...
+export function createDeposit(
+  event: ethereum.Event,
+  asset: Address,
+  amount: BigInt,
+  from: Address
+): void {
+  const account = getOrCreateAccount(from);
+  const token = getOrCreateToken(asset);
+  const pool = getOrCreatePool(token);
+  takeSnapshots(event, pool);
```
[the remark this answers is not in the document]
spitko: This is how snapshots are done in the sdk. It's not perfect and the daily fields are not completely accurate. But doing it after would still miss any deposits made on the same day after the snapshot is taken, since the snapshots are immutable now.

subgraphs/opyn-gamma/src/entities/event.ts
```diff
@@ -0,0 +1,90 @@
...
+  const deposit = new Deposit(
+    Bytes.fromUTF8(
```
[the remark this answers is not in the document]
spitko: Schema had the prefixes in the ID comments. I can remove the prefixes.

subgraphs/opyn-gamma/src/entities/position.ts
```diff
@@ -0,0 +1,120 @@
...
+  position.account = account.id;
+  position.asset = option.underlyingAsset;
+  position.takenHash = event.transaction.hash;
+  position.takenBlockNumber = event.block.number;
+  position.takenTimestamp = event.block.timestamp;
+  position.takenPrice = getUnderlyingPrice(event, option);
+  position.premium = BIGINT_ZERO;
+  position.premiumUSD = BIGDECIMAL_ZERO;
+  position.amount = amount;
+  position.amountUSD = bigIntToBigDecimal(amount, 8).times(position.takenPrice);
```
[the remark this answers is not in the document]
spitko: Yes, oTokens always have 8 decimals.

ishraq8:
QA looks good

Schema was recently updated: https://github.com/messari/subgraphs/pull/1967

but shouldn't affect data and the fields are optional. @spitko let me know if it will be difficult to update. Thanks for this!

subgraphs/opyn-gamma/src/entities/event.ts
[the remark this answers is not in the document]
dmelotik: Ah I think I interpreted this wrong. If I understand correctly now, this will actually be accurate bc the snapshot is taken beforehand IF the current event is on a new day, so it will update a previous day with all events that occurred before the event being handled.

dmelotik:
Also @spitko you can mark the ethereum deployment as `prod` now that the QA is passed

subgraphs/opyn-gamma/src/entities/event.ts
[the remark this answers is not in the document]
spitko: You're right in that a snapshot is taken whenever 1 or 24 hours have passed since the last one, but previous (already taken) ones can't be updated, since they're immutable. All the events that happened after the last snapshot will be included in the next one, whenever that is taken.

subgraphs/opyn-gamma/src/entities/snapshot.ts
```diff
@@ -0,0 +1,638 @@
...
+    snapshot.dailyWithdrawPremiumUSD = pool.cumulativeWithdrawPremiumUSD.minus(
+      prevSnapshot.cumulativeWithdrawPremiumUSD
+    );
+    snapshot.dailyTotalLiquidityPremiumUSD =
+      pool.cumulativeTotalLiquidityPremiumUSD.minus(
+        prevSnapshot.cumulativeTotalLiquidityPremiumUSD
+      );
+    snapshot.dailyCollateralVolumeUSD =
+      snapshot.cumulativeCollateralVolumeUSD!.minus(
+        prevSnapshot.cumulativeCollateralVolumeUSD!
```
dmelotik, line 203: I don't know for sure, but I think the runtime error may be caused because we are always assuming this field is populated when it may be null

spitko: The runtime error was already fixed, I just forgot to deploy the other network deployments.
But I ended up removing that field anyway for now.

ishraq8:
opyn qa looks good, left some comments in issue for documentation.

subgraphs/opyn-gamma/src/entities/pool.ts
dmelotik: Hmmm, maybe I was misreading something. It looks good to me! Sorry for any confusion

dmelotik (approved):
Nice work @spitko ! Sorry on the delay here

PR #1851, opened 2023-03-10
chore(); update arrakis version / fix error

dmelotik:
Update goldinch / arrakis versions

- Fix arrakis assembly script error


steegecs (approved):
Do you want to run this linter?

dmelotik:
Closing this, there is already an open arrakis PR, and I think it makes sense to do this in there. @tnkrxyz can use this as a model

PR #1868, opened 2023-03-13
fix(#minor); aura-finance; fix out of range indexing error

dhruv-chauhan:
Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Faura-finance-mainnet

### DIFF
--- a/subgraphs/aura-finance/src/common/getters.ts
+++ b/subgraphs/aura-finance/src/common/getters.ts
@@ -204,23 +204,24 @@ export function getOrCreateBalancerPoolToken(
       if (tokens[idx].lastPriceUSD! == BIGDECIMAL_ZERO) {
         const unknownPricePoolToken = tokens[idx];
 
-        const weights = getPoolTokenWeights(poolAddress, popIndex);
-
         const knownPricePoolTokenValueUSD = bigIntToBigDecimal(
           balances[knownPricePoolTokenIndex],
           knownPricePoolToken.decimals
         ).times(knownPricePoolToken.lastPriceUSD!);
 
-        const unknownPricePoolTokenValueUSD = divide(
-          weights[idx].times(knownPricePoolTokenValueUSD),
-          weights[knownPricePoolTokenIndex]
-        );
-        unknownPricePoolToken.lastPriceUSD = divide(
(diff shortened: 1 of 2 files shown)


subgraphs/aura-finance/src/common/getters.ts
```diff
@@ -202,25 +202,26 @@ export function getOrCreateBalancerPoolToken(
...
-          weights[idx].times(knownPricePoolTokenValueUSD),
-          weights[knownPricePoolTokenIndex]
-        );
-        unknownPricePoolToken.lastPriceUSD = divide(
-          unknownPricePoolTokenValueUSD,
-          bigIntToBigDecimal(balances[idx], unknownPricePoolToken.decimals)
-        );
-
-        unknownPricePoolToken.save();
+        if (weights.length == tokens.length) {
```
jaimehgb, line 206: Although the subgraph will keep indexing with this fix, it is not addressing the root cause. 

If the weights refer to the inputTokens of a given pool there should always be as many tokens as weights. Have you looked into what is causing the weights to not match the number of tokens?

dhruv-chauhan: Here's what I got,

- [Balancer Euler Boosted Pool Token (USDT) (bb-e-USDT)](https://etherscan.io/address/0x3c640f0d3036ad85afa2d5a9e32be651657b874f#events) was paused, disabling all methods and setting them to default, hence `getScalingFactors` (used to calculate weights) returns empty array.
- This balancer pool token is part of [bb-euler-USD Pool](https://app.balancer.fi/#/ethereum/pool/0x50cf90b954958480b8df7958a9e965752f62712400000000000000000000046f), which is boosted on Aura [Pool#52](https://app.aura.finance/#/pool/52).
- We have an internal field `vault._active` which turns false if the pool is halted on Aura. For now, looks like the pool is active (shutdown: false), hence the mismatch.

Screenshot from 2023-03-13 23-42-43

@jaimehgb

jaimehgb: Ahh, that makes sense. My bad then 👍

jaimehgb (approved):

PR #1873, opened 2023-03-13
fix(#patch): fix rocketpool build on case sensitive systems

jaimehgb:
Rocketpool was not building from scratch. Needed to update some paths and things


dmelotik (approved):
approved.

Issue #1511, opened 2022-12-12
Build Morpho Subgraph


dmelotik:
Morpho: https://www.morpho.xyz/

Notes:
- We want to use the 3.0.1 lending schema. We have a lending sdk [here](https://github.com/messari/subgraphs/tree/master/subgraphs/_reference_/src/sdk/protocols/lending) to help do all of the tedious actions. The developer only needs to make the protocol specific mappings, then they can use the logic hidden in the SDK.
- Compound V3 is an [example](https://github.com/messari/subgraphs/tree/master/subgraphs/compound-v3) of this in action.
- Since Morpho has liquidity in both Morpho and Aave or Compound we want to make sure we capture that nuance. To do that `protocol/market.totalValueLockedUSD` should be the total liquidity in aave/morpho. But then we should have helper fields called `protocol._morphoTotalValueLockedUSD` and `market._morphoTotalValueLockedUSD` to know the difference.
- In addition we probably want positions to have an extra field along the lines of `position._isMorpho`. This should tell someone that the position is in morpho P2P lending or aave.
- Since Morpho separates the protocol into morpho-compound and morpho-aave we should do the same with our subgraphs. This does not require more code, just setting up a new [configuration folder](https://github.com/messari/subgraphs/tree/master/subgraphs/abracadabra/protocols/abracadabra/config/deployments) watching the other contracts.
- Since morpho has aave rates/P2P rates I think we should focus on surfacing the P2P rates. I am even okay with leaving out the rates from aave pools. It actually seems out of the scope of the Morpho subgraph.
- Generally revenue in a lending protocol is going to be the interest borrowers pay. I think revenue in the Morpho subgraph should be the amount of an asset the borrow pays when P2P lending is happening. And if morpho takes a cut that is protocol side revenue.
- A README.md should be added like other protocols that outlines some of the unique things about morpho that are described here.

```[tasklist]
### Deployment QA Status
- [ ] https://github.com/messari/subgraphs/issues/2114
- [ ] https://github.com/messari/subgraphs/issues/2113
```

bhor123:
Hello @bye43 .This is Bhor Sharma, I wanted to contribute for developing the morpho subgraph. Can you please guide me , I was thinking of using the existing aave subgraph but if you have already made some progress please shed some light

bye43:
[...]

PR #1886, opened 2023-03-15
feat(): add morpho-compound and morpho-aave-v2 subgraph

julien-devatom:
## New subgraph
Implement Morpho protocol subgraph, including morpho-compound and morpho-aave-v2 on mainnet.

## About Morpho
[Morpho](https://morpho.xyz) is an on-chain peer-to-peer layer on top of lending pools. Rates are seamlessly improved for borrowers and lenders while preserving the same guarantees.

Test: 
- https://okgraph.xyz/?q=dmelotik%2Fmorpho-compound-ethereum
- https://okgraph.xyz/?q=dmelotik%2Fmorpho-aave-v2-ethereum

Closes: #1511


dmelotik:
@julien-devatom A few notes before I begin reviewing.

- I ran the GH actions. For the linting, [this should explain our linting](https://github.com/messari/subgraphs/blob/master/docs/WALKTHROUGH.md#github-actions) which is failing rn
- We will have to rethink the having 2 protocol entities. It makes sense, but it doesn't fit very nicely into our current BE infra. 
    - There are 2 ways we could go:
           - 1- we create 1 protocol and 1 subgraph. This downside of this is that there is little discernment between the 2 morpho deployments in our FE/the subgraph
           - 2- (my preferred option), we create 2 separate subgraph deployments, 1 for each morpho deployment. That would create 2 morpho pages on https://messari.io/protocol-explorer/all-protocols (which gives more nuance)
           
The issue here is that on our protocol page we only discern by network and not deployments rn, this is subject to change though.

julien-devatom:
> @julien-devatom A few notes before I begin reviewing.
> 
> * I ran the GH actions. For the linting, [this should explain our linting](https://github.com/messari/subgraphs/blob/master/docs/WALKTHROUGH.md#github-actions) which is failing rn
> * We will have to rethink the having 2 protocol entities. It makes sense, but it doesn't fit very nicely into our current BE infra.
>   
>   * There are 2 ways we could go:
>     - 1- we create 1 protocol and 1 subgraph. This downside of this is that there is little discernment between the 2 morpho deployments in our FE/the subgraph
>     - 2- (my preferred option), we create 2 separate subgraph deployments, 1 for each morpho deployment. That would create 2 morpho pages on https://messari.io/protocol-explorer/all-protocols (which gives more nuance)
> * We also have a lending [SDK](https://github.com/messari/subgraphs/tree/master/subgraphs/_reference_/src/sdk/protocols/lending) that will alleviate a lot of the dips in the cumulative values on snapshots
> 
> The issue here is that on our protocol page we only discern by network and not deployments rn, this is subject to change though.

Hey @dmelotik, I'll check the lintinh today. for the 2 protocol entities I totally agree that the second option is more suited. This is what we have done in our first subgraphs. Imo the best way to do that is to not touche the code inside src and just create 2 deployments w 2 templates indexing morpho-aave and morpho-compound. 
Is there a way for me to explore the subgraph in your app, with a kind of a sandbox? It will be easier for me to check how you are handling the rates etc, because on Morpho, there is 4 differents rates depending if you are matched or not (or partially matched sometime).

For the SDK, I found it after having coded the mappings. Do you think that this is a good time investment to move to the sdk now?

dmelotik:
@julien-devatom 

- Agreed, splitting into 2 different manifest templates and 2 different deployments will be the best option for this!
- Yes! You can see the one @jaimehgb deployed [here](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmY4GrGq8VKTAq4T6a6jcAoCHHzyvxSnKhRipx7Zrznzbv&tab=protocol). But when you look at your own deployments you can plug in the endpoint into [okgraph](https://okgraph.xyz/?q=jaimehgb%2Fmorpho) > click "Messari Visualizer"
- Rates will be interesting on the morpho subgraph. In our Aave subgraph we have 4 rates (stable/variable/ for lender/borrower). I think the way you are currently doing rates with P2P and Pooled is going to be good. 
- The thing about the SDK is that we know it works as long as you enter the right values so it makes debugging/maintenance much easier. On the flip side if debugging isn't too frustrating you already have everything done.
- It seems like markets are missing outputToken and some cumulativeUniqueUsage metrics.
     - Does morpho track positions using a token??

One of the big things I notice is that cumulative values are dropping when they should be strictly increasing. I am not 100% sure the culprit of this here. 1- entities don't always save when expected if they are loaded somewhere while concurrently being saved.

dmelotik:
Test deployment:

- Aave V2: https://okgraph.xyz/?q=dmelotik%2Fmorpho-aave-v2-ethereum

dmelotik:
Thank you @julien-devatom ! Those warnings are there to make sure we aren't missing anything for a protocol. Your explanation here makes sense.

I think given the subgraph is almost done it doesn't make sense to use the SDK at this point.

Would you mind if I made a commit to this PR to make some of the code more in line with the general code format of our other subgraphs? I think that will be the easiest thing to do going forward, and I will be able to do a more thorough review as I do that.

dmelotik:
@julien-devatom when you say "morpho specific improvements", do you mean new fields to watch that are specific to morpho (and/or any fixes to these current fields?)

If it is only the new deployment, that should not be a big issue. But, if I understand the improvements you are talking about it may make it easier to know what to do/not to do on our end.

julien-devatom:
> @julien-devatom when you say "morpho specific improvements", do you mean new fields to watch that are specific to morpho (and/or any fixes to these current fields?)

both but only for fields related to Morpho. For example, Ive added a field to the Market entity, `p2pSupplyInterestsUSD`, tracking the p2p interests earned by suppliers matched. I have other ideas in mind that can be interesting for a lending protocol like Morpho. Obviously this is going to be fully transparent on the Messari dashboard.
 
> I can see how adding a new deployment would be weird if both subgraphs are different since each morpho deployment has a unique factory contract...

The interface exposed by the different Morpho deployments is mainly the same. For morpho aave v3, this is a bit different since the protocol is distinguish the collateral and the supply (only the supply can be matched p2p) but I've refactored the common mapping for morpho aave v3 and I can reuse 90% of the code for the 3 versions of the Morpho protocol. Morpho aave v3 is going to be live really soon and I'll submit a new PR for the subgraph day 1.

dmelotik:
@julien-devatom this makes sense.

There are some things we will need to update in the subgraph before deploying (ie, fix build, fix lint, set versions correctly, reduce the uncaught exceptions/error handling).

Can we do this in a non-invasive way in order to make it easy to make upgrades here as you make updates on morpho?

Maybe I can make the changes, (not changing the schema or function at all), then you can just copy that over to your subgraph and we are in sync?

julien-devatom:
Yeah that's ok for me, i think this is the best way to do

julien-devatom:
Hey @dmelotik , any update? :)

dmelotik:
Hey @julien-devatom I had to branch off into a separate branch in order to push commits: https://github.com/messari/subgraphs/pull/2029

But it is close to done, I just need to add rewards to the subgraph. If you have any insights on where to grab morpho rewards that would be great

julien-devatom:
@dmelotik morpho rewards distribution is fully offchain (with only a merkle tree on chain for the distribution) and based on ages that are defining rates periodically. You have more information in [this repository](https://github.com/morpho-dao/morpho-rewards)

dmelotik:
@julien-devatom I finished it up if you want to take a look #2029 

It still needs to be reviewed and QA'd on our end

dmelotik:
Closing now that #2199 is merged in. If you notice any issues in our subgraph data or logic please feel free to open an issue, thanks for the help to get this going!

PR #1924, opened 2023-03-22
fix(#minor);Added position value at close for snapshot

ishraq8:
At close of position snapshot takes value before it goes to zero. Allows for profit and loss calculations.

### DIFF
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -1797,6 +1797,18 @@ type PositionSnapshot @entity(immutable: true) {
   " Notional value of collateral position, in USD "
   collateralBalanceUSD: BigDecimal!
 
+  " Notional value of position right before close, in asset units. In other events match balance. "
+  closeBalance: BigInt
+
+  " Notional value of collateral position right before close, in collateral units. In other events match collateralBalance."
+  closeCollateralBalance: BigInt
+
+  " Notional value of position right before close, in USD, at time of open. In other events match balanceUSD. "
+  closeBalanceUSD: BigDecimal
+
+  " Notional value of collateral position right before close, in USD. In other events match collateralBalanceUSD. "
+  closeCollateralBalanceUSD: BigDecimal
+
   " Block number of this snapshot "
   blockNumber: BigInt!
 
--- a/schema-derivatives-perpfutures.graphql
+++ b/schema-derivatives-perpfutures.graphql
@@ -1,5 +1,5 @@
 # Subgraph Schema: Derivatives Perpetual Futures
-# Version: 1.2.0
+# Version: 1.2.1
 # See https://github.com/messari/subgraphs/blob/master/docs/SCHEMA.md for details
 
 enum Network {


ishraq8:
simplified the field to `realisedpnl` @corerouter pointed out the smartcontract emits this field and we can simply take it from there. Thank you!

schema-derivatives-perpfutures.graphql
```diff
@@ -1797,6 +1797,9 @@ type PositionSnapshot @entity(immutable: true) {
   " Notional value of collateral position, in USD "
   collateralBalanceUSD: BigDecimal!
 
+  " Realised profit and loss in USD at the time of close or event "
+  realisedPnlUSD: BigInt
```
corerouter, line 1801: Suggest to change realisedPnlUSD type from BigInt to BigDecimal to be consistent with other USD's type.

corerouter:
> simplified the field to `realisedpnl` @corerouter pointed out the smartcontract emits this field and we can simply take it from there. Thank you!

Looks good for me. Just leave a comment for the realisedpnlUSD type. Thanks!

schema-derivatives-perpfutures.graphql
ishraq8: done

dmelotik (approved):
LGTM, which subgraphs being built rn and already built should we update with this? @ishraq8

schema-derivatives-perpfutures.graphql
```diff
@@ -1797,6 +1797,15 @@ type PositionSnapshot @entity(immutable: true) {
...
 
+  " Notional value of position right before position closed, in USD, only useful at close events. "
+  closeBalanceUSD: BigDecimal
+
+  " Notional value of collateral position ight before position closed, in USD, only useful at close events "
+  closeCollateralBalanceUSD: BigDecimal
+
+  " Realised profit and loss in USD at the time of close or event "
+  realisedPnlUSD: BigDecimal
+
```
dmelotik, line 1808: Doesn't it make more sense to add these to the `Position` and not the snapshot? This is only going to be recorded once, not necessary for each snapshot

dmelotik (commented):
Can you help me understand the difference between balanceUSD and colalteralBalanceUSD?

schema-derivatives-perpfutures.graphql
ishraq8: moved `closeBalanceUSD` and `closeCollateralBalanceUSD` to position not in snapshot. `realisedPnlUSD` i left in snapshot because it might be good to track during the lifetime of position.

ishraq8:
> Can you help me understand the difference between balanceUSD and colalteralBalanceUSD?

`collateralBalanceUSD` is the collateral value the user inputed. `balanceUSD` is the notional value of the entire position (collateral+leverage).

schema-derivatives-perpfutures.graphql
dmelotik: agreed, that could change throughout the life of the position. Would we also want to track it in the position as well?

ishraq8: lets track in position as well. Added. Thanks for going through this and giving input. really valuable!

dmelotik (approved):
nice addition @ishraq8 you can now put BE developer on your resume 🤫

PR #1941, opened 2023-03-23
fix(#minor); Uniswap V3 Arbitrum; Blacklist some tokens

steegecs:
**Context:**
- Pricing outliers for Trident, Blue, and Arbitrum. These are all fake versions of these tokens and they are screwing up the metrics. So here I blacklist them.


dmelotik (approved):
lfgtm m8

PR #1957, opened 2023-03-27
feat(); Uniswap V3; Added fake token detection for more robust pricing

steegecs:
**Context:**

As per the pricing investigation that is currently underway, we have discovered that many of the pricing issues result from fake versions of tokens. The pool that ends up being the outlier in the instances seen, are often composed of a token that is a fake version of a whitelist token and another token, sometimes fake as well, that has an outlier price. This fix seeks to mitigate pricing issues from these scenarios by not getting USD metrics that contain a fake version of a whitelisted token.

### DIFF
--- a/subgraphs/uniswap-v3/src/common/entities/token.ts
+++ b/subgraphs/uniswap-v3/src/common/entities/token.ts
@@ -1,4 +1,4 @@
-import { Address, BigInt, Bytes, ethereum, log } from "@graphprotocol/graph-ts";
+import { Address, BigInt, Bytes, ethereum } from "@graphprotocol/graph-ts";
 import { NetworkConfigs } from "../../../configurations/configure";
 import { ERC20 } from "../../../generated/Factory/ERC20";
 import {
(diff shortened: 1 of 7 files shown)


subgraphs/uniswap-v3/src/common/entities/token.ts
```diff
@@ -106,28 +110,63 @@ export function getOrCreateTokenWhitelist(
...
+
+    tokenWhitelistSymbol.address = tokenAddress;
+    tokenWhitelistSymbol.save();
+  }
+
+  return tokenWhitelistSymbol;
+}
+
+export function isFakeWhitelistToken(token: Token): bool {
+  const tokenWhitelistSymbol = _TokenWhitelistSymbol.load(token.symbol);
```
dmelotik, line 130: Make case insensitive and chop beginning/ending whitespace?

dmelotik (approved):
Can you add this description to the readme for uni v3?

Also lint thingy

PR #1960, opened 2023-03-28
chore(); Add logging docs/severity

dmelotik:
The goal of this PR is to outline a standard logging format for subgraphs going forward.

In addition, we want to set clear definitions for levels of severity of subgraph logs.


docs/WALKTHROUGH.md
```diff
@@ -80,6 +80,61 @@ export function handleDeposit(event: Deposit): void {
 
 Play around with this, compare hashes on etherscan and help yourself understand some of the behavior.
 
+## Using Logs in Subgraphs
+
+There is a standard logging format that we follow in subgraphs to ensure subgraph data consumers have a framework to understand what is happening in a subgraph. There are five different types of log severity levels that we use in subgraphs:
+
+- `log.debug`: This is only used for debugging purposes. It should not be found in any subgraph code in our `master` branch.
```
jaimehgb, line 87: We can enforce this then at the CI to prevent merges with it

docs/WALKTHROUGH.md
```diff
@@ -80,6 +80,61 @@ export function handleDeposit(event: Deposit): void {
...
+- `log.warning`: This is used for issues/warnings that may or may not be an issue. The issue will not create any major issues in a subgraph and should not affect downstream data consumers. This could be a divide by 0, but it may be the actual value so it is not an error that needs to be fixed.
+- `log.error`: This is used for errors that need to be addressed, but are not fatal to the subgraph's function. This should be cause for a fix, but the subgraph can continue to execute. This may cause a minor data issue that should be addressed soon. This would be synonymous to a [#minor](https://github.com/messari/subgraphs/blob/master/docs/CONTRIBUTING.md#minor) fix in our versioning system.
+- `log.critical`: This will be used for fatal errors. When this log is reached a subgraph will stop executing. This error will be cause for a major data issue that would create a major issue downstream if the data is consumed. This needs to be fixed ASAP and is synonymous to a [#major](https://github.com/messari/subgraphs/blob/master/docs/CONTRIBUTING.md#major) fix in our versioning system.
```
jaimehgb, line 91: The value passed to `log.critical` is not shown in the logs, it only shows that it stopped with a critical error.

We might require people to always log a `log.error` right before the `log.critical` so we can see what was the actual message

docs/WALKTHROUGH.md
```diff
@@ -80,6 +80,61 @@ export function handleDeposit(event: Deposit): void {
...
+
+Each log severity has a different purpose and should be used accordingly.
+
+> The goal is to remove ambiguity of log meaning for devs, help downstream consumers understand what is happening on a subgraph-level, and use downtime judiciously.
+
+### Logging Format
+
+Logs should be descriptive and formatted as follows to remain easy to parse through and understand. There are some pieces of information your logs should have:
+
+- The function name that the log is found in
```
jaimehgb, line 101: It's unfortunate that AS doesn't support accessing the stack trace yet: https://www.assemblyscript.org/stdlib/error.html#instance-members

We could add the function name and line without bothering the dev to do so.

docs/WALKTHROUGH.md
```diff
@@ -80,6 +80,61 @@ export function handleDeposit(event: Deposit): void {
...
+- A description of what the log is
+- Any relevant information that is needed to understand the log
+- A prefix to differentiate messari subgraph logs
+- If possible, the transaction hash should also be included
```
jaimehgb, line 105: These 2 we can add via a very simple logger wrapper:

```
class MessariLogger {
  ev : ethereum.Event;
  constructor(event: ethereum.Event) {
    this.ev = event;
  }
  
  private formatLog(msg: string): string {
    return `messari_log: tx: ${this.ev.transaction.hash.toHexString()} :: ${msg}`
  }
  public debug(...): void {...}
  public info(...): void {...}
  public warning(...): void {}
  public error(...): void {...}
  public critical(...): void {...} // this one can do the log error + critical too that I mentioned above
}
```

Then on the mapping, instantiate this every time a log is needed, and pass the logger around instead of using it as a global. And prevent the default logger from being used on the CI

jaimehgb (commented):
🔥

docs/WALKTHROUGH.md
dmelotik: I ran a test of this. It does actually show up. It just doesn't show up in the error message field when pinging for the health of a subgraph from the graph api.

See the pic in this msg: https://discord.com/channels/953684103012683796/1090341413700501524/1090833508664086578

docs/WALKTHROUGH.md
dmelotik: RIP

docs/WALKTHROUGH.md
dmelotik: I like this https://github.com/messari/subgraphs/issues/1977

docs/WALKTHROUGH.md
dmelotik: https://github.com/messari/subgraphs/issues/1977 i like

dmelotik: added into this PR

docs/WALKTHROUGH.md
jaimehgb: Oh! This is either new, or they don't show in the UI in the hosted-service (which is where I haven't seen them). But this works then!

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  constructor(event: ethereum.Event) {
+    this.event = event;
+  }
+
+  private formatLog(funcName: string, msg: string): string {
+    return `messari_log: [${funcName}}] tx: ${this.event.transaction.hash.toHexString()} :: ${msg}`;
+  }
+
+  info(funcName: string, msg: string): void {
```
jaimehgb, line 14: I'd keep the ability of passing variables from the user to be formatted too. Otherwise they need to take care of the formatting to string before logging.

```
info(funcName: string, msg: string, params: Array<string>): void {
  log.info(this.formatLog(funcName, msg), params);
}
```

Also, since we are already used to the format `log.info(template, params)`, I'd make the funcName be the last parameter (so people don't get confused and accidentally put the message there).

Or! Even better. When you use the logger you have to instantiate it first, right? Make the function where it is instantiated one of the constructor arguments. Then we don't need to set the name every time we log. The only issue is if you pass the logger down to other aux functions in you handler. I can think of a hack for this, but feels a bit ugly, like fighting against the dev 😅 . But here it goes, for inspiration:

```
class Logger { // unexported, so we don't have access to the type, this way we cannot make it a function argument
  wrap(): RefLogger {}
}

export MakeLogger(...); // the constructor to be able to instantiate the logger in the first place

export class RefLogger {
  private logger: Logger
  getLogger(fnName: string): Logger;
}
```

This way, if we want to use it in mutliple functions we'd need to do something like:
```
export function handleEvent(ev: Event): void {
  const logger = MakeLogger("handleEvent", ev, ....);
  logger.info("something something");
  someAuxStuff(ev, logger.wrap());
}

function someAuxStuff(ev: Event, log: RefLogger): void { // RefLogger because since Logger is not exported we can only pass it this way
  const logger = log.getLogger("someAuxStuff");
  
  logger.warning("uh, bad things happening");
  moreAuxStuff(ev, logger.wrap());
}

function moreAuxStuff(ev: Event, log: RefLogger): void {
  const logger = log.getLogger("moreAuxStuff");
  // ...
}
```

It could happen that you pass the logger down a few levels but only use it in the inner most call. Since you need to unwrap it with a function name, we might not have the whole stack, but at least the function name where the message was logged.

dmelotik: lmk what you think of the changes I made

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -1,26 +1,50 @@
...
+      }}] tx: ${this.event.transaction.hash.toHexString()} :: ${msg}`;
+    }
+
+    return `messari_log: tx: ${this.event.transaction.hash.toHexString()} :: ${msg}`;
   }
 
-  info(funcName: string, msg: string): void {
-    log.info(this.formatLog(funcName, msg), []);
+  info(msg: string, args: string[] | null = null): void {
+    if (!args) args = [];
```
jaimehgb, line 31: Instead of using `null` as the default, do `[]`. This way you don't need to check `if(!args)` every time? :P

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -1,26 +1,50 @@
...
-    log.error(this.formatLog(funcName, msg), []);
+  critical(msg: string, args: string[] | null = null): void {
+    if (!args) args = [];
+    log.critical(this.formatLog(msg), args);
   }
-  critical(funcName: string, msg: string): void {
-    log.critical(this.formatLog(funcName, msg), []);
+
+  updateFunctionName(fnName: string): void {
+    this.fnName = fnName;
```
jaimehgb, line 48: Not critical, but useful nevertheless: keep all the history of function names. This way we can also print with the log some rudimentary stack trace

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -1,26 +1,50 @@
...
+ * @param funcName The name of the function that triggered the log message
+ * @param msg The message to log (add `{}` to the message to add additional arguments)
+ * @param args Any additional arguments to add to the msg (same format as graph-ts log functions)
+ */
 export class Logger {
   event: ethereum.Event;
+  fnName: string | null;
 
-  constructor(event: ethereum.Event) {
+  constructor(event: ethereum.Event, fnName: string | null = null) {
```
jaimehgb, line 15: In which circumstances we'd want to allow a null function name?

Can we force the user to always set a fnName?

dmelotik: never

subgraphs/_reference_/src/common/utils/logger.ts
dmelotik: done

docs/WALKTHROUGH.md
```diff
@@ -80,6 +80,63 @@ export function handleDeposit(event: Deposit): void {
...
+
+Examples:
```
tnkrxyz, line 109: we should convert these examples to use the Logger class

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Logger class is used to log messages to be picked up by our logging system.
+ *
+ * @param event The event that triggered the log message
+ * @param funcName The name of the function that triggered the log message
+ * @param msg The message to log (add `{}` to the message to add additional arguments)
+ * @param args Any additional arguments to add to the msg (same format as graph-ts log functions)
+ */
+export class Logger {
+  event: ethereum.Event;
+  fnName: string; // this will act as a rudimentary stack trace
```
tnkrxyz, line 13: rename to funcName to match param doc above. or the other way around. I prefer funcName personally.

subgraphs/_reference_/src/common/utils/logger.ts
```diff
@@ -0,0 +1,46 @@
...
+    log.warning(this.formatLog(msg), args);
+  }
+  error(msg: string, args: string[]): void {
+    log.error(this.formatLog(msg), args);
+  }
+  critical(msg: string, args: string[]): void {
+    log.critical(this.formatLog(msg), args);
+  }
+
+  addFunctionName(fnName: string): void {
```
tnkrxyz, line 43: `appendFuncName(funcName: string)`?

tnkrxyz (commented):
Nice work!

subgraphs/_reference_/src/common/utils/logger.ts
jaimehgb: I like it

jaimehgb (approved):

PR #1967, opened 2023-03-31
Feat(#minor); Added `collateralvolume` on options schema 

ishraq8:
`VolumeUSD` currently refers to notional volume at pool level. This is the most relevant metric. in discussions from dopex it was suggested to add `collateralVolume` as an additional metric as it can also be tracked. it is a less relevant metric but makes our schema more comprehensive.

### DIFF
--- a/schema-derivatives-options.graphql
+++ b/schema-derivatives-options.graphql
@@ -462,6 +462,9 @@ type FinancialsDailySnapshot @entity(immutable: true) {
   " All historical trade notional volume of underylying assets in USD "
   cumulativeVolumeUSD: BigDecimal!
 
+  " Current Collateral assets in USD in this pool that was used for trading, amount taken at snapshot"
+  collateralUsedUSD: BigDecimal
+
   " Daily volume of Collateral assets in USD in this pool that was used for trading"
   dailyCollateralVolumeUSD: BigDecimal
 
@@ -895,6 +898,9 @@ type LiquidityPoolDailySnapshot @entity(immutable: true) {
 
   " Daily trade notional volume occurred when contract created, in USD "
   dailyVolumeUSD: BigDecimal!
+
+  " Current Collateral assets in USD in this pool that was used for trading, amount taken at snapshot"
+  collateralUsedUSD: BigDecimal
 
   " Daily volume of collateral when contract created, in USD, that was used for trading "
   dailyCollateralVolumeUSD: BigDecimal
--- a/schema-derivatives-options.graphql
+++ b/schema-derivatives-options.graphql
@@ -1,5 +1,5 @@
 # Subgraph Schema: Derivatives Options
-# Version: 1.3.0
+# Version: 1.3.1
 # See https://github.com/messari/subgraphs/blob/master/docs/SCHEMA.md for details
 
 enum Network {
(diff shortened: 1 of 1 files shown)


schema-derivatives-options.graphql
dmelotik, line 1: we probably also want to add this to the protocol entity and financial daily snapshot?

ishraq8: added, thank you. Originally wanted to keep it limited to pool level but it is fine living in both, if we do it might as well do it for both. appreciate the input.

jaimehgb (approved):

ishraq8:
> Is this not the same as deposit volume, which already has a field under LiquidityPool? If we want this also under protocol level, maybe it should be called `depositVolume` instead for consistency?

Good question, @spitko,  we were having a discussion for dopex about notional volume (which includes leverage) and actual collateral volume (which does not include leverage). The `depositVolume` currently includes leverage (volume can be artificially higher). It also depends on protocol, some protocols could have `depositVolume` the same as `collateralVolume` it depends on how leverage is handled.

Generally notional volume is a more important metric and this `collateralVolume` is a nice to have addition. Let me know if that clarifies and if there are any issues.

spitko:
I don't think deposit volume should ever include leverage? Deposit is a transfer of x input tokens to pool, and volume of that deposit would be value of those x tokens, no? Same would apply for withdraw and withdraw volume.

Where it gets confusing is exercised, closed and cumulative volume. LiquidityPool only has a reference to the input token (i.e **collateral** token for an option), but from what I understand, we want to use the option's **underlying** token value to calculate those fields. There is no reference in `LiquidityPool` to the asset using which these fields are calculated. 

Additionally it is not clear what all should be counted under `cumulativeVolumeUSD`.
The field comment currently states this is:

> All historical notional volume in USD taken of underlying at time of contract minting

which I understood as that it would include just mint volume (using the option **underlying** value), but should this also include something else? And if I'm understanding this correctly, `collateralVolumeUSD` would include all the same volume as `cumulativeVolumeUSD`, but calculated using the **collateral** token instead?

ishraq8:
@spitko Talked more with @corerouter and clarified the schema. You are right it is confusing and `depositvolume` should be all deposits without leverage. 

`collateralvolume` is more in reference to trading (perpfutures are kind of a combination of lending and dex). So it is looking at the collateral used in the trading volume rather than taking the notional trading volume. This could be different than `depositvolume` because not the entirety of the deposit may be used as collateral for trading.  Let me know if that clarifies. Happy to discuss further, would like the schema to be clear to everyone before merge.

spitko:
I think we have different understandings of notional value. The way I see it is that notional value is equal to `(controlled underlying amount) * (current underlying price)`.
But what you're saying is that `cumulativeVolumeUSD` should include just trades, i.e sum of purchased options notional value? 

> `collateralvolume` is more in reference to trading (perpfutures are kind of a combination of lending and dex). So it is looking at the collateral used in the trading volume rather than taking the notional trading volume. This could be different than `depositvolume` because not the entirety of the deposit may be used as collateral for trading.

Deposits are not really used for trading. Deposits are used as collateral to sell options for some underlying token.
When an option is bought, the buyer pays a premium, and receives an option that controls some amount of underlying token. The notional value of that trade would therefore be `(underlying token amount) * (current underlying price)`.
And if the option is backed by some other token, the collateral value in usd would be equal to the notional value.

PR #1973, opened 2023-04-02
feat(#minor); optimism-bridge-v2; Upgrade to 1.2 schema

spitko:
https://okgraph.xyz/?q=spitko%2Foptimism-bridge-v2-optimism
https://okgraph.xyz/?q=spitko%2Foptimism-bridge-v2-ethereum


dmelotik (commented):
Thanks for updating this @spitko !

What about `totalValueImported` on ethereum?

jaimehgb:
> What about `totalValueImported` on ethereum?

Unless you can bridge native optimism tokens to mainnet it is correct if it is 0.

Value imported / exported refers to the value of foreign tokens currently liquid in the chain, and the value of this chain native tokens currently liquid abroad respectively.

dmelotik:
@jaimehgb if there is value exported from optimism, wouldn't that mean there should be value imported on ethereum?

spitko:
> @jaimehgb if there is value exported from optimism, wouldn't that mean there should be value imported on ethereum?

Yes it would. But in the case of Optimism Gateway, it is not possible to transfer any optimism native tokens to ethereum network. Which is why the TVL and exported value are both 0 on optimism.

dmelotik:
oops, I thought optimism has a TVE. All good

PR #1978, opened 2023-04-04
fix(#minor); GMX Subgraph; Upgrade gmx subgraph with new perp schema which adds position value at close for snapshot

corerouter:
Upgrade GMX with the latest perp schema which adds position value at close for snapshot (https://github.com/messari/subgraphs/pull/1924)


ishraq8:
Did qa on: https://okgraph.xyz/?q=corerouter%2Fgmxtest

data looks accurate, just need codre review. Thank you @corerouter great work!

PR #1987, opened 2023-04-06
fix(minor); Uniswap V3; Fix position data + unique users on protocol entity

steegecs:
**Context:**
Position data has been incorrectly tracked. In the data, the position count is 0 for all pools and accounts. Additionally, the open and closed position counts also needed to be fixed for pools and accounts. This PR changes the position count data for open, closed, and cumulative position data. 

Additionally, cumulativeUniqueUsers, traders, and LPs were not being properly tracked. This PR also addresses this on the DEXAmmProtocol entity.

https://thegraph.com/hosted-service/subgraph/messari/pancakeswap-v3-ethereum
https://thegraph.com/hosted-service/subgraph/steegecs/uniswap-v3-celo?version=pending


subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
```diff
@@ -157,9 +166,22 @@ export function handleDecreaseLiquidity(event: DecreaseLiquidity): void {
 }
 
 export function handleTransfer(event: Transfer): void {
-  const account = getOrCreateAccount(event.params.to);
   const position = getOrCreatePosition(event, event.params.tokenId);
 
+  const account = getOrCreateAccount(event.params.to);
+  const oldAccount = getOrCreateAccount(position!.account);
```
dmelotik, line 172: Instead of assuming we have a position, I think we should be checking if it is null and throw a warning (or error) if not. It would crash this, but also mean there is something else wrong that is deeper

subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
```diff
@@ -157,9 +166,22 @@ export function handleDecreaseLiquidity(event: DecreaseLiquidity): void {
...
+  if (isClosed(position!)) {
+    account.closedPositionCount += INT_ONE;
+    oldAccount.closedPositionCount -= INT_ONE;
+  } else {
+    account.openPositionCount += INT_ONE;
+    oldAccount.openPositionCount -= INT_ONE;
+  }
+
   // position was not able to be fetched
   if (position == null) {
```
dmelotik, line 186: Ig, just move this above the new code block and add a log

subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
```diff
@@ -157,9 +166,22 @@ export function handleDecreaseLiquidity(event: DecreaseLiquidity): void {
 }
 
 export function handleTransfer(event: Transfer): void {
-  const account = getOrCreateAccount(event.params.to);
   const position = getOrCreatePosition(event, event.params.tokenId);
 
+  const account = getOrCreateAccount(event.params.to);
+  const oldAccount = getOrCreateAccount(position!.account);
+
+  account.positionCount += INT_ONE;
```
dmelotik, line 174: Is it possible for a transfer to not create a new position, but to just add to an existing one?

dmelotik:
Also test deployment? :)

subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
steegecs: Yeah, this operation should not create any new positions. I decided to handle this as just transferring the position from one account to the other and incrementing this value. Along with the open and closed position count depending on if it is opened or closed.

dmelotik:
idk if you can squeeze a fix in here, but this log message doesn't seem accurate: https://github.com/messari/subgraphs/blob/master/subgraphs/uniswap-v3-forks/src/common/price/price.ts#L197

subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
dmelotik: I see, I missed the part that it was in an if block that only executed if the account was created

subgraphs/uniswap-v3-forks/src/mappings/position-manager.ts
dmelotik: Should we add a log if the pos is not found? Wouldn't that mean something else is wrong

steegecs: Yeah, good point. I'll add an error here since now it ideally should not occur.

dmelotik (approved):
lgtm

PR #1993, opened 2023-04-10
feat (#major); aave-forks; upgrade aave-forks to schema 3.0.1 and use sdk

tnkrxyz:
See #1519 a list of tasks in this PR.

Almost there. I have a few questions, @dmelotik, maybe you can help:

- ~~What should be the FeeType for Flashloan fees? FeeType.OTHER?~~ Added FeeType.FLASHLOAN_FEE_LP and FeeType.FLASHLOAN_FEE_PROTOCOL

- ~~Currently risk type is set at protocol level, but aave-v3 allow both isolated and global risk type at user/account level. It seems it may be necessary to add a RiskType.MIXED?~~ Noted in README

- ~~interestRateType info is available in Borrow events, but not in Repay events. Unless we use callHandler for Repay (but then it still won't work for chains not supporting callHandler), we don't know the interestRateType for Repays. This creates a problem that repays cannot find matched borrowing positions to subtract: https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/BorrowLogic.sol#L32-L47 . But if we ignore interestRateType for borrows, we may combine stable and variable interest positions (this may be fine?) There is a similar issue for Liquidations, we don't know the interestRateType for liquidated Borrow position and both Variable and Stable positions may be liquidated at the same time.~~ came up with a solution to detect interestRateType for repays and liquidations

A couple possible improvements to lending sdk I hope to check whether makes sense:  
-  use enum type to define function argument type, e.g. `getOrCreateRewardType(rewardTokenType: string)`  ==> `getOrCreateRewardType(rewardTokenType: RewardTokenType)`. We just need to add `export type RewardTokenType = string` after the enum definition. Similarily for InterestRateType, PositionSide, TransactionType

- specify a tokenPricier object in the manger class, that basically calls `getAssetPriceInUSDC()` (this is how bridge sdk handle prices) so the sdk can determine best place to refresh prices.

### Deployment & Dashboard
- aave-amm-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-amm-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-amm-ethereum&tab=protocol
- aave-arc-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-arc-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-arc-ethereum&tab=protocol
- aave-rwa-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-rwa-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-rwa-ethereum&tab=protocol
- aave-v2-avalanche
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v2-avalanche?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v2-avalanche&tab=protocol
- aave-v2-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v2-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v2-ethereum&tab=protocol
- aave-v2-polygon
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v2-polygon?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v2-polygon&tab=protocol
- aave-v3-arbitrum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-arbitrum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-arbitrum&tab=protocol
- aave-v3-avalanche
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-avalanche?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-avalanche&tab=protocol
- aave-v3-fantom
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-fantom?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-fantom&tab=protocol
- aave-v3-harmony
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-harmony?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-harmony&tab=protocol
- aave-v3-polygon
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-polygon?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-polygon&tab=protocol
- aave-v3-optimism
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-optimism?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-optimism&tab=protocol
- aave-v3-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/aave-v3-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum&tab=protocol
- geist-finance-fantom
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/geist-finance-fantom?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/geist-finance-fantom&tab=protocol
- spark-lend-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/spark-lend-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/spark-lend-ethereum&tab=protocol
- radiant-capital-arbitrum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/radiant-arbitrum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/radiant-arbitrum&tab=protocol
- radiant-capital-v2-arbitrum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/radiant-capital-v2-arbitrum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/radiant-capital-v2-arbitrum&tab=protocol
- radiant-capital-v2-bsc
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/radiant-capital-v2-bsc?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/radiant-capital-v2-bsc&tab=protocol
- uwu-lend-ethereum
    - deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/uwu-lend-ethereum?version=pending
    - dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/uwu-lend-ethereum&tab=protocol


subgraphs/aave-forks/protocols/aave-v3/src/constants.ts
```diff
@@ -14,11 +14,12 @@ export const USDC_POS_TOKEN_ADDRESS =
 /////////////////////////////
 
 export namespace Protocol {
+  export const PROTOCOL = "Aave";
   export const NAME = "Aave v3";
   export const SLUG = "aave-v3";
 }
 
-export namespace TokenType {
+export namespace IavsTokenType {
```
dmelotik, line 22: What is this used for?

subgraphs/aave-forks/protocols/aave-v3/README.md
```diff
@@ -24,7 +26,7 @@ Portion of the Total Revenue allocated to the Protocol
 
 Sum across all Pools:
 
-`(Pool Oustanding Borrow Amount * Pool Borrow Rate) * (Pool Reserve Factor)`
+`(Pool Oustanding Borrow Amount * Pool Borrow Rate) * (Pool Reserve Factor) + liquidiated collateral amount * liquidiation penalty * liquidation procotol fee percentage`
```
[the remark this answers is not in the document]
tnkrxyz: getting there.

subgraphs/aave-forks/protocols/aave-v3/src/constants.ts
tnkrxyz: It is used to determine the reward token type based on whether the token is aToken, vToken, or sToken. see https://github.com/messari/subgraphs/blob/75aa0e594b18c307b660cf798b3e51dead2177de/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts#L185-L193 . Since schema 3.0.1 already has token.TokenType, I renamed it to IavsTokenType (Input, A, V or S token).

dmelotik:
We can either use fee type other and note in the readme or update the schema. Same w the interest rate type.

I am going to opt for the readme option, I don't think the extra work to update is worth it for 1 protocol.

dmelotik:
We currently don't specify stable vs variable borrows.. do we need to add it?

I think it is valuable info, but we can skip for now and make a note in the readme. I think the logic we have is working regardless of borrow type.

Also, you would know which side based on the token, variable and stable borrow tokens have different contract addresses. You can add a field to the token for which side it is and compare

subgraphs/aave-forks/protocols/aave-v3/README.md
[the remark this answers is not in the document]
tnkrxyz: good point! I agree it is better to not double counting flashloan revenue to LP. For revenue details, we are [not providing   detailed FeeType for supply side revenue currently anyway](https://github.com/messari/subgraphs/blob/3c01f4df091a0fba1b435eccfa8a19626586eb14/subgraphs/_reference_/src/sdk/protocols/lending/constants.ts#L83), right.

tnkrxyz:
> We currently don't specify stable vs variable borrows.. do we need to add it?
 
> I think it is valuable info, but we can skip for now and make a note in the readme. I think the logic we have is working regardless of borrow type.

I can add an interestRateType field to Borrow entity if you think it makes sense to add it now.
 
> Also, you would know which side based on the token, variable and stable borrow tokens have different contract addresses. You can add a field to the token for which side it is and compare

I believe the asset arg in Borrow and Repay event are the underlying (input) token, not the vToken or sToken, so we cannot know whether it is stable or variable interest type. The info is available in the [params struct](https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/BorrowLogic.sol#L198), but not in the event.

https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/BorrowLogic.sol#L157 and https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/BorrowLogic.sol#L262

Similarly, the LiquidationCall event only provides the asset token:
https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/LiquidationLogic.sol#L233-L241

subgraphs/aave-forks/protocols/aave-v3/README.md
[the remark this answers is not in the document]
dmelotik: To me `PROTOCOL_FEE` is from interest generated, we already split it up between supply/protocol side. The fee type should be looked at as the source type of the interest/revenue/fees

dmelotik:
@tnkrxyz 

>  believe the asset arg in Borrow and Repay event is the underlying (input) token
Ah, this is true

Okay this is what I found (lmk if this doesn't work, but it makes sense in my head):

- Like you said we can get the interest rate mode from the [borrow event](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L161)
- That is what creates the position so it is already in the position entity (it is not possible for a user to have a variable and stable borrow from the same market, afaik), so no need to add an identifier for that in the position entitiy.
- You _could_ add a helper field in the Borrow entity if you would like, but I don't think it is necessary. Could be a nice to have.
- BUT, a user can swap borrow modes. So we need to be watching the [`SwapBorrowRateMode`](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L351) event now.

subgraphs/aave-forks/protocols/aave-v3/README.md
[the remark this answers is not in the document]
tnkrxyz: so, it'll not be a problem for flash loan revenue details to the supply side because we are not breaking supply side revenue by interest/floash loan etc, and the fee type will all be PROTOCOL_FEE for supply side revenue?
```
export namespace FeeType {
  export const LIQUIDATION_FEE = "LIQUIDATION_FEE";
  export const ADMIN_FEE = "ADMIN_FEE";
  export const PROTOCOL_FEE = "PROTOCOL_FEE";
  export const MINT_FEE = "MINT_FEE";
  export const WITHDRAW_FEE = "WITHDRAW_FEE";
  export const OTHER = "OTHER";
}
```

tnkrxyz:
> @tnkrxyz
> 
> > believe the asset arg in Borrow and Repay event is the underlying (input) token
> > Ah, this is true
> 
> Okay this is what I found (lmk if this doesn't work, but it makes sense in my head):
> 
> * Like you said we can get the interest rate mode from the [borrow event](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L161)

> * That is what creates the position so it is already in the position entity (it is not possible for a user to have a variable and stable borrow from the same market, afaik), so no need to add an identifier for that in the position entitiy.

I believe it is possible for a user to borrow in both variable and stable interest rate mode at the same time. For example, in [this dune query](https://dune.com/queries/2394749), you can see user 0x7c04e80e5d8a19ef014e51e27b3d92cc4d87b515 borrowed 25000000000 with variable rate mode (mode=2) at block 27289290, and then another 25000000000 with the stable rate mode (mode = 1) at block 27289554. Neither position was repaid until block 27633344 and 27633420, according to [this query](https://dune.com/queries/2394847). And there was no SwapBorrowRateMode event.

Since this is the case, if we remove the interestRateType suffix from positionID, we will be adding the later stable borrow in the above example into the previous variable borrow position. But may be this is OK?

> * You _could_ add a helper field in the Borrow entity if you would like, but I don't think it is necessary. Could be a nice to have.

If we are not having the interest rate type in position, it'll be a good idea to have a helper field for interest rate type in Borrow entity.

> * BUT, a user can swap borrow modes. So we need to be watching the [`SwapBorrowRateMode`](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L351) event now.

If we are not tracking interestRateType for positions, Borrow, Repay, or Liquidation), then we don't need to watching this, right?

dmelotik:
@tnkrxyz you are right!

- As a note you have a deeper knowledge on this issue than I do rn, so feel free to do what you need to get it to work (ideally avoiding callHandlers)

Sounds like you can have 2 pos in a market with different interest rate types, so we should definitely separate by interest rate type and add any helper fields that are needed to enable this.

lmk if you have any other questions!

subgraphs/aave-forks/protocols/aave-v3/README.md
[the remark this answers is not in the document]
dmelotik: So to answer the question, yes. I will check w our ppl internally and ensure this is okay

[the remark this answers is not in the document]
dmelotik: Actually, how would you go about ensuring the rev to LP from flashloans gets added to `supplySideRev`, but we still add it to the RevenueDetails as type `FLASHLOAN` (you can add this enum type), and avoid double counting?

I think that is what we should do tbh. Sorry for all the confusion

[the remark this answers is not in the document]
dmelotik: Okay, sorry for the confusion.

We want to be as specific as possible on aave, so let's add a new `FeeType` called `FLASHLOAN_LP_FEE` and `FLASHLOAN_PROTOCOL_FEE`, then whatever you need to do in order to separate the 2 without double counting. And just to be transparent we should mark the methodology either in a comment in the code or readme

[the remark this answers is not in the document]
tnkrxyz: Yes, we can do that. I will add a helper entity/field to avoid double counting.

tnkrxyz:
> @tnkrxyz you are right!
> 
> * As a note you have a deeper knowledge on this issue than I do rn, so feel free to do what you need to get it to work (ideally avoiding callHandlers)
> 
> Sounds like you can have 2 pos in a market with different interest rate types, so we should definitely separate by interest rate type and add any helper fields that are needed to enable this.
> 
> lmk if you have any other questions!

I have to think how to achieve this without using callHandler.

dmelotik:
@tnkrxyz I think one of the biggest issues with the call handlers is not being supported on every chain and we need to support all chains.

One way I can think of would be:
- Watch the Repay events emitted, then go up in the event logs to see which DebtToken was burned.
- You will have to watch the [`useATokens`](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L262) params in the event to know which log index you are looking at for the repay type.

tnkrxyz:
> @tnkrxyz I think one of the biggest issues with the call handlers is not being supported on every chain and we need to support all chains.
> 
> One way I can think of would be:
> 
> * Watch the Repay events emitted, then go up in the event logs to see which DebtToken was burned.
> * You will have to watch the [`useATokens`](https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/BorrowLogic.sol#L262) params in the event to know which log index you are looking at for the repay type.

Yes, this can work! 

I ended up [adding an _AccountDebtBalance helper entity](https://github.com/messari/subgraphs/pull/1993/commits/f249cf65aab2b0d38135d6ed8ebeb2e9dfd3e244) to keep track of vToken and sToken balance for borrower account in the market. Inside handleRepay and handleLiquidation, I check which balance(s) goes down. For repay, only one type of debt can be repaid in a repay call and it should be the one with lower balance. Similarly for liquidation, but here, both token balances can go down. I am running test deployment to see if it works as expected.

subgraphs/aave-forks/protocols/aave-v3/README.md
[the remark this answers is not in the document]
tnkrxyz: This is now addressed in https://github.com/messari/subgraphs/pull/1993/commits/4421fdcc81003bfedcf6730fb57da6249c4b98f7

dmelotik:
> For repay, only one type of debt can be repaid in a repay call and it should be the one with lower balance

Doesn't the repayer specify in the call function parameters which side is being paid? So it could be either?

tnkrxyz:
> > For repay, only one type of debt can be repaid in a repay call and it should be the one with lower balance
> 
> Doesn't the repayer specify in the call function parameters which side is being paid? So it could be either?

@dmelotik yes, correct. I see that my sentence above may be confusing, what I meant is that, for each repay event, it is either Variable or Stable debt that's been repaid, not both types in one event. And the type should correspond to the (v or s) token with the lower balance. For liquidation, both v & s token (if both > 0) can be repaid/liquidated within the same event. sorry for the confusion.

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
```diff
@@ -350,21 +288,76 @@ export function handleRepay(event: Repay): void {
...
     event.params.reserve,
-    getProtocolData(),
+    protocolData,
     event.params.user
   );
 }
 
 export function handleLiquidationCall(event: LiquidationCall): void {
+  // if liquidator chooses to receive AToken, create a position for liquidator
+  let createLiquidatorPosition = false;
```
dmelotik, line 298: I am struggling to find in the aave contracts.

But when a `createLiquidatorPosition` is true does it send out a `BalanceTransfer` event? I am asking for my understanding and to make sure we are not double counting anything.

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
```diff
@@ -443,8 +439,122 @@ function getAssetPriceInUSDC(
   }
 
   // last resort, should not be touched
-  const inputToken = getOrCreateToken(tokenAddress);
+  const inputToken = Token.load(tokenAddress)!;
```
dmelotik, line 442: Even though this should never be reached I want to be stricter about leaving potential unwanted crashes, so we should handle the null exception

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
tnkrxyz: good catch! I believe there is a Transfer (and BalanceTransfer) event emitted when the AToken in the liquidation is transferred from liquidatee to liquidator. So we don't ever need to set createLiquidatorPosition to true for aave-forks. fixed in https://github.com/messari/subgraphs/pull/1993/commits/e3100a3be93625a2b3f379bead26bb9c85fed253

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
tnkrxyz: this is part of the code I didn't touch in the refactoring, but what you suggest makes sense.

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
dmelotik: If `createLiquidatorPosition` is always false, why leave the logic in there?

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
```diff
@@ -350,21 +288,77 @@ export function handleRepay(event: Repay): void {
...
+}
+
+export function handleSwapBorrowRateMode(event: Swap): void {
+  const interestRateMode = event.params.rateMode.toI32();
+  if (
+    ![InterestRateMode.STABLE, InterestRateMode.VARIABLE].includes(
+      interestRateMode
+    )
+  ) {
+    return;
```
dmelotik, line 339: probably want to throw an error log here. This would warrant code changes if aave introduced another mode

dmelotik, line 339: probably want to throw an error log here. This would warrant code changes if aave introduced another mode

dmelotik, line 339: probably want to throw an error log here. This would warrant code changes if aave introduced another mode

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -95,39 +124,82 @@ export function _handleReserveInitialized(
...
+    underlyingToken,
+    event,
+    protocolData
+  );
+  const market = manager.getMarket();
+  const outputTokenManager = new TokenManager(outputToken, event);
+  const vDebtTokenManager = new TokenManager(
+    variableDebtToken,
+    event,
+    TokenType.NON_REBASING
```
dmelotik, line 143: aTokens, variableDebtTokens, and stableDebtTokens are REBASING

subgraphs/aave-forks/schema.graphql
```diff
@@ -1,5 +1,5 @@
-# Subgraph Schema: Lending Protocol (Extended)
-# Version: 2.0.1
+# Subgraph Schema: Lending Protocol
+# Version: 3.0.2
```
dmelotik, line 2: Let's not update the version since the additions are just new enums.

Could you add the new enum fields to the overall schema and add them to the SDK reference for lending?

subgraphs/aave-forks/src/helpers.ts
```diff
@@ -1059,14 +97,125 @@ export function getBorrowBalance(
...
+    market.isActive,
+    market.canUseAsCollateral,
+    market.canBorrowFrom,
+  ];
+  market.save();
+}
+
+export function restorePrePauseState(market: Market): void {
+  if (!market._prePauseState || market._prePauseState!.length !== 3) {
+    log.warning(
```
dmelotik, line 180: log an error instead. Something would probably be wrong with the code if this happened

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -148,118 +220,159 @@ export function _handleCollateralConfigurationChanged(
...
+      [asset.toHexString()]
+    );
+    return;
+  }
+
+  market._liquidationProtocolFee = liquidationProtocolFee
     .toBigDecimal()
-    .div(exponentToBigDecimal(INT_TWO));
+    .div(exponentToBigDecimal(INT_FOUR));
+  log.info(
```
dmelotik, line 325: remove?

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +511,590 @@ export function _handleReserveDataUpdated(
...
-  market.totalDepositBalanceUSD = market.inputTokenBalance
-    .toBigDecimal()
-    .div(exponentToBigDecimal(inputToken.decimals))
-    .times(assetPriceUSD);
-  market.totalValueLockedUSD = market.totalDepositBalanceUSD;
-
   // calculate new revenue
   // New Interest = totalScaledSupply * (difference in liquidity index)
+  if (!market._liquidityIndex) {
+    market._liquidityIndex = BIGINT_ONE;
```
dmelotik, line 547: Should really be set to RAY: https://github.com/aave/protocol-v2/blob/master/contracts/protocol/libraries/logic/ReserveLogic.sol#L173

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +511,590 @@ export function _handleReserveDataUpdated(
...
-  );
-  if (positionId === null) {
-    log.warning(
-      "[handleRepay] Position not found for account: {} in transaction; {}",
-      [accountID.toHexString(), event.transaction.hash.toHexString()]
-    );
-    return;
+  // since there may be interest accrued, the balance difference may not exactly
+  // equal repay amount. Here we check whether the balance goes down
+  if (
```
dmelotik, line 806: I am still trying to figure out if this works?

here is a scenario:
- You borrow 100 USDC variable
- It generates 50USDC in interest, but you have not updated the position ever
- You repay 10 USDC variable
- The new balance would be 140USDC - 100USDC and you would not get an interest rate type

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +511,590 @@ export function _handleReserveDataUpdated(
...
-    protocol.save();
+  const amountUSD = amount
+    .toBigDecimal()
+    .div(exponentToBigDecimal(inputToken.decimals))
+    .times(inputTokenPriceUSD);
+
+  // according to logic in _calculateAvailableCollateralToLiquidate()
+  // liquidatedCollateralAmount = collateralAmount - liquidationProtocolFee
+  // liquidationProtocolFee = bonusCollateral * liquidationProtocolFeePercentage
+  // bonusCollateral = collateralAmount - collateralAmount / liquidationBonus
```
dmelotik, line 872: typo? collateralAmount - collateralAmount

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +511,590 @@ export function _handleReserveDataUpdated(
...
-  market.save();
+  const tokenManager = new TokenManager(asset, event);
+  const amountUSD = tokenManager.getAmountUSD(amount);
+  manager.createFlashloan(asset, account, amount, amountUSD);
+
+  let reserveFactor = market.reserveFactor;
+  if (!reserveFactor) {
+    reserveFactor = BIGDECIMAL_ZERO;
+  }
+  const protocolRevenueShare = reserveFactor.div(BIGDECIMAL_HUNDRED);
```
dmelotik, line 1033: isn't reserve factor already normalized to a decimal and not percentage?

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +511,590 @@ export function _handleReserveDataUpdated(
...
+  // save premiumUSDToDeduct to be deducted from total revenue to avoid double counting
+  flashloanPremium.premiumUSDToDeduct = premiumUSDToDeduct;
+  flashloanPremium.save();
 }
 
 /////////////////////////
 //// Transfer Events ////
 /////////////////////////
-
 export function _handleTransfer(
```
dmelotik, line 1088: won't we need to add a flag to this for the debt type since we can have transfers of stokens and vtokens and they affect different pos entities

subgraphs/aave-forks/src/helpers.ts
```diff
@@ -1059,14 +97,125 @@ export function getBorrowBalance(
...
-    const sTokenContract = AToken.bind(Address.fromString(market._sToken!));
+    const sTokenContract = AToken.bind(Address.fromBytes(market._sToken!));
     const trySDebtTokenBalance = sTokenContract.try_balanceOf(account);
     sDebtTokenBalance = trySDebtTokenBalance.reverted
       ? BIGINT_ZERO
       : trySDebtTokenBalance.value;
   }
 
-  const totalDebt = sDebtTokenBalance.plus(vDebtTokenBalance);
+  // const totalDebt = sDebtTokenBalance.plus(vDebtTokenBalance);
```
dmelotik, line 107: remove

subgraphs/aave-forks/src/helpers.ts
```diff
@@ -1059,14 +97,125 @@ export function getBorrowBalance(
...
+  const balanceResult = aTokenContract.try_balanceOf(account);
+  if (balanceResult.reverted) {
+    log.warning(
+      "[getCollateralBalance]failed to get aToken {} balance for {}",
+      [market.outputToken!.toHexString(), account.toHexString()]
+    );
+    return collateralBalance;
+  }
+
+  let exchangeRate = BIGDECIMAL_ONE;
```
dmelotik, line 124: excahnge rate is always 1 in aave since all of the aTOkens and debt tokens are rebasing

subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
```diff
@@ -629,195 +607,180 @@ function getAssetPriceFallback(
...
-
-  const sorter: Array<Array<string>> = [];
-  for (let i = 0; i < ref.length; i++) {
-    sorter[i] = [ref[i], arr1[i].toString(), arr2[i].toString()];
-  }
-
-  sorter.sort(function (a: Array<string>, b: Array<string>): i32 {
-    if (a[0] < b[0]) {
-      return -1;
+function getIsIsolatedFlag(event: ethereum.Event): boolean {
```
dmelotik, line 705: should this be in the src/helpers.ts file?

dmelotik (commented):
Looking good @tnkrxyz !

Most comments are small, the biggest one that I think still needs to be fleshed out is the transfers for different debt tokens and the repays.

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
tnkrxyz: good point! It is not needed for aave-forks. Do you think it makes sense to merge this upstream to the sdk, as it may be useful for other protocols?

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
tnkrxyz: fixed.

tnkrxyz: same as above?

tnkrxyz: same as above?

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: fixed

subgraphs/aave-forks/schema.graphql
tnkrxyz: done.

subgraphs/aave-forks/src/helpers.ts
tnkrxyz: fixed

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: done

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: good catch! fixed

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: good point. I will look for an alternative implementation - probably using `Transfer()` event as you suggested

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: it is correct `collateralAmount - (collateralAmount / liquidationBonus)`. I will reformulate.

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: good catch. I believe it has been represented as percentage in current 2.0.1 implementation: https://github.com/messari/subgraphs/blob/8d0999ea1f259e90e93a41725d73c49bfd79220a/subgraphs/aave-forks/src/mapping.ts#L217-L219

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: I think you're right with my accountDebtBalance implementation. If we move to use Transfer event, then this would be as it is, correct?

subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
tnkrxyz: moved.

subgraphs/aave-forks/src/helpers.ts
tnkrxyz: simplified.

subgraphs/aave-forks/src/helpers.ts
tnkrxyz: removed

tnkrxyz:
Will do another push once I implement the new method for detecting intereateRateType for Repay / liquidatations

dmelotik:
@tnkrxyz I noticed 2 other things, pricing in aave v3 ethereum looks inflated, maybe something changed with our pricing logic? And If you query for revenue detail I think the 2 arrays are not in parallel. I think this is actually an issue with the SDK that it is not inserting the revenue in the correct order.

tnkrxyz:
> @tnkrxyz I noticed 2 other things, pricing in aave v3 ethereum looks inflated, maybe something changed with our pricing logic? And If you query for revenue detail I think the 2 arrays are not in parallel. I think this is actually an issue with the SDK that it is not inserting the revenue in the correct order.

@dmelotik , which subgraph(s) you are looking at? I don't think I changed the pricing logic in the refactoring - the only major change is about the oracle. since 3.0.1 moved the oracle to market entity, for price query not tied to a market, I added a _defaultOracle helper, but I cannot think of any reason this would change the price.

For revenue, can it be because we now include flash loan and liquidation revenue?

tnkrxyz:
@dmelotik I pushed an alt way to determine interest rate type for repay/liquidations in [5954c03](https://github.com/messari/subgraphs/pull/1993/commits/5954c03c976894a684560e5c3feb36530415b48e). please take a look when you get a chance.

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
dmelotik: If there is a standard way you can implement, I am weary about there being a standard implementation. Since liquidations may pay in market output tokens or in the input token it adds more complexity

subgraphs/aave-forks/src/mapping.ts
dmelotik: Yeah I think I never added this into the SDK when I originally made it, but there should be an optional field in `createTransfer` for the type... Right?

subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
dmelotik: good point, resolving

dmelotik:
> I pushed an alt way to determine interest rate type for repay/liquidations

Don't you need to move up in the event log tree to see the 5th previous log and then check if the event.address of that log matches the stable or variable debt token? That is what I had in my mind, but we should test this and see if it is working.

> which subgraph(s) you are looking at?

The revenue on Aave V3 Ethereum is way higher than the TVL. Maybe on the liquidation revenue we are not normalizing something in native terms??

subgraphs/aave-forks/protocols/aave-amm/src/mapping.ts
tnkrxyz: ok, will remove.

tnkrxyz:
> > I pushed an alt way to determine interest rate type for repay/liquidations
> 
> Don't you need to move up in the event log tree to see the 5th previous log and then check if the event.address of that log matches the stable or variable debt token? That is what I had in my mind, but we should test this and see if it is working.
> 

That is what I tried in my first attempt, but it is tricker than I had thought, while the Transfer event is 5 before Repay; the Transfer event can be either 9 or 11 before LiquidationCall:
e.g. in this [v2 tx](https://polygonscan.com/tx/0x2578371e35a1fa146aa34ea3936678c20091a08d55fd8cb46e0292fd95fe7f60#eventlog) the Transfer is at logIndex 296 and LiquidationCall 305, however in this [v3 tx](https://optimistic.etherscan.io/tx/0xea6110df93b6581cf47b7261b33d9a1ae5cc5cac3b55931a11b7843641780958#eventlog) the Transfer is at 1 and LiquidationCall 12. And because liquidation can burn both Variable and Stable token in the same transactions, it became even trickier to be sure of the difference in logIndex. open to any suggestion how best to address this.


> > which subgraph(s) you are looking at?
> 
> The revenue on Aave V3 Ethereum is way higher than the TVL. Maybe on the liquidation revenue we are not normalizing something in native terms??

look into this

tnkrxyz:
> @tnkrxyz I think the liquidation logic is easy (this is aave v3): https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/LiquidationLogic.sol#L322-L349
> 
> They burn variable first until none left then stable.

Not sure I am following here. maybe I missed something. In a liquidation, I should always call subtractPosition on variable debt position if its position balance > 0. 

> 
> So the logic just needs to sift through events for repay



> @tnkrxyz I think the liquidation logic is easy (this is aave v3): https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/LiquidationLogic.sol#L322-L349
> 
> They burn variable first until none left then stable.

good point. Is this what you have in mind: https://github.com/messari/subgraphs/pull/1993/commits/4f5ed831c85545ffd57c81620a3cce5cff974e5c
 
> So the logic just needs to sift through events for repay

Refactored in https://github.com/messari/subgraphs/pull/1993/commits/9b56a26bb2b2dff3f00effa6500f182bef1f8b7f

dmelotik:
This looks good! Could you redeploy and fix the lint error?

Not sure if you have used this feature yet, but if you deploy with `messari b -d --base aave-forks` then type in `aave-forks` on the next prompt it does all the deployments at once

tnkrxyz:
> This looks good! Could you redeploy and fix the lint error?
> 
> Not sure if you have used this feature yet, but if you deploy with `messari b -d --base aave-forks` then type in `aave-forks` on the next prompt it does all the deployments at once

yes, will do both after I figure out & fix the bug in revenue.

tnkrxyz:
@dmelotik , want to see if you have any idea of a very odd behavior I ran into when I was debugging the exceptionally high revenue in aave-v3-ethereum. It is caused by market entity not being saved in `_handleReserveDataUpdated`, thus market._liquidityIndex was never saved. However, when I added [market.save at line 552](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#L552), it didn't work, [the revenue is still exceptionally high like before](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum-552&tab=protocol) and [the query](https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum-552/graphql?query=query+MyQuery+%7B%0A++markets+%7B%0A++++id%0A++++name%0A++++_liquidityIndex%0A++%7D%0A++revenueDetails+%7B%0A++++id%0A++++amountsUSD%0A++++sources+%7B%0A++++++type%0A++++%7D%0A++%7D%0A++fees+%7B%0A++++rate%0A++++type%0A++%7D%0A++flashLoanPremiums+%7B%0A++++premiumRateToProtocol%0A++++premiumRateTotal%0A++++premiumUSDToDeduct%0A++%7D%0A++flashloans%28orderBy%3A+feeAmountUSD%2C+orderDirection%3A+desc%29+%7B%0A++++amount%0A++++amountUSD%0A++++feeAmount%0A++++feeAmountUSD%0A++++timestamp%0A++++asset+%7B%0A++++++symbol%0A++++%7D%0A++++hash%0A++%7D%0A++financialsDailySnapshots+%7B%0A++++dailyBorrowUSD%0A++++dailyDepositUSD%0A++++dailyFlashloanUSD%0A++++dailyLiquidateUSD%0A++++dailyProtocolSideRevenueUSD%0A++++dailyRepayUSD%0A++++dailySupplySideRevenueUSD%0A++++dailyTotalRevenueUSD%0A++++revenueDetail+%7B%0A++++++amountsUSD%0A++++++sources+%7B%0A++++++++type%0A++++++%7D%0A++++++id%0A++++%7D%0A++%7D%0A%7D) shows _liquidityIndex is always at the initial value of 1 RAY for all markets.

image


However, if I moved `market.save()` to [Line 647](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#L647) (and commented out 552), [the revenues now look right](https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum-647&tab=protocol) and the [market._liquidityIndex query has expected values](https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum-647/graphql?query=query+MyQuery+%7B%0A++markets+%7B%0A++++id%0A++++name%0A++++_liquidityIndex%0A++%7D%0A++revenueDetails+%7B%0A++++id%0A++++amountsUSD%0A++++sources+%7B%0A++++++type%0A++++%7D%0A++%7D%0A++fees+%7B%0A++++rate%0A++++type%0A++%7D%0A++flashLoanPremiums+%7B%0A++++premiumRateToProtocol%0A++++premiumRateTotal%0A++++premiumUSDToDeduct%0A++%7D%0A++flashloans%28orderBy%3A+feeAmountUSD%2C+orderDirection%3A+desc%29+%7B%0A++++amount%0A++++amountUSD%0A++++feeAmount%0A++++feeAmountUSD%0A++++timestamp%0A++++asset+%7B%0A++++++symbol%0A++++%7D%0A++++hash%0A++%7D%0A++financialsDailySnapshots+%7B%0A++++dailyBorrowUSD%0A++++dailyDepositUSD%0A++++dailyFlashloanUSD%0A++++dailyLiquidateUSD%0A++++dailyProtocolSideRevenueUSD%0A++++dailyRepayUSD%0A++++dailySupplySideRevenueUSD%0A++++dailyTotalRevenueUSD%0A++++revenueDetail+%7B%0A++++++amountsUSD%0A++++++sources+%7B%0A++++++++type%0A++++++%7D%0A++++++id%0A++++%7D%0A++%7D%0A%7D)
image

This is very strange and unexpected to me. Why where `market.save()` is called makes any difference? What is even more strange is that with `market.save()` at Line 647, the TVL is now way off, while it looks right with market.save() at Line 552. I appreciate any suggestion you may have.

image

image

tnkrxyz:
> Sorry, but a few more requests (aave v3/forks):
> 
> * Can you watch the events to update `position.isIsolated`? I think it is `LendingPool_EVENT_IsolationModeTotalDebtUpdated()`

IsolatedMode is already handled in [handleBorrow](https://github.com/tnkrxyz/subgraphs/blob/1f94716fd01ebc25f4aab8ea16a6ea9d9cf2bb75/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts#L329-L343). Is there anything else needs to be done?

> * Could you add a field `position._eMode` defaulted to false (only updatable on collateral side)? Updated with `LendingPool_EVENT_UserEModeSet()`
> * Could you also add a field `market._siloedBorrowing` defaulted to false? Updated with `LendingConfigurator_EVENT_SiloedBorrowingChanged()`

done in 1f94716

dmelotik:
@tnkrxyz you are right, I think I misunderstood isolation mode. To exit you simply need to repay the debt position: https://docs.aave.com/developers/whats-new/isolation-mode#exit-isolation-mode

subgraphs/aave-forks/src/mapping.ts
dmelotik: @tnkrxyz actually `manager.createTransfer()` does have an optional field for `interestRateType` why do we not use this here?

dmelotik:
@tnkrxyz I think I found the issue!

- You load a market at the beginning of the function [here](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#LL464C18-L464C36)
- Then you create an instance of manager [here](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#L472) which also loads a market
- Now you are updating 2 market entities at once and 1 is overriding the other and updates aren't happening as expected

To fix:
- you should create the `manager` instance first, then use `manager.getMarket()` to get the market entity if you need to make changes outside of the SDK (for the helper fields). See comp v3 for an example: https://github.com/messari/subgraphs/blob/master/subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts#L101-L109
- So likely `getMarketFromToken()` will have to be changed to return the marketID and inputTokenID and not the market entity
- I see the incorrect pattern being used in multiple places. The fix should be consistent throughout

tnkrxyz:
> @tnkrxyz I think I found the issue!
> 
> * You load a market at the beginning of the function [here](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#LL464C18-L464C36)
> * Then you create an instance of manager [here](https://github.com/tnkrxyz/subgraphs/blob/286ef91bbff497b8f01ab3d2c556868113b970e4/subgraphs/aave-forks/src/mapping.ts#L472) which also loads a market
> * Now you are updating 2 market entities at once and 1 is overriding the other and updates aren't happening as expected
> 
> To fix:
> 
> * you should create the `manager` instance first, then use `manager.getMarket()` to get the market entity if you need to make changes outside of the SDK (for the helper fields). See comp v3 for an example: https://github.com/messari/subgraphs/blob/master/subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts#L101-L109
> * So likely `getMarketFromToken()` will have to be changed to return the marketID and inputTokenID and not the market entity
> * I see the incorrect pattern being used in multiple places. The fix should be consistent throughout

This is fixed in [691c6c7](https://github.com/messari/subgraphs/pull/1993/commits/691c6c723db103f14bed6f7e785744a04455df60). Funny enough, _handleReserveDataUpdated is the only place I saved market entity returned from getMarketFromToken().

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: fixed.

tnkrxyz:
@dmelotik , ok, this PR is ready for (hopefully) final review and merge.

dmelotik:
I am seeing crazy `depositUSD`/`borrowUSD` prices

Quick fix: we just need to normalize `amountUSD` with `inputToken.decimals` in each type of transaction: https://github.com/messari/subgraphs/blob/6b7078b10f679e56519aca22b84780ad8c709871/subgraphs/aave-forks/src/mapping.ts#L672

I am also noticing the new subgraph has less `borrowCount` and no `RepayCount` which indicates to me that positions are not being matched up in repays (and that is what looks like is happening in the logs)

Something I can tell is wrong: In both examples of checking for transfer events to get the interest rate type in Repay there are approvals that occur. But that is not always the case. In this example the Transfer event we want to see is only 4 before and not 5. To account for this we may want to search for the `Transfer` event in a range of 4-5.

However, it should still pickup the ones that are 5 away, yet there are no repays

tnkrxyz:
> I am seeing crazy `depositUSD`/`borrowUSD` prices
> 
> Quick fix: we just need to normalize `amountUSD` with `inputToken.decimals` in each type of transaction:
> 
> https://github.com/messari/subgraphs/blob/6b7078b10f679e56519aca22b84780ad8c709871/subgraphs/aave-forks/src/mapping.ts#L672
> 
> I am also noticing the new subgraph has less `borrowCount` and no `RepayCount` which indicates to me that positions are not being matched up in repays (and that is what looks like is happening in the logs)
> 
> Something I can tell is wrong: In both examples of checking for transfer events to get the interest rate type in Repay there are approvals that occur. But that is not always the case. In this example the Transfer event we want to see is only 4 before and not 5. To account for this we may want to search for the `Transfer` event in a range of 4-5.
> 
> However, it should still pickup the ones that are 5 away, yet there are no repays

looking into these.

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -148,118 +222,157 @@ export function _handleCollateralConfigurationChanged(
...
-    account.save();
+  market._liquidationProtocolFee = liquidationProtocolFee
+    .toBigDecimal()
+    .div(exponentToBigDecimal(INT_FOUR));
+  market.save();
+}
 
-    protocol.cumulativeUniqueUsers += 1;
-    protocol.save();
+export function _handleReserveUsedAsCollateralEnabled(
```
dmelotik, line 332: These functions need to also set the `position.isCollateral` field

subgraphs/aave-forks/src/constants.ts
```diff
@@ -103,6 +93,7 @@ export const BIGINT_ONE = BigInt.fromI32(1);
 export const BIGINT_THREE = BigInt.fromI32(3);
 export const BIGINT_TEN = BigInt.fromI32(10);
 export const BIGINT_TEN_TO_EIGHTEENTH = BigInt.fromString("10").pow(18);
+export const BIGINT_ONE_RAY = BigInt.fromString("10").pow(27);
```
dmelotik, line 96: this should be 1^27, that is why we are getting neg rev in some pools

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: I believe isCollateral is handled here: https://github.com/tnkrxyz/subgraphs/blob/0b84b4042ac0a8c9f16c38502f26166fce4ece89/subgraphs/aave-forks/src/mapping.ts#L694-L703

subgraphs/aave-forks/src/constants.ts
tnkrxyz: isn't 1^27 = 1? :grin:

subgraphs/aave-forks/src/mapping.ts
dmelotik: See this [query](https://api.thegraph.com/subgraphs/name/tnkrxyz/aave-v3-ethereum/graphql?query=query+MyQuery+%7B%0A++positions%28where%3A+%7BisCollateral%3A+true%7D%29+%7B%0A++++id%0A++++side%0A++++type%0A++++balance%0A++++isCollateral%0A++%7D%0A%7D)

tnkrxyz: there is a bug in checking _enabledCollaterals array. fixed in 
104d24d

[query deployment with fix](https://api.thegraph.com/subgraphs/id/QmdVZ8gguWt6kcCijoYGW9suF59JXsTdAQvDR37cua7N42/graphql?query=query+MyQuery+%7B%0A++positions%28where%3A+%7BisCollateral%3A+true%7D%29+%7B%0A++++id%0A++++side%0A++++type%0A++++balance%0A++++isCollateral%0A++%7D%0A%7D)

subgraphs/aave-forks/src/constants.ts
dmelotik: @tnkrxyz any word on this?

tnkrxyz:
@dmelotik , the negative revenue is caused by the deduction of flash loan premium from totalRevenue in handleReserveDataUpdate, but I am not sure why LiquidityIndex in reserveDataUpdate event does not include the flashloan premium to LP as we believed. 

For example, according to these subgraph log
```
  {
    "timestamp": "2023-06-13T23:25:09.137907068Z",
    "text": "[_handleFlashLoan]premiumAmount=30000000000000000000 premiumUSDTotal=52620.9987 premiumUSDToProtocol=7914.19820448 premiumUSDToLP=44706.80049552 premiumUSDToDeduct=52599.95030052 market=0x4d5f47fa6a74757f35c14fd3a6ef8e3c9bc514e8 tx=0xeb87ebc0a18aca7d2a9ffcabf61aa69c9e8d3c6efade9e2303f8857717fb9eb7-809 timestamp=1686531995, data_source: LendingPool, component: UserMapping"
  },
  {
    "timestamp": "2023-06-13T23:25:09.153314086Z",
    "text": "Done processing trigger, gas_used: 11637673778, data_source: LendingPool, handler: handleFlashloan, total_ms: 83, transaction: 0xeb87…9eb7, address: 0x8787…a4e2, signature: FlashLoan(indexed address,address,indexed address,uint256,uint8,uint256,indexed uint16)"
  },
  {
    "timestamp": "2023-06-13T23:25:09.168726304Z",
    "text": "Applying 52 entity operation(s), block_hash: 0x101132f147d5d6103cfdcb3dbad67577a3486f9fb7de632a21778d71701b0360, block_number: 17460610"
  },
  {
    "timestamp": "2023-06-13T23:25:09.253221374Z",
    "text": "[_handleReserveDataUpdated]totalRevenueDeltaUSD=-52537.48083701298829788585188957448 protocolSideRevenueDeltaUSD=-5253.748083701298829788585188957448 supplySideRevenueDeltaUSD=-47283.732753311689468097266700617032 flashloandPremiumToDeduct=52599.95030052 reserveFactor
=0.1 market=0x23878914efe38d27c4d67ab83ed1b93a74d4086a tx=0x1a520aea1fa610109a332a7408aa8d595fd83a40328b4709438937ee83c5b989-256 timestamp=1686532175, data_source: LendingPool, component: UserMapping"
  },
```

At flashloan tx 0xeb87ebc0a18aca7d2a9ffcabf61aa69c9e8d3c6efade9e2303f8857717fb9eb7, the premium amount is 30 WETH, or $52620.9987, of which $52599.95030052 goes to the LP and supposedly should be deducted from total revenue. However, the ReserveDataUpdated event followed (0x1a520aea1fa610109a332a7408aa8d595fd83a40328b4709438937ee83c5b989), the calculated total revenue from LiquidityIndex is only $62.46. Revenue became negative after deducting flash loan premium of $52599.95030052.

From what I can see, the premium to LP is accrued to LiquidityIndex [in the smart contract](https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/FlashLoanLogic.sol#L233). I am not sure what is wrong and leads to the revenue not included in our totalRevenue. Any idea?

dmelotik:
Let me ask the aave team. This is something we need to ensure we have right. And it may be the case that we don't need to subtract the revenue

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +513,584 @@ export function _handleReserveDataUpdated(
...
 
-  // update usage metrics
-  snapshotUsage(
-    protocol,
-    event.block.number,
-    event.block.timestamp,
-    liquidate.liquidatee,
-    EventType.LIQUIDATEE,
-    true // only count this liquidate as new tx
+  const feeToProtocol = manager.getOrUpdateFee(
```
dmelotik, line 1059: isnt this updated in the event that updates this?

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +513,584 @@ export function _handleReserveDataUpdated(
...
+
+  if (flashloanPremium.premiumRateToProtocol.gt(BIGDECIMAL_ZERO)) {
+    // according to https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/interfaces/IPool.sol#L634
+    // premiumRateToProtocol is the percentage of premium to protocol
+    premiumUSDToProtocol = premiumUSDTotal
+      .times(flashloanPremium.premiumRateToProtocol)
+      .plus(premiumUSDToProtocol);
+    premiumRateToProtocol = premiumRateTotal
+      .times(flashloanPremium.premiumRateToProtocol)
+      .plus(premiumRateToProtocol);
```
dmelotik, line 1047: I think these don't need to be multiplied by the rate...

Once you get `premiumUSDToProtocol ` and `premiumUSDToLP` that is all you need.

`premiumUSDToDeduct` = `premiumUSDToLP` (https://github.com/aave/aave-v3-core/blob/29ff9b9f89af7cd8255231bc5faf26c3ce0fb7ce/contracts/protocol/libraries/logic/FlashLoanLogic.sol#L236)

And `premiumUSDToProtocol` is protocol rev and `premiumUSDToLP` is supply rev.

subgraphs/aave-forks/src/mapping.ts
```diff
@@ -372,707 +513,584 @@ export function _handleReserveDataUpdated(
...
+  const flashloan = manager.createFlashloan(asset, account, amount, amountUSD);
+  flashloan.feeAmount = premiumAmount;
+  flashloan.feeAmountUSD = premiumUSDTotal;
+  flashloan.save();
+
+  let reserveFactor = market.reserveFactor;
+  if (!reserveFactor) {
+    reserveFactor = BIGDECIMAL_ZERO;
+  }
+  const protocolRevenueShare = reserveFactor;
```
dmelotik, line 1031: this should be https://etherscan.io/address/0x87870bca3f3fd6335c3f4ce8392d69350b4fa4e2#readProxyContract#F4

~ 4% rn which means premiumUSDToPortocol in the example transaction should be ~$2090

And PremiumToLP ~$50,169.6

tnkrxyz: @dmelotik , thanks for helping looking to this. I don't think that's it. The totalRevenueDelta at the ReserveDataUpdate event in my example tx is about $62, way lower than $50,169 and we will still get a very negative revenue.

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: not sure I understand what you mean here.

subgraphs/aave-forks/src/mapping.ts
tnkrxyz: I did this because I want to be precise in the split of premium between LP and protocol. The flashloan premium to LP accrued to LiquidityIndex is split between LP and protocol based on reserve factor. my calculation here include the protocol's share of flashloan premium to LP as protocol revenue.

dmelotik:
@tnkrxyz A few things I found:

- I looked at the largest flashloan in aave v3 ethereum ([here](https://etherscan.io/tx/0xeb87ebc0a18aca7d2a9ffcabf61aa69c9e8d3c6efade9e2303f8857717fb9eb7)) and there is a spike in revenue in the WETH market on the same day. It is definitely getting distributed to LPs through the liquidity Index. I just think we are doing something wrong in the subgraph code now.
- Aave V2 also handles this differently, so we will need to have a way to switch between the different logics in flashloan distribution

tnkrxyz:
> @tnkrxyz A few things I found:
> 
> * I looked at the largest flashloan in aave v3 ethereum ([here](https://etherscan.io/tx/0xeb87ebc0a18aca7d2a9ffcabf61aa69c9e8d3c6efade9e2303f8857717fb9eb7)) and there is a spike in revenue in the WETH market on the same day. It is definitely getting distributed to LPs through the liquidity Index. I just think we are doing something wrong in the subgraph code now.
> * Aave V2 also handles this differently, so we will need to have a way to switch between the different logics in flashloan distribution

@dmelotik , I think I finally figured out what is causing this and am testing a fix. more update soon.

tnkrxyz:
@dmelotik , ok, here is what I found so far: the reason we are getting negative revenue is due to mis-match of FlashLoan event and ReserveDataUpdated event in the current implementation. The ReserveDataUpdated event associated (with the updated liquidityIndex) with a FlashLoan event is emitted before the FlashLoan event. My current code assumes the ReserveDataUpdated event is after the  FlashLoan event, which leads to the accrued LP revenue lower than flashloan premium.

I implemented a fix and it fixes the large negative revenue we have seen, but it still has some tiny negative revenues (the most negative number I see is totalRevenueDeltaUSD=$-0.00106344452993117741167457972125, most is about $-0.0002 to $-0.0005; see [this query](https://api.thegraph.com/explorer/graphql?query=%7B%0A%20%20subgraphLogs%28%0A%20%20%20%20subgraphId%3A%20%22QmafExARHcWzFkytuy93o3pt7rxkgWk19JDB5piEsvpB4C%22%0A%20%20%20%20first%3A%202000%0A%20%20%20%20order%3A%20NewestFirst%0A%20%20%20%20filters%3A%20%5B%22error%22%2C%20%22warning%22%2C%20%22info%22%2C%20%22debug%22%5D%0A%20%20%20%20searchText%3A%20%22totalRevenueDeltaUSD%3D%5C-%22%0A%20%20%29%20%7B%0A%20%20%20%20timestamp%0A%20%20%20%20text%0A%20%20%7D%0A%7D)), which I suspect is due to rounding, but I am not certain.

dmelotik:
Nice job catching this @tnkrxyz ! I reviewed the new code and it looks good, makes sense to me.

I would think those small, small discrepancies are also due to rounding. I see 2 ways to go about it.
- Either we just leave it since it doesn't actually make any unexpected values
- We make any small neg values 0

I would say 3 things before we merge:
- Solve merge conflicts (apologies for that)
- Come up with a solution for the above
- Check out the versions, for some reason your latest `aave-v3-ethereum` deployment shows `schemaVersion = 2.0.1`

tnkrxyz:
> 
> I would say 3 things before we merge:
> 
> * Solve merge conflicts (apologies for that)

merged.

> * Come up with a solution for the above

`$0` > negative revenue > `$-0.01` is set to 0.

> * Check out the versions, for some reason your latest `aave-v3-ethereum` deployment shows `schemaVersion = 2.0.1`

fixed.

dmelotik (approved):
Good work @tnkrxyz !

PR #1999, opened 2023-04-11
fix(#minor); Uniswap V3 Optimism; Instantiate ticks at re-genesis

steegecs:
**Context:**
The `liquidityGross` and `liquidityNet` values on the tick entities and the `totalLiquidity` value on the liquidity pool entities were incorrect for this network because there was a re-genesis of Optimism, and all previous data was lost. Because these values are usually tracked in an accumulating manner, this breaks the calculation for Optimism. This PR addresses this issue by running contract calls at the beginning of indexing to pull these liquidity values for each tick and updating the necessary fields before continuing with the regular indexing protocol. This was an adjustment to the backfill already done with the Optimism pools, and it should not impact the data of any other chain for Uniswap or Pancakeswap. 

**Changes:**
- Improved the performance of the exponentiation utility function to avoid computational limits. 
- Added the TickLens ABI and used the contract to pull tick data in the Optimism backfill. 
- Updated the Optimism backfill process to include the instantiation of ticks with current liquidity values and the `totalLiquidity` field on liquidity pool entities that is an aggregate of these values.

**Deployment:**
https://thegraph.com/hosted-service/subgraph/steegecs/uniswap-v3-optimism?version=pending

### DIFF
--- a/subgraphs/uniswap-v3-forks/src/common/utils/backfill.ts
+++ b/subgraphs/uniswap-v3-forks/src/common/utils/backfill.ts
@@ -1,4 +1,4 @@
-import { Address, ethereum, log, BigInt } from "@graphprotocol/graph-ts";
+import { Address, ethereum, BigInt } from "@graphprotocol/graph-ts";
 import { BIGINT_TWO, BIGINT_ZERO, INT_ONE, INT_ZERO } from "../constants";
 import { Pool } from "../../../generated/Factory/Pool";
 import { TickLens } from "../../../generated/Factory/TickLens";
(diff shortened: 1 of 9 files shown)


subgraphs/uniswap-v3-forks/src/common/utils/backfill.ts
```diff
@@ -64,6 +69,42 @@ export function populateEmptyPools(event: ethereum.Event): void {
...
+        tick.liquidityNet = populatedTick.liquidityNet;
+        tick.lastUpdateTimestamp = event.block.timestamp;
+        tick.lastUpdateBlockNumber = event.block.number;
+        tick.save();
+
+        totalLiquidityGross = totalLiquidityGross.plus(
+          populatedTick.liquidityGross
+        );
+      }
+    }
```
jaimehgb, line 104: Can you add somewhere in here a link to some docs where someone could look to understand what's going on in here?

To know what 887272 means, tickBitmap, words, etc

jaimehgb (approved):

subgraphs/uniswap-v3-forks/src/common/utils/backfill.ts
steegecs: Sure! I added some docs and a comment. Let me know if that works.

PR #2001, opened 2023-04-12
Revert "chore (); additional eslint rules"

dmelotik:
Reverts messari/subgraphs#1955

I think we should consider reverting and fixing issues in the `_reference_` subgraph before we push these linter changes


tnkrxyz (approved):
approved

PR #2004, opened 2023-04-12
chore(); aura; increment subgraph version

dmelotik:
The version was not updated on the last change. We need to update it otherwise the downstream logic fails to redeploy to the decentralized network.


steegecs (approved):
LG

PR #2019, opened 2023-04-16
fix(#minor); stargate; Fix transferFrom addresses for bridge transfer-out

dhruv-chauhan:
Test deployment, 
https://okgraph.xyz/?q=dhruv-chauhan%2Fstargate-mainnet


dmelotik (approved):
lgtm!

PR #2034, opened 2023-04-20
fix(#patch): fixed an issue with calculating bridge TVE when manually calling setInputTokenBalance

jaimehgb:


steegecs (approved):
Looks good!

Issue #2007, opened 2023-04-13
Axelar - MAINNET QA


SerenityFund:
## Review Info

| Description | Info | 
| :-------- | :--- | 
| Subgraph Reviewed |  https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/axelar-ethereum&tab=protocol
| Schema Version  |  1.1.0  | 
| Subgraph Version |  1.0.0  | 
| Methodology Version  |	1.0.0 |
| Evidence Spreadsheet |  https://docs.google.com/spreadsheets/d/14m3StI-RskdRGi8uh6Qx9zbdnkScknd3/edit?usp=sharing&ouid=104027705798814439445&rtpof=true&sd=true |

### Revenue
- Axelar does not generate revenue since it does not impose fees, and only the relayer gas fees are collected. You may refer to this link: https://axelar.network/blog/cross-chain-gas-costs-on-satellite-explained
- Refer to the sheet 'Protocol'   

### Total Pool Count and Total Supported Token Count
- Axelar's official website displays that there are solely 18 tokens native to Ethereum, which contrasts with the 24 tokens listed on the subgraph.

Please refer to the attached worksheet for more details.

bye43:
@tnkrxyz QA for axelar.

bye43:
@tnkrxyz Just wanted to follow up on this QA in case it got buried!

tnkrxyz:
Response to QA comments ** WIP**

> ## Review Info

> ## Metric to Review

> ### Revenue
> * Axelar does not generate revenue since it does not impose fees, and only the relayer gas fees are collected. You may refer to this link: https://axelar.network/blog/cross-chain-gas-costs-on-satellite-explained

I am aware of this, but I remember seeing somewhere that it is possible to enable protocol fees (current set to 0). Let me see if I can find it.

> 
> ### Total Pool Count and Total Supported Token Count
> * Axelar's official website displays that there are solely 18 tokens native to Ethereum, which contrasts with the 24 tokens listed on the subgraph.
 
> Please refer to the attached worksheet for more details.

bye43:
Dashboard linke for ETH:  https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmYqtQzyWSc5LPevXUesaw4bP5VzmK6QXVH98s1H9PeNaK&tab=protocol

pietydespises:
Looks good :)

[...]

PR #2045, opened 2023-04-21
fix (#patch); axelar; fixed revenue and non-evm address

tnkrxyz:
This PR fixes the revenue issue found in the axelar QA (#2007 ), and the handle of non-evm src addresses in `handleContractCallApproved`.

- Test deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/axelar-ethereum?version=pending&selected=logs
- Dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/id/QmYqtQzyWSc5LPevXUesaw4bP5VzmK6QXVH98s1H9PeNaK&tab=protocol

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -8518,8 +8518,8 @@
         "status": "dev",
         "versions": {
           "schema": "1.2.0",
-          "subgraph": "1.0.0",
-          "methodology": "1.0.0"
+          "subgraph": "1.0.1",
+          "methodology": "1.0.1"
         },
         "files": {
           "template": "axelar.template.yaml"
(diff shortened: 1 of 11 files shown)


dmelotik (commented):
One comment. Should we remove the `AxelarGasService` contract from the manifest all together?

tnkrxyz:
> One comment. Should we remove the `AxelarGasService` contract from the manifest all together?

Good question! Personally I prefer to keep it, as we likely need it when axelar bridge starts to collect fees, but of course it isn't needed and can be removed.

dmelotik:
@tnkrxyz let's keep it then

dmelotik (approved):
Nice work

PR #2048, opened 2023-04-23
fix(#minor); perpfutures-sdk; fix pool snapshots, snapshot daily activity

dhruv-chauhan:


subgraphs/_reference_/src/sdk/protocols/perpfutures/poolSnapshot.ts
```diff
@@ -458,35 +458,20 @@ export class PoolSnapshot {
     snapshot.cumulativeUniqueUsers = this.pool.cumulativeUniqueUsers;
 
     const dailyActivityHelper = initActivityHelper(
-      Bytes.fromUTF8("daily-".concat(this.dayID.toString()))
+      Bytes.fromUTF8("daily-".concat(day.toString()))
     );
     snapshot.dailyActiveUsers = dailyActivityHelper.activeUsers;
+    snapshot.dailyActiveDepositors = dailyActivityHelper.activeDepositors;
```
dmelotik, line 464: Good catch on this!

dmelotik (approved):
Good catch!!

@bye43 we should update any perpfuture subgraph using the sdk with this. Otherwise they will not have accurate `dailyActiveDepositors` etc

dmelotik:
I will make an issue to track this @dhruv-chauhan, but you should also have permissions if you find any bugs in the future.

#2056

bye43:
cc: @ishraq8 for visibility

PR #2064, opened 2023-04-26
feat(#minor); dex schema; add reserve amount fields for dex events

dhruv-chauhan:


steegecs (commented):
Great! Can you update the subgraph and schema versions in the deployment.json?

steegecs (commented):
Would you be able to update this for the Uniswap Forks as well since you change this code? Its a lot of version updates XD.

dmelotik (approved):
Approving, I will update the other forks in another PR

PR #2094, opened 2023-05-09
chore(); allowed string literal in ethereum.decode call

tnkrxyz:
Allowing usage of string literal in `ethereum.decode()` e.g. `ethereum.decode("address", log.data)`. Otherwise, we have to go through
```
const ADDRESS = "address";
ethereum.decode(ADDRESS, log.data);
```
and adds no more readability.


dmelotik (approved):
agreed

tnkrxyz:
Unfortunately, this doesn't work. will debug (& do a new PR) when I have time.

PR #2099, opened 2023-05-10
Update schema-derivatives-perpfutures.graphql

ishraq8:
Currently `stakeSideRevenue` data is collected at the financial daily snapshots level and was not in pool level. Back-end derives all metrics from `pool snapshots` and `usage snapshots`. I added this metric into the pool level metrics. 

@dhruv-chauhan since you built the SDK, will this effect the SDK?

If approved I will put in an issue to upgrade GMX and Gains, Mummy Finance Still under review.


jaksonkallio (commented):
This looks good to me! We'd be able to start ingesting stake-side revenue as a part of `LiquidityPoolDailySnapshot`

dhruv-chauhan (approved):
lgtm!

Supporting changes for,
perpfutures sdk: PR #2105
gains-trade subgraph: PR #2103

Issue #2115, opened 2023-05-14
Goldfinch Ethereum: Indexing errors 


bye43:
ethereum: 17260147
https://okgraph.xyz/?q=messari%2Fgoldfinch-ethereum
```A determinstic fatal error occured at block 17260147: Mapping aborted at src/mappings/membership/capital_ledger.ts, line 57, column 23, with no message wasm backtrace: 0: 0x67e4 - <unknown>!src/mappings/membership/capital_ledger/handleCapitalErc721Deposit in handler `handleCapitalErc721Deposit` at block #17260147 (2056146d6e4204abc4cde3f0e447cc97c45de5d21f54213b3fb0be77958629f6)```

dmelotik:
Goldfinch added this pool https://app.goldfinch.finance/pools/0x032f7299621c3b68e5d7aceabd567b65e2284da7 in as a non-standard pool. We want to close my PR and simply hardcode this pool in, and make a note in the PR that the subgraph may fail in the future. And we may need to add it in again

bye43:
@ishraq8 whenever you get a chance can you take a look at this?

Deployment and dashboard
- deployment: https://thegraph.com/hosted-service/subgraph/tnkrxyz/goldfinch-ethereum?selected=logs
- dashboard: https://subgraphs.messari.io/subgraph?endpoint=https://api.thegraph.com/subgraphs/name/tnkrxyz/goldfinch-ethereum&tab=protocol

dmelotik:
@ishraq8 with a focus on this pool: `0x032f7299621c3b68e5d7aceabd567b65e2284da7` as it was the one we added in that was causing isssues

tnkrxyz:
> Goldfinch added this pool https://app.goldfinch.finance/pools/0x032f7299621c3b68e5d7aceabd567b65e2284da7 in as a non-standard pool. We want to close my PR and simply hardcode this pool in, and make a note in the PR that the subgraph may fail in the future. And we may need to add it in again

@dmelotik , Goldfinch added a new type of "markets" in addition to the traditional tranched pool, called CallableLoan. The `GoldfinchFactory` contract emits an `CallableLoanCreated(indexed address,indexed address)` event when a CallableLoan market is created. This is causing an indexing error because CallableLoanCreated was not handled, which is fixed in #2195 by adding a handler for CallableLoanCreated and a template for CallableLoan. CallableLoan markets are handled in a way similar to Tranched Pool.

ishraq8:
[...]

PR #2123, opened 2023-05-15
feat(#minor); goldfinch; fix runtime error

dmelotik:
#2115

Fixes runtime error caused by using `assert()` on a null entity.

This is an uncommon error handling procedure in our repo. So I went through the rest of the code and got rid of instances where `assert()` was used and replace with a more consitent `if(!entity) log error` syntax

Testing: https://okgraph.xyz/?q=dmelotik%2Fgoldfinch-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2458,7 +2458,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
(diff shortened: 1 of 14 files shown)


dmelotik:
https://github.com/messari/subgraphs/issues/2115#issuecomment-1557595469

PR #2127, opened 2023-05-16
feat(#major); hop-protocol; initial version

itsjerryokolo:
Syncing...

Test Subgraphs:
xDai : https://okgraph.xyz/?q=QmX5NmQ28AxVviv8WREC9JoFpR5gzwQKm3d9GGcnzkxEPT
Polygon : https://okgraph.xyz/?q=QmNojzQe3nFZH3sizzZS7CSSRLeF1PRardPjn9kb5pVg1L
Optimism : https://okgraph.xyz/?q=QmVR1MpaVBMAMfhiFd3W8t9jszEmTBXdkRb42zfhFtiwja
Arbitrum : https://okgraph.xyz/?q=QmbW4d6ug1P8RoQq6tu5Jy4AJeL7ngi1s4ob2rmHb3WpfC
Ethereum : https://okgraph.xyz/?q=QmTP91LxyZ8hxThCSsdNF8YsA7yZReRApQyijsNnkCf9m6


subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
```diff
@@ -0,0 +1,255 @@
...
+    log.warning("S6 - chainID: {}, inputToken: {}", [
+      event.params.chainId.toString(),
+      inputToken,
+    ]);
+
+    pool.addInputTokenBalance(event.params.amount);
+
+    let receipt = event.receipt;
+
+    if (receipt) {
```
jaimehgb, line 183: Prefer early returns over big `if` blocks :P

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/token.ts
```diff
@@ -0,0 +1,93 @@
...
+      new TokenInit(),
+      event
+    );
+
+    sdk.Tokens.getOrCreateToken(event.address);
+
+    if (event.params.from.toHexString() == bridgeAddress) {
+      const acc = sdk.Accounts.loadAccount(event.params.to);
+
+      acc.countTransferIn();
```
jaimehgb, line 90: Is this the only way to determine if an account received a transfer? Can we look at the bridge contracts instead of all the tokens?

subgraphs/hop-protocol/src/common/getters.ts
```diff
@@ -0,0 +1,249 @@
+// import { log } from "@graphprotocol/graph-ts"
```
jaimehgb, line 1: I don't think you need any of the files in `common` if you are using the sdk, right?

subgraphs/hop-protocol/protocols/hop-protocol/config/templates/hop-protocol.template.yaml
```diff
@@ -0,0 +1,389 @@
...
+          file: ./abis/Prices/Calculations/SushiSwap.json
+        - name: SushiSwapFactory
+          file: ./abis/Prices/Calculations/SushiSwap.json
+        - name: SushiSwapPair
+          file: ./abis/Prices/Calculations/SushiSwap.json
+        - name: CalculationsSushiSwap
+          file: ./abis/Prices/Calculations/SushiSwap.json
+      eventHandlers:
+        - event: Transfer(indexed address,indexed address,uint256)
+          handler: handleTransfer
```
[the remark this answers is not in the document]
itsjerryokolo: I'll remove.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
```diff
@@ -0,0 +1,255 @@
...
+      new Pricer(),
+      new TokenInit(),
+      event
+    );
+    sdk.Accounts.loadAccount(event.params.newBonder);
+  }
+}
+
+export function handleTransferSentToL2(event: TransferSentToL2): void {
+  if (event.params.chainId.toString() == "42170") return;
```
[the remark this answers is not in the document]
itsjerryokolo: This is a chain Arbitrum Nova. It broke the subgraph because that network isn't tracked. 
It isn't officially supported.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
itsjerryokolo: Okay, I'll change that.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/token.ts
itsjerryokolo: I'll remove the token contracts.

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.json
```diff
@@ -0,0 +1,143 @@
...
+    }
+  ],
+  "bridges": [
+    {
+      "dataSourceName": "HopL2Bridge",
+      "token": "USDC",
+      "address": "0x0e0E3d2C5c292161999474247956EF542caBF8dd",
+      "startBlock": 440027
+    },
+    {
```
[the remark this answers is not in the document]
itsjerryokolo: Do you mean like a factory style contract?

deployment/deployment.json
```diff
@@ -6811,94 +6811,6 @@
     "protocol": "multichain",
     "project": "multichain",
     "deployments": {
-      "multichain-arbitrum": {
```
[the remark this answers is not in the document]
itsjerryokolo: I'll check.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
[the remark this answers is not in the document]
itsjerryokolo: Left this. The mapping logic breaks if chainId is unrecognised.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    let receipt = event.receipt;
+
+    if (receipt) {
+      log.warning("S7 - chainID: {}, inputToken: {}, txHash: {}", [
+        event.params.chainId.toString(),
+        inputToken,
+        event.transaction.hash.toHexString(),
+      ]);
+      for (let index = 0; index < receipt.logs.length; index++) {
```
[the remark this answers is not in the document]
itsjerryokolo: done

deployment/context/versions.template.mustache
```diff
@@ -2,15 +2,15 @@ import { Versions as VersionsInterface } from "../../../deployment/context/inter
 
 export class VersionsClass implements VersionsInterface {
   getSchemaVersion(): string {
-    return "{{ versions.schema }}";;
+    return "{{ versions.schema }}";
```
dmelotik, line 5: Was this broken?

subgraphs/package-lock.json
dmelotik, line 1: Undo the deletion of this file

subgraphs/gmx/src/mappings/glpManager.ts
dmelotik, line 1: The changes in the gmx folder will need to be undone. That should be fixed by doing a rebase though, I think. These files had a name change since you branched off I believe

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.ts
```diff
@@ -0,0 +1,302 @@
...
+      return this.getPolygonCrossTokenFromTokenAddress(tokenAddress);
+    else if (chainId == "1")
+      return this.getMainnetCrossTokenFromTokenAddress(tokenAddress);
+    else {
+      log.critical("Chain not found", []);
+      return "";
+    }
+  }
+
+  getArbitrumCrossTokenFromTokenAddress(tokenAddress: string): string {
```
dmelotik, line 73: Is this function necessary if it doesn't do anything?

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.json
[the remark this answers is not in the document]
dmelotik: Yeah, a factory contract would make this easier and more sustainable, does hop not have a factory contract? Even if there is one on mainnet and not the L2s?

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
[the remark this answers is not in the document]
dmelotik: Can we add this chain to the `Network` enum in the schema and then support transfers to that chain? Without it we are missing TVL exported there and netVolume.

As a side note: If we were to leave this in, It would be good to add a comment in the code explaining this early return, and to make the `42170` portion a constant

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
```diff
@@ -0,0 +1,255 @@
...
+      event
+    );
+    sdk.Accounts.loadAccount(event.params.newBonder);
+  }
+}
+
+export function handleTransferSentToL2(event: TransferSentToL2): void {
+  if (event.params.chainId.toString() == "42170") return;
+
+  if (NetworkConfigs.getBridgeList().includes(event.address.toHexString())) {
```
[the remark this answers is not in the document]
dmelotik: I think this goes back to the factory conversation. If we cannot track a factory contract / use templates, then yes we should probably keep this in there.

However, a better logic flow could be:
```typescript 
if (!NetworkConfigs.getBridgeList().includes(event.address.toHexString())) {
    log.error("Missing config");
    return;
}

... then continue func
```

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
dmelotik: let's also log a warning there saying there is no receipt as that would be an unexpected event

subgraphs/package-lock.json
itsjerryokolo: Alright

subgraphs/gmx/src/mappings/glpManager.ts
itsjerryokolo: Okay

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.ts
itsjerryokolo: Yeah, that function is only used on L1. All the configs share the same interface. 
Removing that throws an error that we need the function is missing IIRC

deployment/context/versions.template.mustache
itsjerryokolo: Not sure how that came about. I'll take a look.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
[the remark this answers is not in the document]
itsjerryokolo: Okay. Looks like they just added Arbitrum Nova support. ETH and MAGIC are the tokens supported for now.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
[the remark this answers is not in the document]
itsjerryokolo: I'll change that the logic.

subgraphs/hop-protocol/protocols/hop-protocol/src/mainnet/bridge.ts
itsjerryokolo: Okay

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.ts
dmelotik: Can you make a comment in the code as to why we need it then?

itsjerryokolo: Okay, I'll add those.

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.json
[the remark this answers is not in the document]
itsjerryokolo: No, they don't

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.ts
itsjerryokolo: @dmelotik I opted to just return the tokenAddress in the function even though it will never be called. 
Does this solution work? If not I can revert. 
Leaving a comment means I'd have repeat the same comments a lot.

subgraphs/gmx/src/mappings/glpManager.ts
dmelotik: hmm, I still see these files. Since the `gmx` folder is no longer in master I think you can just delete that folder

subgraphs/package-lock.json
dmelotik: Am I missing something, it looks like the file changes are still there..?

subgraphs/hop-protocol/protocols/hop-protocol/config/deployments/hop-protocol-arbitrum/configurations.ts
dmelotik: I am not sure if typescript allows optional functions on an interface, but another option is to extend the interface.

If it works as is though, we can leave it.

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  getAmountValueUSD(token: Token, amount: BigInt): BigDecimal {
+    const _amount = bigIntToBigDecimal(amount, token.decimals);
+    return getUsdPrice(Address.fromBytes(token.id), _amount);
+  }
+}
+
+const conf = new BridgeConfig(
+  "0x03d7f750777ec48d39d080b020d83eb2cb4e3547",
```
dmelotik, line 50: Can we make this a constant in a `constants.ts` file?

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
```diff
@@ -0,0 +1,356 @@
...
+  "HOP-"
+    .concat(dataSource.network().toUpperCase().replace("-", "_"))
+    .concat("-BRIDGE"),
+  "hop-".concat(dataSource.network().replace("-", "_")).concat("-bridge"),
```
dmelotik, line 54: Maybe you can just make a function to return the BridgeConfig, rewriting all of these transformations each time is pretty tedious and makes it harder to read

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+}
+
+export function handleTokenSwap(event: TokenSwap): void {
+  if (NetworkConfigs.getPoolsList().includes(event.address.toHexString())) {
+    const amount = event.params.tokensSold;
+
+    const fees = amount
+      .times(BigInt.fromString("4"))
+      .div(BigInt.fromString("10000"));
```
dmelotik, line 75: This is another thing that could live as a constant

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
```diff
@@ -0,0 +1,356 @@
...
+    log.warning("FEES 2- fees: {}, fees: {}, amount: {}", [
+      fees.toBigDecimal().toString(),
+      fees.toString(),
+      amount.toString(),
+    ]);
+
+    const inputTokenOne = NetworkConfigs.getTokenAddressFromPoolAddress(
```
dmelotik, line 82: We should add error checks here to ensure the array has a length of 2, this could cause unwanted runtime errors. I would also `log.error` here.

Same goes for `poolConfig` below

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
```diff
@@ -0,0 +1,356 @@
...
+  );
+
+  if (!pool.isInitialized) {
+    pool.initialize(poolName, poolSymbol, BridgePoolType.LIQUIDITY, token);
+  }
+  if (!hPool.isInitialized) {
+    hPool.initialize(hPoolName, poolSymbol, BridgePoolType.LIQUIDITY, hToken);
+  }
+
+  if (
```
dmelotik, line 186: Can you add a comment as to why this transaction needs special care?

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/bridge.ts
```diff
@@ -0,0 +1,268 @@
...
+    .concat("-BRIDGE"),
+  "hop-".concat(dataSource.network().replace("-", "_")).concat("-bridge"),
+  BridgePermissionType.PERMISSIONLESS,
+  Versions
+);
+
+export function handleTransferFromL1Completed(
+  event: TransferFromL1Completed
+): void {
+  log.warning("TransferFromL1 - bridgeAddress: {},  hash: {}", [
```
dmelotik, line 72: Most of the warnings should probably just be info if you still want to log them: https://github.com/messari/subgraphs/blob/master/docs/WALKTHROUGH.md#using-logs-in-subgraphs

subgraphs/gmx/src/mappings/glpManager.ts
itsjerryokolo: Okay. I rebased hoping it would be deleted. 
I'll check

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/bridge.ts
itsjerryokolo: Okay, I'll change them

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
itsjerryokolo: Okay

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
itsjerryokolo: Okay, I'm importing a config now, which fixes this.

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
itsjerryokolo: Okay, I'm importing a config now, which fixes this.

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
itsjerryokolo: Okay

subgraphs/package-lock.json
itsjerryokolo: I added it now.

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/amm.ts
itsjerryokolo: Done

subgraphs/gmx/src/mappings/glpManager.ts
itsjerryokolo: I removed the gmx folder now

subgraphs/package-lock.json
dmelotik: You might just have to copy/paste the file from the master branch. It doesn't seem like the rebase is working here

itsjerryokolo: @dmelotik Manually added it now

subgraphs/hop-protocol/protocols/hop-protocol/src/L2/reward.ts
```diff
@@ -0,0 +1,185 @@
...
+  log.info("Staked --> emitter: {}, poolAddress: {}, amount: {}", [
+    event.address.toHexString(),
+    poolAddress,
+    amount.toString(),
+  ]);
+
+  const poolConfig = NetworkConfigs.getPoolDetails(poolAddress);
+  log.info("Staked 1 --> poolAddress: {},", [poolAddress]);
+
+  const poolName = poolConfig[1];
```
dmelotik, line 114: error checking for array size in this file too

itsjerryokolo: I’ll fix that.

dmelotik (approved):
Nice work @itsjerryokolo !

itsjerryokolo:
Thanks @dmelotik

Issue #2138, opened 2023-05-18
#bug; Perp futures sdk no snapshot issue


dmelotik:
**Describe the bug**
No snapshots being taken

**Why is it happening?**
https://github.com/messari/subgraphs/blob/c9b5ad3de69a7934f35b38fcc12d9f40a4f46a59/subgraphs/_reference_/src/sdk/protocols/perpfutures/protocol.ts#L132C1-L136
- protocol _lastUpdateTimestamp is updated to the current event timestamp even before taking the snapshots
- So, when the takeSnapshot function is called in ProtocolSnapshot constructor the last snapshot time and current event date matches hence it does not take any snapshots

![](https://cdn.discordapp.com/attachments/1097516279260258364/1107642727425916928/image.png)

PR #2147, opened 2023-05-19
fix(#patch); perpfutures; update sdk

dmelotik:
#2138


dhruv-chauhan (approved):
nit: could you also update SDK version in Readme title?

PR #2148, opened 2023-05-20
chore(); schema; update schemas with directives

dhruv-chauhan:


dmelotik:
@dhruv-chauhan could you focus on perpfutures in this PR? No need to scrap any work, but do those first and we will merge that in. Thanks! Also, no need to update the versions

cc @lowjiansheng

schema-bridge.graphql
```diff
@@ -982,7 +982,7 @@ type PoolHourlySnapshot @entity(immutable: true) {
   rewardTokenEmissionsUSD: [BigDecimal!]
 }
 
-type PoolRouteSnapshot @entity(immutable: true) {
+type PoolRouteSnapshot @entity(immutable: true) @regularPolling {
```
dmelotik, line 985: @lowjiansheng what do you think here? This is both a daily and hourly snapshot...

lowjiansheng: I think having it as an hourlySnapshot would be good. We'll just poll the data every hour instead of every day.

dmelotik (approved):
lgtm, let's merge, this. Then finish the rest in a different PR. Thanks @dhruv-chauhan

PR #2155, opened 2023-05-24
fix(#minor); lido-ethereum; totalRevenue for lido v2

dhruv-chauhan:
Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fdev


subgraphs/lido/protocols/lido/config/templates/lido.template.yaml
```diff
@@ -1,6 +1,11 @@
 specVersion: 0.0.4
 schema:
   file: ./schema.graphql
+features:
+  - grafting
+graft:
+  base: Qme9Vb2FTRRKaoR25bDbtL6JzTcmuQfTgbfmoQXigUynNd
+  block: 17172465
```
dmelotik, line 8: Can you either remove this from the template or add it to the config so we can turn it on/off?

PR #2156, opened 2023-05-24
Add directives to Generic Protocols

lowjiansheng:
# Description
- Add additional GraphQL directives to various Generic Protocols subgraph schemas. These directives will be helpful to Messari's internal data ingestion. It also helps in identifying the type and nature of the affected entities.

Affected subgraphs: Lido, RocketPool, TornadoCash

### DIFF
--- a/subgraphs/lido/schema.graphql
+++ b/subgraphs/lido/schema.graphql
@@ -35,7 +35,7 @@ enum ProtocolType {
   # Will add more
 }
 
-type Token @entity {
+type Token @entity @regularPolling {
   " Smart contract address of the token "
   id: ID!
 
@@ -63,7 +63,7 @@ enum RewardTokenType {
   BORROW
 }
 
-type RewardToken @entity {
+type RewardToken @entity @regularPolling {
   " { Reward token type }-{ Smart contract address of the reward token } "
   id: ID!
 
@@ -78,7 +78,7 @@ type RewardToken @entity {
 ##### Protocol Metadata #####
 #############################
 
-type Protocol @entity {
+type Protocol @entity @regularPolling {
   " Smart contract address of the protocol's main contract (Factory, Registry, etc) "
   id: ID!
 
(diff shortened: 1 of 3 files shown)


subgraphs/lido/schema.graphql
```diff
@@ -439,13 +439,13 @@ type PoolHourlySnapshot @entity {
...
 # Helps to accumulate total unique users
-type Account @entity {
+type Account @entity @regularPolling {
   " Address of the account "
   id: ID!
 }
 
 # Helper entity for calculating daily/hourly active users
-type ActiveAccount @entity {
+type ActiveAccount @entity @regularPolling {
```
dmelotik, line 448: Do we want this? Essentially shows unique users on a day/hour

lowjiansheng: Yeah i think this will be useful to have, but it does sound like there'll be a ton of data we're polling regularly in that case.

dmelotik: The count of `ActiveAccount` is probably 60-80% of all transactions ? so quite a bit

dmelotik:
Maybe this you want `@transaction`? https://github.com/messari/subgraphs/blob/bd9f56fb936c5b9808a3e202dadea4f29549fe83/subgraphs/rocket-pool/schema.graphql#L518

Could be good to see if you are trying to get deep on some transactions

lowjiansheng:
> Maybe this you want `@transaction`?
> 
> https://github.com/messari/subgraphs/blob/bd9f56fb936c5b9808a3e202dadea4f29549fe83/subgraphs/rocket-pool/schema.graphql#L518
> 
> Could be good to see if you are trying to get deep on some transactions

Yeah, but I realized that this entity does not use the standard `timestamp` field which the poller uses for time filtering. Therefore, we will not be able to poll it anyway even if I put this directive.

PR #2158, opened 2023-05-24
fix(#minor); Uniswap V3; Fix wBTC price.

steegecs:


dmelotik (approved):
lgtm, i will redeploy all these on my hosted to sync

PR #2172, opened 2023-05-30
fix(#schema); Uniswap V3 Forks; Make DEX schema compatible with old schema, and add corresponding changes to Uniswap Forks

steegecs:
**Context:** 
The DEX schemas from version 1.3.~ and the new extended schema were not compatible in the naming of certain volume fields on the LiquidityPool and LiquidityPoolSnapshot entities. This PR updates these problematic fields to be consistent between the schemas and applies the necessary changes to the Uniswap Forks code.


dmelotik (approved):
Approving, with the caveat of updating a major schema version

PR #2173, opened 2023-05-30
docs(); Bytes as IDs perf; Performance test for bytes as IDs updated in response to issue being resolved by Edge and Node

steegecs:
**Context:**
We ran a performance test using bytes as IDs in April 2022. The results from the test indicated that using bytes as IDs and immutable entities did not have the expected benefits on indexing speed as was expected. This issue has been addressed by Lutter as Edge and Node, and the performance is now improved with Bytes and immutable IDs. The issue on the graph-node can be viewed here: https://github.com/graphprotocol/graph-node/issues/3534.


dmelotik (approved):
lgtm

Issue #2024, opened 2023-04-17
#feature; Add interest tracking at pos level (lending)


dmelotik:
**What do you think is missing?**
Add new fields to lending schema to allow for a generic model to give data consumers the tools to calculate the current value of positions with interest.

**Describe the solution you'd like**
- Update schema / SDK
- Update existing subgraph to follow the schema

**Additional context**
Generally lending protocols have an index that tracks interest index.

Similar issue, see: #1689

dmelotik:
New fields to add:

- `Market.liquidityIndex`
- `Market.borrowIndex`
- `Position.principal`
- `PositionSnapshot.principal`
- `PositionSnapshot.interestIndex`

See Compound v3 implementation for details on what each one means. 

Action Items:
- [ ] Apply this update to the standard lending schema / concurrently upgrade the lending SDK
- [ ] Add this change to aave forks (after #1993 )

Issue #1689, opened 2023-02-03
#feature; Support pos rate tracking in lending


dmelotik:
**What do you think is missing?**
For lending subgraphs he is tracking return more accurately on a user level. For protocols with non-rebasing collateral tokens he is keeping track of entry conversion rate and final and average. See the photo for the data. We could do a similar thing for rebasing tokens and it would cover most cases.

This will be helpful for account and position metrics on a data app. There is potential for a more intricate DA that looks into positions across protocols.

![image (1)](https://user-images.githubusercontent.com/60583747/216718554-8399550f-6de7-4370-86f7-4d63f85e1a31.png)

bye43:
Needs schema version change

PR #2176, opened 2023-05-31
feat(#minor); lending schema upgrade; support pos level interest accumulated

dmelotik:
Support interest generated per position and update sdk to reflect this.

#1689
#2024


subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
```diff
@@ -227,6 +228,11 @@ export class DataManager {
     return this.market;
   }
 
+  updateLastMarketTimestamp(): void {
```
dhruv-chauhan, line 231: alt: rename function to `saveMarket()` and replace everywhere there is a `this.market.save()`

subgraphs/_reference_/src/sdk/protocols/lending/position.ts
```diff
@@ -303,6 +309,14 @@ export class PositionManager {
       .times(priceUSD);
     snapshot.blockNumber = event.block.number;
     snapshot.timestamp = event.block.timestamp;
+
+    snapshot.principal = this.position!.principal;
```
dhruv-chauhan, line 313: `if (this.position!.principal) snapshot.principal = this.position!.principal;`

subgraphs/_reference_/src/sdk/protocols/lending/position.ts
```diff
@@ -303,6 +309,14 @@ export class PositionManager {
...
+    if (this.position!.side == PositionSide.BORROWER) {
```
dhruv-chauhan, line 314: `if (this.market.borrowIndex && this.position!.side == PositionSide.BORROWER) { ... }`
same for supply index

subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
dmelotik: Let me fix that. It is only intended to be updated with supply/borrowIndex since that timestamp is needed to calculate delta between last snapshot and current index

dhruv-chauhan (approved):
looks good!

PR #2177, opened 2023-05-31
chore(); schema; update subgraph schemas with directives

dhruv-chauhan:


dmelotik (approved):
lgtm! Thanks @dhruv-chauhan

Issue #2163, opened 2023-05-25
BadgerDAO Ethereum: pricing issue in deposit volume


bye43:
- Looks like there is a pricing issue w/ a deposit on 04/27/23 for `0x7e7e112a68d8d2e221e11047a72ffc1065c38e1a`
- [Query](https://api.thegraph.com/subgraphs/name/messari/badgerdao-ethereum/graphql?query=%7B%0A++vaults+%28%0A++++where%3A+%7B%0A++++++id%3A%220x7e7e112a68d8d2e221e11047a72ffc1065c38e1a%22%0A++++%7D%0A++%29%7B%0A++++id%0A++++name%0A++++deposits+%28%0A++++++where%3A%7B%0A++++++++timestamp_gte%3A1682557200%0A++++++++timestamp_lt%3A+1682643600%0A++++++%7D%0A++++%29%7B%0A++++++id%0A++++++hash%0A++++++timestamp%0A++++++amountUSD%0A++++++amount%0A++++%7D%0A++%7D%0A%7D)

bye43:
@harsh9200 Pricing still looks off here. [Etherscan](https://etherscan.io/tx/0xfc874d3d231eddb30eb30310eae5270ca028f365fdfc61715d8a895123e61547) is saying that the deposit value was around 5k and the [subgraph is saying that it was about $200](https://api.thegraph.com/subgraphs/name/messari/badgerdao-ethereum/graphql?query=%7B%0A++vaults+%28%0A++++where%3A+%7B%0A++++++id%3A%220x7e7e112a68d8d2e221e11047a72ffc1065c38e1a%22%0A++++%7D%0A++%29%7B%0A++++id%0A++++name%0A++++inputToken+%7B%0A++++++id%0A++++++lastPriceUSD%0A++++%7D%0A++++deposits+%28%0A++++++where%3A%7B%0A++++++++timestamp_gte%3A1682557200%0A++++++++timestamp_lt%3A+1682643600%0A++++++%7D%0A++++%29%7B%0A++++++id%0A++++++hash%0A++++++timestamp%0A++++++amountUSD%0A++++++amount%0A++++%7D%0A++%7D%0A%7D). Looks to be still an issue with the `DIGG` pricing since the `lastPriceUSD` is reporting $125 and [CMC is saying it is ~2.5K](https://coinmarketcap.com/currencies/digg/)

bye43:
[...]

PR #2181, opened 2023-06-01
feat(#minor); BadgerDAO Subgraph; Fix deposit amount bug

harsh9200:
Fixes: #2163 
Deployment: https://thegraph.com/hosted-service/subgraph/messari/badgerdao-ethereum?version=pending


dmelotik:
Can you increment versions (minor version change)

subgraphs/badgerdao/src/modules/Deposit.ts
```diff
@@ -92,7 +92,13 @@ export function Deposit(
       constants.BIGINT_ZERO
     );
 
-    depositAmount = totalSupply.minus(vault.outputTokenSupply!);
+    const sharesMinted = totalSupply.minus(vault.outputTokenSupply!);
+
+    depositAmount = vault.outputTokenSupply!.isZero()
```
dmelotik, line 97: Shouldn't this be done after we set the new vault.inputTokenBalance and vault.outpuTokenSupply?

harsh9200: No, this needs to be done with the old values as we need to find the amount deposited in the pool using the shares minted during a deposit transaction.

subgraphs/balancer-forks/src/prices/oracles/InchOracle.ts
dmelotik, line 1: Out of scope of this PR?

dmelotik (approved):
Approving and can be merged once that last comment is fixed! Nice work @harsh9200

subgraphs/balancer-forks/src/prices/oracles/InchOracle.ts
harsh9200: No, this is a required change for the fix, cause this bug was related to the digg token pricing and only the inchOracle method returns the correct price for that token.

PR #2197, opened 2023-06-09
Update uniswap-v3-fork schema directive

lowjiansheng:
# Description
- The Position Snapshot should be tagged as a @snapshot instead of @regularPolling as the entities are immutable.


dmelotik (approved):
godspeed

Issue #2062, opened 2023-04-25
Build Camelot V2 Subgraph


bye43:
- On [1.3.1 schema](https://github.com/messari/subgraphs/blob/master/schema-dex-amm.graphql)
- See other [uniswap-v2 forks](https://github.com/messari/subgraphs/tree/master/subgraphs/uniswap-forks)
```[tasklist]
### Deployment QA Status
- [ ] Arbitrum
```

bye43:
@spitko

PR #2198, opened 2023-06-09
feat(#major); camelot-v2; Add Camelot V2 subgraph

spitko:
#2062 
https://okgraph.xyz/?q=spitko%2Fcamelot-v2-arbitrum


dhruv-chauhan (commented):
You should try to use files at https://github.com/messari/subgraphs/tree/master/subgraphs/uniswap-forks/src, unless protocol-specific logic to minimize code repetition.
I reckon most of code in `common/` can be reused

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
```diff
@@ -0,0 +1,364 @@
...
+
+  if (!token.lastPriceUSD || token.lastPriceUSD!.equals(BIGDECIMAL_ZERO)) {
+    token.save();
+    return priceSoFar;
+  }
+
+  // If priceSoFar 10x greater or less than token.lastPriceUSD, use token.lastPriceUSD
+  // Increment buffer so that it allows large price jumps if seen repeatedly
+  if (
+    priceSoFar.gt(token.lastPriceUSD!.times(BIGDECIMAL_TWO)) ||
```
dhruv-chauhan, line 183: nit: update comment to: `If priceSoFar 2x greater or lesser...`

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
```diff
@@ -0,0 +1,364 @@
...
+      token.save();
+      return token.lastPriceUSD!;
+    }
+  }
+
+  token._largePriceChangeBuffer = 0;
+  token._largeTVLImpactBuffer = 0;
+
+  token.save();
+  return priceSoFar;
```
dhruv-chauhan, line 197: should save `token.lastPriceUSD = priceSoFar` before returning

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
```diff
@@ -0,0 +1,364 @@
...
+  price: BigDecimal
+): bool {
+  const newTokenTotalValueLocked = convertTokenToDecimal(
+    token._totalSupply,
+    token.decimals
+  ).times(price);
+
+  // If price is too high, skip this pool
+  if (newTokenTotalValueLocked.gt(BIGDECIMAL_TEN_BILLION)) {
+    log.warning("Price too high for token: {} from pool: {}", [
```
dhruv-chauhan, line 223: probably an error log

dmelotik: https://github.com/messari/subgraphs/blob/master/docs/WALKTHROUGH.md#using-logs-in-subgraphs

spitko: warning seems correct to me in this case

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
spitko: Added, but this is copied from the common src.

This also works as is, since findUSDPricePerToken only has one usage and that saves token.lastPriceUSD.

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
spitko: done

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
```diff
@@ -0,0 +1,364 @@
...
+        stableAmount = pool.inputTokenBalances[1];
+      }
+    } else {
+      if (pool.inputTokenBalances[0] > stableAmount) {
+        nativeAmount = pool.inputTokenBalances[1];
+        stableAmount = pool.inputTokenBalances[0];
+      }
+    }
+  }
+  nativeToken.lastPriceUSD = safeDiv(stableAmount, nativeAmount);
```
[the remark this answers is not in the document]
spitko: The amounts are already BigDecimals, `_LiquidityPoolAmount` is a custom entity in the uniswap-v2 forks schema.

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
```diff
@@ -0,0 +1,364 @@
...
+          let newPrice = BIGDECIMAL_ZERO;
+          if (stable) {
+            newPrice = calculateRateForStablePair(
+              pool,
+              pair,
+              token,
+              whitelistToken
+            ).times(whitelistToken.lastPriceUSD!);
+          } else {
+            newPrice = safeDiv(
```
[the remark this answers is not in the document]
spitko: same here, tokenBalances are already BigDecimals

spitko:
> You should try to use files at https://github.com/messari/subgraphs/tree/master/subgraphs/uniswap-forks/src, unless protocol-specific logic to minimize code repetition. I reckon most of code in `common/` can be reused

I tried using the common code at the start, but the issue is that it is not easy to extend it. For example, in order to support Camelot's [directional fees](https://docs.camelot.exchange/protocol/amm/amm-v2/dynamic-directional-fees) for pools, I had to modify `creators.ts:createPoolFees`, which is used in `getters.ts:getLiquidityPool`, which has over 10 usages, that would all have to be changed to use my implementation. 
In the end it seemed better from a readability standpoint to just copy everything to `camelot-v2/src`.

subgraphs/uniswap-forks/protocols/camelot-v2/src/price.ts
dmelotik: The shared uniswap fork code uses warning, so I agree to stick w the precedent here

dmelotik:
> I tried using the common code at the start, but the issue is that it is not easy to extend it. For example, in order to support Camelot's [directional fees](https://docs.camelot.exchange/protocol/amm/amm-v2/dynamic-directional-fees) for pools, I had to modify `creators.ts:createPoolFees`, which is used in `getters.ts:getLiquidityPool`, which has over 10 usages, that would all have to be changed to use my implementation. In the end it seemed better from a readability standpoint to just copy everything to `camelot-v2/src`.

@steegecs do you have any thoughts on if there is a better way to organize this code? Or is this the best approach when some functions need to be changed?

steegecs:
> Thanks for review @dhruv-chauhan, fixed the issues mentioned :)
> 
> > You should try to use files at https://github.com/messari/subgraphs/tree/master/subgraphs/uniswap-forks/src, unless protocol-specific logic to minimize code repetition. I reckon most of code in `common/` can be reused
> 
> I tried using the common code at the start, but the issue is that it is not easy to extend it. For example, in order to support Camelot's [directional fees](https://docs.camelot.exchange/protocol/amm/amm-v2/dynamic-directional-fees) for pools, I had to modify `creators.ts:createPoolFees`, which is used in `getters.ts:getLiquidityPool`, which has over 10 usages, that would all have to be changed to use my implementation. In the end it seemed better from a readability standpoint to just copy everything to `camelot-v2/src`.

Do you even use the 

> Thanks for review @dhruv-chauhan, fixed the issues mentioned :)
> 
> > You should try to use files at https://github.com/messari/subgraphs/tree/master/subgraphs/uniswap-forks/src, unless protocol-specific logic to minimize code repetition. I reckon most of code in `common/` can be reused
> 
> I tried using the common code at the start, but the issue is that it is not easy to extend it. For example, in order to support Camelot's [directional fees](https://docs.camelot.exchange/protocol/amm/amm-v2/dynamic-directional-fees) for pools, I had to modify `creators.ts:createPoolFees`, which is used in `getters.ts:getLiquidityPool`, which has over 10 usages, that would all have to be changed to use my implementation. In the end it seemed better from a readability standpoint to just copy everything to `camelot-v2/src`.

Yeah, I would be happy to see them combined, because so much of the code is shared, but I see your point @spitko. I think we could have all the code shared between them by using an `isDirectional` flag or something along these lines, but I think a good intermediate approach would be to only modify event handlers that are impacted by this difference. 

If I understand correctly, this is mostly justa difference for the pool creation event, and the swap event plus the additional fee update event. You could just only create handlers for these events in the camelot v2 protocol folder, and have most of the code used by these handlers imported from the common src code. The biggest difficulty is the `getLiquidityPool` function as @spitko pointed out because of the `createFees` function. We could add a flag to this function, though, that only calls the createPoolFees if some condition is met like.

```
export function getLiquidityPool(
  poolAddress: string,
  blockNumber: BigInt
): LiquidityPool {
  const pool = LiquidityPool.load(poolAddress)!;
  if (NetworkConfigs.getProtocolSlug() != "camelot-v2") {
    pool.fees = createPoolFees(poolAddress, blockNumber);
    pool.save();
  }
  return pool;
}
```

Or we could move the `createPoolFees` function out into the `updateVolumeAndFees()` function where it will always get updated in time for the swap fees. Then since this function would be different for Camelot V2 anyways because of the directional fees. This way, most all of the code can be shared except a couple of handlers (handlePairCreated, handleSwap, + handleFeePercentUpdated) and a few functions (createPoolFees, createLiquidityPool, and updateVolumeAndFees). 

I think I like this last approach. Let me know what you think @spitko.

spitko:
Thanks @steegecs for the tip about moving the `createPoolFees()` function call out of `getLiquidityPool()`. I updated the code to reuse the common code as much as possible; in the end there was just one function that had to be changed under [`price.ts`](https://github.com/messari/subgraphs/pull/2198/commits/68aaa37c9ce19804068d11dc99fd001c1d413e53#diff-09a2fae4a7d7b7b24a93411976756bfe9ea41db40acbba640c8952f4c5e77883R115-R126). There was one Camelot specific `Pair` contract function call, which would fail to compile for other uniswap fork subgraphs, because that function did not exist in the other ABIs. But I worked around this by changing it to a raw contract function call instead.

But since I also now applied the other pricing fix to the common src, should I now bump the other uniswap-forks subgraphs' versions?

steegecs:
> Thanks @steegecs for the tip about moving the `createPoolFees()` function call out of `getLiquidityPool()`. I updated the code to reuse the common code as much as possible; in the end there was just one function that had to be changed under [`price.ts`](https://github.com/messari/subgraphs/pull/2198/commits/68aaa37c9ce19804068d11dc99fd001c1d413e53#diff-09a2fae4a7d7b7b24a93411976756bfe9ea41db40acbba640c8952f4c5e77883R115-R126). There was one Camelot specific `Pair` contract function call, which would fail to compile for other uniswap fork subgraphs, because that function did not exist in the other ABIs. But I worked around this by changing it to a raw contract function call instead.
> 
> Deployed the latest version under a separate subgraph here: https://okgraph.xyz/?q=spitko%2Fcamelot-v2-arbitrum-new
> 
> But since I also now applied the other pricing fix to the common src, should I now bump the other uniswap-forks subgraphs' versions?

Looking at this now. But yeah, we should then bump the version of the other Uniswap Forks.

subgraphs/uniswap-forks/protocols/camelot-v2/README.md
```diff
@@ -0,0 +1,88 @@
+# Camelot V2 Subgraph
+
+- The code that faciliates the basic operations of this exchange are forked from Uniswap V2. The two main contracts that are forked and used to index data from swaps, deposits, withdraws, and more are the `UniswapV2Factory` and `Pair` contracts.
```
steegecs, line 3: Will you include in the README how the camelot subgraph differs from the rest in the Uniswap Forks library?

subgraphs/uniswap-forks/protocols/camelot-v2/src/mappings/pool.ts
```diff
@@ -0,0 +1,90 @@
...
+import { createPoolFees } from "../../src/common/creators";
+import {} from "../../src/common/updateMetrics";
+import { INT_THREE, PairType } from "../common/constants";
+import { createSwapHandleVolumeAndFees } from "../common/creators";
+import { _HelperStore } from "../../../../generated/schema";
+
+// Handle transfers event.
+// The transfers are either occur as a part of the Mint or Burn event process.
+// The tokens being transferred in these events are the LP tokens from the liquidity pool that emitted this event.
+export function handleTransfer(event: Transfer): void {
```
steegecs, line 33: This is a nit pick, and you may like how you have it now better, so I will leave it up to you, but you could also just point the subgraph.yaml template for camelot to the handlers in the main /src library for `Transfer`, `Sync`, `Deposit`, and `Withdraw`, so you do not have to create an extra handler for each. I think it would also help clarify what is different here, too.

subgraphs/uniswap-forks/src/price/price.ts
```diff
@@ -111,6 +112,18 @@ export function findUSDPricePerToken(
           largestLiquidityWhitelistTokens = whitelistTokenLocked;
           // token1 per our token * nativeToken per token1
 
+          if (NetworkConfigs.getProtocolSlug() == "camelot-v2") {
+            const newPrice = findStablePairUSDPriceForToken(
```
steegecs, line 116: So if the `newPrice` does not exist, it needs to retrieve the price the normal way? Or is this always supposed to get the price.

steegecs (changes requested):

subgraphs/uniswap-forks/protocols/camelot-v2/src/mappings/pool.ts
spitko: I agree it would be better that way and that's what I tried to do first, but there was some issue with splitting the event handlers to different files, which is why I ended up doing it this way instead.

subgraphs/uniswap-forks/protocols/camelot-v2/README.md
spitko: added, thanks 👍

subgraphs/uniswap-forks/src/price/price.ts
spitko: Yeah, the `findStablePairUSDPriceForToken` function only returns a price if the pool is marked as a stable (Curve) pool. For other pools the normal uniswap price calculation is used.

steegecs (approved):
You have my approval!

subgraphs/uniswap-forks/protocols/ubeswap/src/common/helpers.ts
dhruv-chauhan, line 1: missed updating `getLiquidityPool` for honeyswap,

- https://github.com/spitko/subgraphs/blob/camelot/subgraphs/uniswap-forks/protocols/honeyswap/src/common/creators.ts#L142
- https://github.com/spitko/subgraphs/blob/camelot/subgraphs/uniswap-forks/protocols/honeyswap/src/common/creators.ts#L156

dhruv-chauhan (commented):
One small comment, otherwise good. :call_me_hand:

subgraphs/uniswap-forks/protocols/ubeswap/src/common/helpers.ts
spitko: thanks, fixed 👍

dhruv-chauhan (approved):
:rocket:

Issue #2200, opened 2023-06-11
Bug with closed Uniswap V3 polygon positions, incorrect NULL result for hashClosed, blockNumberClosed, timestampClosed


Gwendiie:
**Bug:**
When I ran this in the Playground (https://thegraph.com/hosted-service/subgraph/messari/uniswap-v3-polygon), **it gave me null for all hashClosed, blockNumberClosed, timestampClosed, even though the following 3 positions have already closed.** 
To add on further, because i filtered for **liquidity_gt: 0** , which allows me to fetch positions that have actual liquidity provided or borrowed, excluding positions with zero liquidity. By right, these 3 positions below in my response should not appear since these positions have been closed already. 

**What I ran in the playground:** 
{
  positions(where: {
    account: "0x027afe0a7caa646c2c035e9108079e1a9ad50208",
    liquidity_gt: 0
  }) {
    pool {
      name
      symbol
      liquidityTokenType
      inputTokens{
      name
      symbol 
      id}
      tick
      inputTokenBalances
      inputTokenBalancesUSD
      inputTokenWeights
    }
    id
    hashOpened
    hashClosed
    blockNumberClosed
    timestampClosed
    }
    }
    
    
    Response that I got:
    
     {
        "pool": {
          "name": "Uniswap V3 (PoS) Wrapped BTC/XSGD 0.3%",
          "symbol": "(PoS) Wrapped BTC/XSGD",
          "liquidityTokenType": "MULTIPLE",
          "inputTokens": [
            {
              "name": "(PoS) Wrapped BTC",
              "symbol": "WBTC",
              "id": "0x1bfd67037b42cf73acf2047067bd4f2c47d9bfd6"
            },
            {
              "name": "XSGD",
              "symbol": "XSGD",
              "id": "0xdc3326e71d45186f113a2f448984ca0e8d201995"
            }
          ],
          "tick": "58490",
          "inputTokenBalances": [
            "172113123",
            "724820164878"
          ],
          "inputTokenBalancesUSD": [
            "44243.3819161305976196537198756513",
            "538641.3957853791411718642607258928"
          ],
          "inputTokenWeights": [
            "50",
            "50"
          ]
        },
        "id": "0x0000000000970b0900",
        "hashOpened": "0x7514384655d9ff909354d5d6c13a6126087b118e6681216b2d8ac9268019105d",
        **"hashClosed": null,
        "blockNumberClosed": null,
        "timestampClosed": null**
      },
      {
        "pool": {
[...]

PR #2204, opened 2023-06-12
fix(#minor); Uniswap V3 Forks; Fix liquidity adjustment on DecreaseLiquidityEvent for Positions

steegecs:
**Context:**
Liquidity was being increased on the DecreaseLiquidity event. This was causing positions not to close as well as an improper balance on the liquidity value of the position. This issue was brought up in #2200.


dmelotik (approved):
lgtm

PR #2215, opened 2023-06-15
fix(#minor); velodrome-finance-optimism; Fix unexpected null indexing error

dhruv-chauhan:
Breakpoint: https://github.com/messari/subgraphs/blob/master/subgraphs/velodrome-finance/src/mappings/helpers/rewards.ts#L28

Deployments: 

V1
- graft: https://okgraph.xyz/?q=Qma8BsN2BWCanPYVfM33EGnZm9b45a2CKYhRkQKukSEpqb
- complete: https://okgraph.xyz/?q=QmZqmMiG2DCWZpZQHornogFLYmzYTPL6pt6PevykYcPLod

V2
- complete: https://okgraph.xyz/?q=QmZxVNNp5c9ahwujVcbrrokqsYSYurVDaPHhGntk3xWaTn


dmelotik:
@dhruv-chauhan thanks for this. Why don't you graft the fix so it can sync faster?

dmelotik:
Also, this is the "pool" that was missing. It was never created through the PairCreated event and it is used to drain funds into velodrome v2. https://optimistic.etherscan.io/address/0xda03dc399af3f1545748243aafbc5050a63b17ec#code

We should consider a fix where we hardcode this in instead? to skip if that pool is transacted on...

dhruv-chauhan:
Will request for review, QA once velodrome-v1 graft has completed syncing

dmelotik (commented):
Good call on splitting between V1/V2 @dhruv-chauhan !

A few notes (otherwise looks good) in order to properly make this migration:
- Let's make notes in the readme of what we did here and link their medium article: https://medium.com/@VelodromeFi/velodrome-v2-a-new-era-1bd84509fa23
- The protocol name and slug should also reflect V1/V2 distinction

dmelotik (approved):
Approving the code looks good! Will merge once the QA passes

Issue #2216, opened 2023-06-15
#bug; Compound V3 Revenue Bug


dmelotik:
**Describe the bug**
There are uncharacteristic spikes in revenue in compound V3

**Reproduction**
Where can we see this bug in action?
<img width="985" alt="Screen Shot 2023-06-15 at 4 37 46 PM" src="https://github.com/messari/subgraphs/assets/60583747/a6140724-e2dc-4750-987d-ee083fbb1032">

1. https://subgraphs.messari.io/subgraph?endpoint=messari/compound-v3-ethereum&tab=protocol

**Additional context**
The formula for revenue in compound v3 follows: `totalRev = (lastBaseBorrowIndex - newBaseBorrowIndex) / BASE_INDEX_SCALE * totalBorrowAmount`
See: https://github.com/messari/subgraphs/blob/master/subgraphs/compound-v3/protocols/compound-v3/src/mapping.ts#L1137-L1141

[...]

PR #2217, opened 2023-06-16
fix(#patch); comp v3; rev spike bug

dmelotik:
#2216

https://okgraph.xyz/?q=dmelotik%2Fcompound-v3-polygon
https://okgraph.xyz/?q=dmelotik%2Fcompound-v3-ethereum
https://okgraph.xyz/?q=dmelotik%2Fcompound-v3-arbitrum

Notes:
- Updated Compound V3 to latest schema `3.1.0`
- Added `isNewMarket` to lending SDK and fixed compile errors
- Updated to newest lending SDK in comp v3
- Fixed issue -> needed to stop overwriting the `borrowIndex` each time the proxy contract for the comet market is deployed...

### DIFF
--- a/subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
+++ b/subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
@@ -52,7 +52,7 @@ import { PositionManager } from "./position";
  * need to think about all of the detailed storage changes that occur.
  *
  * Schema Version:  3.1.0
- * SDK Version:     1.0.3
+ * SDK Version:     1.0.4
  * Author(s):
  *  - @dmelotik
  */
@@ -88,6 +88,7 @@ export class DataManager {
   private inputToken!: TokenManager;
   private oracle!: Oracle;
   private snapshots!: SnapshotManager;
+  private newMarket: boolean = false;
 
   constructor(
     marketID: Bytes,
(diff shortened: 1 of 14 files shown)


dhruv-chauhan (commented):
Code looks good, though I couldn't see the change in the deployments you've listed.

dmelotik:
@dhruv-chauhan I think i found the actual fix, updated code

subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
```diff
@@ -227,6 +230,10 @@ export class DataManager {
     return this.market;
   }
 
+  isNewMarket(): boolean {
```
dhruv-chauhan, line 233: will this not always return true? `newMarket` is set to true on creation of a new market, then never updated. :thinking: 

should it be,
```
constructor(...) {
  let _market = Market.load(marketID);
  this.newMarket = false;
  if (!_market) {
    ...
    this.newMarket = true;
  }
}
```

dmelotik: That is what is happening. Defaults to false, but if a new market is made it is set to true

steegecs, line 233: Do you feel its better to have a method for this? I am considering its better to make the field `isNewMarket` or something and just access the field directly to reduce the amount of code needed since its just returning the value.

steegecs (approved):
Looks good!

subgraphs/_reference_/src/sdk/protocols/lending/manager.ts
dmelotik: AFAIK it is better OOP practice to use private variables and encapsulate when possible. And write functions to modify/access those private vars. https://stackoverflow.com/questions/14399929/should-i-use-public-or-private-variables

PR #2237, opened 2023-06-23
chore(); update morpo compound to `prod`

bye43:
Accidentally marked this as morhpo aave v2, but it is for morpho compound


dmelotik (approved):
morpo gtg

PR #2251, opened 2023-06-29
chore(); Aave forks; update versions

dmelotik:
Few things missing:
- Update versions properly
- Update uwu lend readme
- move aave v3 to graft false


tnkrxyz (approved):
thought I fixed the versions. thanks for fixing these.

PR #2259, opened 2023-07-03
fix(#patch); aave-v3; add transfer event for mint to treasury

dhruv-chauhan:
Deployment,

aave-v3-ethereum: https://okgraph.xyz/?q=dhruv-chauhan%2Faave-v3-ethereum

### DIFF
--- a/subgraphs/aave-forks/src/sdk/manager.ts
+++ b/subgraphs/aave-forks/src/sdk/manager.ts
@@ -38,6 +38,7 @@ import {
   ProtocolType,
   Transaction,
   TransactionType,
+  ZERO_ADDRESS,
 } from "./constants";
 import { SnapshotManager } from "./snapshots";
 import { TokenManager } from "./token";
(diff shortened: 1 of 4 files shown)


dmelotik:
Hey @dhruv-chauhan thanks for taking this issue!

A few comments:
- In aave v3 and forks (spark lend, radiant v2) there is an event we can actually watch `mintToTreasury()` https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/libraries/logic/PoolLogic.sol#L106
- And in aave v2 they don't have that event, but this is what happens when we see a mintToTreasury action: https://github.com/aave/protocol-v2/blob/master/contracts/protocol/tokenization/AToken.sol#L167
   - However, it looks like aave-v2 treasury address account actually has the proper positions, so this may only be necessary on aave-v3 and forks
- And to know that the fix is working you can just check if the treasury address has the proper open positions

subgraphs/aave-forks/src/sdk/manager.ts
```diff
@@ -728,6 +729,39 @@ export class DataManager {
     return liquidate;
   }
 
+  createTransferEvent(
```
dmelotik, line 732: Will it work to just use the current `createTransfer` function in the `DataManager`? 

I wasn't expecting to have to update the sdk

dmelotik:
New changes look good, just 1 simplification. Also:

- Here is the [query](https://api.thegraph.com/subgraphs/id/QmThq3i99QeStX3YaCiGn1stVYmpneMqpBYWrcDpYFUEYt/graphql?query=query+MyQuery+%7B%0A++account%28id%3A+%220x464c71f6c2f760dda6093dcb91c24c39e5d6e18c%22%29+%7B%0A++++id%0A++++depositCount%0A++++transferredCount%0A++++positions+%7B%0A++++++market+%7B%0A++++++++name%0A++++++%7D%0A++++++balance%0A++++%7D%0A++%7D%0A%7D) I was using (for ethereum)
- Can you add this to spark lend
- Update all affected versions and deploy

subgraphs/aave-forks/src/sdk/manager.ts
dhruv-chauhan: @dmelotik pushed the minimal changes that'd be needed in the sdk. Basically, skip creating account and position entities and not update usage if the sender is ZERO_ADDRESS (mint).

Now, shall I make these changes in the reference sdk, or consider this as a protocol specific diversion?

subgraphs/aave-forks/src/sdk/manager.ts
dmelotik, line 1: I am not fully confident we need to even make this SDK change. I think it is fine if there is a `0x00...000` account with or without balance. In theory there could be aTokens there, and why not have that data?

dhruv-chauhan: reverted. I was just a bit hesitant counting zero address as a user, but ig that's fine.

(uploading subgraph to IPFS is failing at the moment, will try deploying latest in a bit)

dmelotik:
@dhruv-chauhan the last thing is to update the subgraph versions in `deployment.json` other than that the code is good to go

PR #2265, opened 2023-07-06
fix(#patch); morpho; tvl add up issue

dmelotik:
We are seeing issues with the TVL in financial snapshots not adding up to what we expect from what the marekt TVLs are.

Test:
- https://okgraph.xyz/?q=dmelotik%2Fmorpho-aave-v2-ethereum
- https://okgraph.xyz/?q=dmelotik%2Fmorpho-compound-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -3068,7 +3068,7 @@
         "status": "prod",
         "versions": {
           "schema": "3.0.1",
-          "subgraph": "1.0.0",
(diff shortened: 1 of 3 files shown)


bye43:
@dmelotik looks like it is not working. Did not see the TVL spike for Morpho AAVE that we have in PM.

dmelotik:
Not working. see https://github.com/messari/subgraphs/issues/2267

Issue #2141, opened 2023-05-19
#bug; incorrect logIndex value in yearn-v2-ethereum


chenzongxiong:
**Describe the bug**
The `logIndex` in response is incorrect.

**Reproduction**
Click the following link to retrieve the response. See screenshot.
1. Link to subgraph on 
[`https://api.thegraph.com/subgraphs/name/messari/yearn-v2-ethereum/graphql`](https://api.thegraph.com/subgraphs/name/messari/yearn-v2-ethereum/graphql?query=%7B%0A++deposits%28first%3A+1%29+%7B%0A++++id%0A++++hash%0A++++logIndex%0A++++from%0A++++to%0A++++blockNumber%0A++++timestamp%0A++++amount%0A++%7D%0A%7D)

2. Check the data on etherscan, we can find there are **two** event logs in this transaction, 
   https://etherscan.io/tx/0x00013d2e415b4448f9008517308560b3574d585fc4bb521f34835f22a136b183#eventlog

**Expected behavior**
A clear and concise description of what you expected to see.
Consider the results in etherscan and subgraph, `logIndex` value in subgraph should be equal to **265** or **266**. However, subgraph return **127**

**Screenshots**
<img width="1100" alt="image" src="https://github.com/messari/subgraphs/assets/8354165/49c6dbfc-aaa1-4844-b2d2-5fcce3e2e4d4">
<img width="966" alt="image" src="https://github.com/messari/subgraphs/assets/8354165/851b5447-4af1-42d2-b03b-8cded5d8e675">

**Additional context**

bye43:
cc: @harsh9200

dmelotik:
Thanks @chenzongxiong, this fix is almost in. As a note, there are no `Deposit` or `Withdraw` events on yearn v2, they only emit token transfer events. Therefore we techincally can't assign a `deposit.logIndex`. Our fix is to just set that field to 0 and make a comment of the situation.

PR #2269, opened 2023-07-07
fix(#minor); yearn v2; fix log index

harsh9200:
**Context:** As yearn pools don't emit events for deposit/withdraw the log index should be zero.  

Fixes: #2141 
Deployment: [Yearn v2](https://thegraph.com/hosted-service/subgraph/messari/yearn-v2-ethereum?version=pending)


dmelotik (commented):
Few comments before we can merge:
- can you comment on the line explaining why it needs to be 0 (cuz call handler)
- fix the lint
- Increment the subgraph version

deployment/deployment.json
dmelotik, line 1: update other yearn deployment `subgraphVersions`

dmelotik (approved):
code lgtm! Nice

PR #2298, opened 2023-07-26
fix(#minor); aave-forks; fix aave indexing errors

dhruv-chauhan:
Deployments,

- aave-v2-ethereum: https://okgraph.xyz/?q=QmQVb2SVgVRVyncMMCtg5XUuHcT2qXS7Xs4EpAFTh9XkaR
- aave-v3-arbitrum: https://okgraph.xyz/?q=QmQKQa6yWE4sUGqfhnHkvxm9eWmFhpAxch1fhqg86C1Ja6
- aave-v3-polygon: https://okgraph.xyz/?q=QmYT5Dgm9qdHBVExzbWHGZiAzvR8HcSHk5CyZiVrX8c7Uf
- spark-lend-ethereum: https://okgraph.xyz/?q=QmZncnZ8hnPSua1zN93mgqbGmJMYxJex6kUQFwRyhcayjC

### DIFF
--- a/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
+++ b/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
@@ -803,10 +803,11 @@ function getBalanceTransferAmount(event: ethereum.Event): BigInt {
       const logSignature = thisLog.topics[0];
       if (thisLog.address == event.address && logSignature == eventSignature) {
         const UINT256_UINT256 = "(uint256,uint256)";
-        const decoded = ethereum
-          .decode(UINT256_UINT256, thisLog.data)!
-          .toTuple();
-        btAmount = decoded[0].toBigInt();
+        const decoded = ethereum.decode(UINT256_UINT256, thisLog.data);
+        if (!decoded) continue;
+
+        const logData = decoded.toTuple();
+        btAmount = logData[0].toBigInt();
 
         break;
       }
--- a/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
+++ b/subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
@@ -77,6 +77,7 @@ import {
   BIGINT_ZERO,
   Network,
   INT_FOUR,
+  INT_ZERO,
 } from "../../../src/constants";
 import {
   getMarketFromToken,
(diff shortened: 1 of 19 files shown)


subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
```diff
@@ -791,18 +794,22 @@ function getBalanceTransferAmount(event: ethereum.Event): BigInt {
...
+        continue;
+      }
+      // topics[0] - signature
+      const logSignature = thisLog.topics[0];
+      if (thisLog.address == event.address && logSignature == eventSignature) {
+        const UINT256_UINT256 = "(uint256,uint256)";
+        const decoded = ethereum
+          .decode(UINT256_UINT256, thisLog.data)!
+          .toTuple();
+        btAmount = decoded[0].toBigInt();
```
dmelotik, line 809: should we check if the index exists before calling it? Great bug fix here, this was not easy to find

subgraphs/aave-forks/src/helpers.ts
```diff
@@ -239,55 +239,59 @@ export function getInterestRateType(
...
-      // skip event with logIndex < LogIndexMinus5
-      continue;
-    }
-    if (thisLog.logIndex.equals(logIndexMinus3)) {
-      // break if the logIndex = event.logIndex - 3
-      break;
-    }
+    if (thisLog.topics.length >= INT_THREE) {
+      if (thisLog.logIndex.lt(logIndexMinus5)) {
+        // skip event with logIndex < LogIndexMinus5
```
dmelotik, line 244: this comment no longer applies

dmelotik:
Great work @dhruv-chauhan ! This was not an easy bug to find. There is still an error in aave v3 polygon. Once that is fixed and we merge the PR you can redeploy all of the aave changes

dhruv-chauhan:
Found the issue with aave-v3-polygon,

`RewardToken` is set as an immutable entity in the [schema](https://github.com/dhruv-chauhan/subgraphs/blob/aave-indexing-err/subgraphs/aave-forks/schema.graphql#L81), but we're trying to update a helper field `rewardToken._distributionEnd` [here](https://github.com/dhruv-chauhan/subgraphs/blob/aave-indexing-err/subgraphs/aave-forks/src/mapping.ts#L1277).

Making the entity mutable will fix this. If we don't want to do that, we might have to create an additional helper entity for this helper field. 
@dmelotik

dmelotik:
@dhruv-chauhan I would say it is fine to make `RewardToken` not immutable. This shouldn't affect the syncing/querying that much since we had it that way before

subgraphs/aave-forks/protocols/aave-v3/src/mapping.ts
dhruv-chauhan: checking if `log.data` is decoded into the given signature, if yes, it will have this index

dmelotik (approved):
lgtm!

Issue #2299, opened 2023-07-27
Uniswap-V3 Ethereum: X token causing TVL to go into the trillions


bye43:
- Looks like X/`0x8e002692a1604fdde5cb21cdce4002dda9833496` is being priced incorrectly causing the TVL to be in the trillions 
- [Link to query](https://api.thegraph.com/subgraphs/name/messari/uniswap-v3-ethereum/graphql?query=%7B%0A++liquidityPool+%28id%3A+%220x8e002692a1604fdde5cb21cdce4002dda9833496%22%29+%7B%0A++++name%0A++++id%0A++++totalValueLockedUSD%0A++%7D%0A%7D++)

PR #2300, opened 2023-07-27
fix(#minor); Uniswap V3 Etheruem; Blacklist X token

steegecs:
**Context:** 

X token is causing the TVL on Uniswap V3 Ethereum to be in the trillions. The purpose of this PR is to blacklist this token so that it can be grafted and fixed. 

Connect to issue #2299

https://api.thegraph.com/subgraphs/name/messari/uniswap-v3-ethereum/graphql?query=%7B%0A++liquidityPool+%28id%3A+%220x8e002692a1604fdde5cb21cdce4002dda9833496%22%29+%7B%0A++++name%0A++++id%0A++++totalValueLockedUSD%0A++++inputTokens+%7B%0A++++++id%0A++++++name%0A++++++lastPriceUSD%0A++++%7D%0A++%7D%0A%7D++

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -3732,7 +3732,7 @@
         "status": "prod",
         "versions": {
           "schema": "4.0.0",
-          "subgraph": "1.5.0",
+          "subgraph": "1.6.0",
           "methodology": "1.0.0"
         },
         "files": {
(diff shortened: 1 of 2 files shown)


dmelotik:
@steegecs what is the block I should graft at?

steegecs:
> @steegecs what is the block I should graft at?

I grafted it! I tried doing it too early the last time, and it kept an old price causing the spike to stay, but the fix is pending on hosted.

dmelotik (approved):
approved

PR #2302, opened 2023-07-28
feat(#minor); aave-forks; update position.principal for aave-forks

dhruv-chauhan:
Deployments,

aave-v3-ethereum: https://okgraph.xyz/?q=QmR5QYfpcCptn2H7hC8mkXZcE7unVW7nvBVe4G1Yy3XAw7

Query,

https://api.thegraph.com/subgraphs/id/QmR5QYfpcCptn2H7hC8mkXZcE7unVW7nvBVe4G1Yy3XAw7/graphql?query=%7B%0A++positions%7B%0A++++id%0A++++hashOpened%0A++++principal%0A++++balance%0A++%7D%0A%7D

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -140,7 +140,7 @@
         "status": "prod",
         "versions": {
           "schema": "3.1.0",
-          "subgraph": "2.0.2",
+          "subgraph": "2.0.4",
           "methodology": "1.1.0"
         },
         "files": {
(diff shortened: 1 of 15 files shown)


deployment/deployment.json
dhruv-chauhan, line 1: subgraph/sdk version update considering #2298 goes out first

dmelotik:
My comments:

- Removing the principal field makes the sdk less generic, we are assuming that the new balance is always the principal (which is not the case in comp v3)
- This looks like the principal balance of the debt side: https://docs.aave.com/developers/tokens/debttoken#principalbalanceof
- and for the supply side, `token.balanceOf()` will return the principal + interest: https://docs.aave.com/developers/tokens/atoken#eip20-methods so _maybe_ the `token.scaledBalanceOf()`

PR #2303, opened 2023-07-28
fix(#patch); quickswap; swap only error

dmelotik:
Made the early return on bad tokens before calling the token data.

grafting: https://okgraph.xyz/?q=messari%2Fquickswap-swap-polygon


dhruv-chauhan (approved):
looks good!

PR #2315, opened 2023-08-07
fix(#patch); gmx-forks; update subtractArrays method to handle uneven arrays

dhruv-chauhan:
Deployments,

- gmx-arbitrum-graft (graft, block: 104485000): https://okgraph.xyz/?q=dhruv-chauhan%2Fgmx-arbitrum-graft
- gmx-arbitrum (complete): https://okgraph.xyz/?q=dhruv-chauhan%2Fgmx-arbitrum

Query,

https://api.thegraph.com/subgraphs/id/QmfYbKE9pBpXii8CYemccw9XdsDc2Hd3hDiFahpB1JBiit/graphql?query=%7B%0A++liquidityPoolDailySnapshots%28%0A++++first%3A5%2C%0A++++orderBy%3Adays%2C%0A++++orderDirection%3Adesc%2C%0A++++where%3A+%7B%0A++++++days_in%3A+%5B19533%2C19534%2C19535%5D%0A++++%7D%0A++%29%7B%0A++++days%0A++++inputTokens+%7B%0A++++++id%0A++++%7D%0A++++dailyVolumeByTokenUSD%0A++%7D%0A%7D

### DIFF
--- a/subgraphs/gmx-forks/src/sdk/protocols/perpfutures/pool.ts
+++ b/subgraphs/gmx-forks/src/sdk/protocols/perpfutures/pool.ts
@@ -10,7 +10,11 @@ import { Perpetual } from "./protocol";
 import { TokenManager } from "./tokens";
 import { PoolSnapshot } from "./poolSnapshot";
 import * as constants from "../../util/constants";
-import { exponentToBigDecimal, safeDivide } from "../../util/numbers";
+import {
+  exponentToBigDecimal,
+  poolArraySort,
+  safeDivide,
+} from "../../util/numbers";
 
 import {
   LiquidityPoolFee,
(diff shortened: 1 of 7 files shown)


subgraphs/gmx-forks/src/sdk/protocols/perpfutures/pool.ts
```diff
@@ -770,20 +766,6 @@ export class Pool {
     cumulativeOutflowVolumeByTokenAmount.push(constants.BIGINT_ZERO);
     cumulativeOutflowVolumeByTokenUSD.push(constants.BIGDECIMAL_ZERO);
 
-    poolArraySort(
```
dhruv-chauhan, line 773: imo, we don't need to sort these arrays as it will disturb the graphs which are built based on token's index. Keeping a new token, and corresponding zero values, at the last index should be just fine.

dmelotik: Looks like yes to me

dmelotik:
Other than sorting arrays the PR lgtm @dhruv-chauhan

subgraphs/gmx-forks/src/sdk/protocols/perpfutures/pool.ts
dmelotik: Also, if we stop sorting the arrays then we should remove this func since this is the only use of it

ishraq8:
Qa looks good

subgraphs/gmx-forks/src/sdk/protocols/perpfutures/pool.ts
dhruv-chauhan: Looks like inputTokens array will in fact be always sorted, and we would need to sort the related arrays as suggested 
in the docs.

---
_Noting here that inputTokens-related metrics at any point should be used in reference to inputTokens' then order. Reading them in isolation can cause unexpected values._

_Currenlty internal dashboard does not handle for inputTokens' order change. For the following example of GMX pool, Token[7] seems to be changed on 2021-11-25 and 2023-06-26_

Screenshot 2023-09-01 at 3 24 16 PM

@dmelotik

ishraq8 (approved):

ishraq8:
@dhruv-chauhan @dmelotik approved this PR. if there is no issues with the code we can merge.

dmelotik:
@ishraq8 thank you ! We discussed a fix to https://github.com/messari/subgraphs/pull/2315#discussion_r1312832747 and I belv it still needs to be implemented, but good to go after that

ishraq8:
Thanks @dmelotik for following up

PR #2345, opened 2023-08-24
feat(#major); aave v3; add base subgraph

dmelotik:
Adding Aave V3 Base

deployment: https://okgraph.xyz/?q=messari%2Faave-v3-base

### DIFF
--- a/subgraphs/aave-forks/src/constants.ts
+++ b/subgraphs/aave-forks/src/constants.ts
@@ -28,6 +28,7 @@ export namespace Network {
   export const XDAI = "XDAI"; // aka Gnosis Chain
 
   export const METIS = "andromeda";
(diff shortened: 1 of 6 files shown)


subgraphs/aave-forks/src/constants.ts
```diff
@@ -28,6 +28,7 @@ export namespace Network {
   export const XDAI = "XDAI"; // aka Gnosis Chain
 
   export const METIS = "andromeda";
+  export const BASE = "base";
```
dhruv-chauhan, line 31: nit: although we use this inside equalsIgnoreCase() so it shouldn't matter, but we can use uppercases here for base and metis like for all other networks

dhruv-chauhan (approved):
Lg!

PR #2346, opened 2023-08-24
fix(#patch); level-finance; remove unnecessary getOrCreatePool call

dhruv-chauhan:
- Error: Trying to overwrite immutable `LiquidityPoolHourlySnapshot` due to duplicate ids
- Cause: 
  - `PoolHourlySnapshot`'s id is created using `pool._lastUpdateTimestamp`, which is internally updated everytime there is an update on pool entity's fields
  - In the current implementation, `getOrCreatePool` is called inside handler `handleLogPoolAddition`, but no field of the pool is updated. This creates a `LiquidityPoolHourlySnapshot` object, but does not update `pool._lastUpdateTimestamp`, which causes indexer to fail in the following iteration.
- Fix: This unnecessary `getOrCreatePool` call can be removed with no impact
- Deployments:
  - https://okgraph.xyz/?q=dhruv-chauhan%2Flevel-finance-bsc
  - https://okgraph.xyz/?q=dhruv-chauhan%2Flevel-finance-arbitrum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -9080,7 +9080,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.1",
-          "subgraph": "1.0.0",
+          "subgraph": "1.0.1",
           "methodology": "1.0.0"
         },
         "services": {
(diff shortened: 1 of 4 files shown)


dmelotik (approved):
lgtm

PR #2347, opened 2023-08-24
fix(#patch); stargate-base; add Base's block rate

dhruv-chauhan:
- Error: missing Base's network key in getStartingBlockRate()
- Fix: Since Base is built on OP stack, it shall have the same block rate as Optimism
- Deployments: 
  - https://okgraph.xyz/?q=dhruv-chauhan%2Fstargate-base


dmelotik (approved):
lgtm

PR #2369, opened 2023-09-13
fix(#patch); goldfinch; check if PoolToken exists

dhruv-chauhan:
- Issue: 
  - `assert(PoolToken.load(poolTokenIds[i]))` fails
  - `PoolToken` with id: 943 was removed from store here: https://github.com/messari/subgraphs/blob/master/subgraphs/goldfinch/src/mappings/pool_tokens.ts#L25
  - `callableLoan.tokens` stores reference to `PoolToken`. Iterating over this array in code still shows the removed token Id.
- Deployment:
  - goldfinch-ethereum: https://okgraph.xyz/?q=dhruv-chauhan%2Fgoldfinch-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2558,7 +2558,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.4.1",
+          "subgraph": "1.4.2",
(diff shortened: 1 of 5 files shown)


dmelotik (approved):
Good change, lgtm!

PR #2374, opened 2023-09-15
fix(#patch); Level Finance Arbitrum; fix negative balance error

shashwatS22:
Arbitrum Deployment: https://thegraph.com/hosted-service/subgraph/shashwats22/level-finance-fixes-arbitrum?version=pending&selected=logs

### DIFF
--- a/subgraphs/level-finance/src/modules/transaction.ts
+++ b/subgraphs/level-finance/src/modules/transaction.ts
@@ -20,7 +20,7 @@ export function transaction(
 ): void {
   const account = getOrCreateAccount(accountAddress, pool, sdk);
   const token = sdk.Tokens.getOrCreateToken(tokenAddress);
-  if (!token._isWhitelisted) return;
+  if (token._isWhitelisted === false) return;
 
   if (transactionType === TransactionType.DEPOSIT) {
     utils.checkAndUpdateInputTokens(pool, token, amount);
--- a/subgraphs/level-finance/src/modules/transaction.ts
+++ b/subgraphs/level-finance/src/modules/transaction.ts
@@ -20,7 +20,7 @@ export function transaction(
 ): void {
   const account = getOrCreateAccount(accountAddress, pool, sdk);
   const token = sdk.Tokens.getOrCreateToken(tokenAddress);
-  if (token._isWhitelisted === false) return;
+  if (!token._isWhitelisted) return;
 
   if (transactionType === TransactionType.DEPOSIT) {
     utils.checkAndUpdateInputTokens(pool, token, amount);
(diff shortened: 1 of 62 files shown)


subgraphs/level-finance/src/modules/transaction.ts
```diff
@@ -31,11 +38,23 @@ export function transaction(
   const poolInputTokens = pool.getInputTokens();
   const idx = pool.getInputTokens().indexOf(token.id);
   const inputTokenBalances = pool.pool.inputTokenBalances;
-  if (transactionType === TransactionType.DEPOSIT) {
-    inputTokenBalances[idx] = inputTokenBalances[idx].plus(amount);
-  }
-  if (transactionType === TransactionType.WITHDRAW) {
-    inputTokenBalances[idx] = inputTokenBalances[idx].minus(amount);
+  log.warning("[Network] {}", [dataSource.network()]);
+  if (dataSource.network() === "arbitrum-one") {
```
dmelotik, line 42: use a constant for "arbitrum-one"

see: https://github.com/messari/subgraphs/blob/04fc7cd2107f1456d61ca2e7d4783c2b72ee99d4/subgraphs/compound-v3/protocols/compound-v3/src/constants.ts#L138

ishraq8:
these fixes look good and passed QA

dmelotik (changes requested):
Code lgtm! @shashwatS22 can you increment the subgraphVersion for `level-finance-arbitrum` in `deployment.json`? Thanks

subgraphs/level-finance/src/modules/transaction.ts
```diff
@@ -31,11 +32,23 @@ export function transaction(
...
+  if (utils.equalsIgnoreCase(network, constants.Network.ARBITRUM_ONE)) {
+    const poolContract = PoolContract.bind(
+      Address.fromBytes(pool.getBytesID())
+    );
+    const tokenBalance = utils.readValue(
+      poolContract.try_poolBalances(Address.fromBytes(token.id)),
+      constants.BIGINT_ZERO
+    );
+    inputTokenBalances[idx] = tokenBalance;
+  } else {
```
dmelotik, line 45: Is BSC not subject to the same negative balance issues? Or does bsc not have the `try_poolBalances()` function available to call?

shashwatS22: It does not have the negative balance issue, but I understand your point that it could be there in the future, so should I add the poolBalances function for BSC too? I didn't add it till now because we generally try to avoid network calls.

dmelotik: I think the current school of thought is to avoid read calls to the chain if they can be replicated with event data. From my understanding + / - doesn't replicate `try_poolBalances()` exactly

shashwatS22: Added similar function to bsc too.

melotik:
bsc deployment: https://thegraph.com/hosted-service/subgraph/shashwats22/level-finance-fixes

melotik (approved):
thanks @shashwatS22 !

PR #2376, opened 2023-09-20
fix(#patch); pancakeswap-v2-swap; add check for token.decimals

dhruv-chauhan:
- Issue: converting raw amount to token decimal terms fails 
  - with error, `big decimal exponent 6145 is outside the -6143 to 6144 range`
  - for token, https://bscscan.com/address/0x6a204b19b8273f346381B476BCAAd6B7b0D71fbd
- Fix: ignore swap events if token decimals out of range [-6143, 6144]
- Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fpancakeswap-v2-swap

### DIFF
--- a/subgraphs/uniswap-forks-swap/protocols/pancakeswap-v2-swap/config/deployments/pancakeswap-v2-swap-bsc/configurations.json
+++ b/subgraphs/uniswap-forks-swap/protocols/pancakeswap-v2-swap/config/deployments/pancakeswap-v2-swap-bsc/configurations.json
@@ -5,7 +5,7 @@
     "address": "0xca143ce32fe78f1f7019d7d551a6402fc5350c73",
     "startBlock": 6809737
   },
-  "graftEnabled": true,
-  "subgraphId": "Qmcye3y4H9FkUqf3UVXocTYJoA7FxGGpaqbzvxLtnbktAi",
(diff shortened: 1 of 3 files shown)


subgraphs/uniswap-forks-swap/protocols/pancakeswap-v2-swap/config/deployments/pancakeswap-v2-swap-bsc/configurations.json
```diff
@@ -5,7 +5,7 @@
     "address": "0xca143ce32fe78f1f7019d7d551a6402fc5350c73",
     "startBlock": 6809737
   },
-  "graftEnabled": false,
-  "subgraphId": "",
-  "graftStartBlock": 0
+  "graftEnabled": true,
```
dmelotik, line 8: turn to false to merge into master.

in my mind we should not merge grafts in bc they may or may not exist in the future and if we want to graft it should be deliberate (not by accident)

dmelotik (approved):
I think this fix is fair, if it is outside of the range there is not a whole lot we can do, plus the token is almost positively a fake token. I wonder how we hit it though on this token since bscscan says it has 0 decimals. But the contract is not decoded so hard to say fs.

PR #2388, opened 2023-09-28
fix(#patch); apeswap-bsc; add BBQ to untracked tokens

dhruv-chauhan:
Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fapeswap-bsc

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -4742,7 +4742,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.2",
-          "subgraph": "1.2.1",
+          "subgraph": "1.2.2",
           "methodology": "1.0.0"
         },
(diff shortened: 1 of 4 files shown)


deployment/deployment.json
```diff
@@ -4742,7 +4742,7 @@
         "status": "prod",
         "versions": {
           "schema": "1.3.2",
-          "subgraph": "1.2.1",
+          "subgraph": "1.2.2",
```
steegecs, line 4745: Should this be a minor change since it needs re-deployed?

dmelotik: Good catch, we have been not been good ab this lately. But we can try again!

dmelotik (approved):
lgtm!

PR #2393, opened 2023-09-29
chore(); monitor; update decentralized network deployments endpoint

dhruv-chauhan:
*need to update `GRAPH_API_KEY` in .env while deploying*


dmelotik (approved):
lgtm!

PR #2401, opened 2023-10-13
chore(); change github username

melotik:
changed username from @dmelotik -> @melotik 

want to update all references


dhruv-chauhan (approved):
might want to fix linting or add ignore rule in `maple-finance-v2/src/sdk/token.ts`

PR #2404, opened 2023-10-16
feat(#major); dashboard; use status endpoint

dhruv-chauhan:
* replaces queries for deployments list (github api) and their indexing status (graph api) with a single query to status endpoint


melotik (commented):
Code changes lgtm! A couple things before merging

- Also consider rendering the names/networks for when the status endpoint takes a long time to return (if not too much lift)
- Otherwise we will add a loader

dhruv-chauhan:
- add a check to fetch deployments only when not present
- add loader

@melotik loading time looks much better now

https://github.com/messari/subgraphs/assets/32761670/bbc9b7b6-51c3-4c37-bc2e-f1287f0f8cc9

melotik (approved):
lgtm! Thanks @dhruv-chauhan 

Note: We know that load times may be slower in some cases. If this becomes an issue we will add in some conditional loading to get data populated quicker.

PR #2408, opened 2023-10-26
chore(); remove magic number checker

melotik:
I am proposing to remove the magic number lint checker rule. We added this in to try and button up our code more.

However, magic numbers are needed in a lot of cases when building subgraphs (such as block numbers, pricing mechanisms, configs). And there are too many use edge cases to try and create a custom rule (we will always be one step behind).

We ended up just ignoring the rule too many time.


dhruv-chauhan (approved):
lg!

PR #2438, opened 2023-12-06
fix(#patch); hop-protocol-xdai; add missing WMATIC_XDAI address

dhruv-chauhan:
- Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fhop-protocol-xdai-graft


melotik (approved):
LGTM

PR #2454, opened 2024-01-15
fix(#patch); venus; handle proxy contracts update

dhruv-chauhan:
- Issues:
  - missing markets
  - deposits not being recorded
- These issues are due to updated implementations for proxy Comproller and VToken equivalent contracts.
- Deployments:
  - (grafted) https://okgraph.xyz/?q=dhruv-chauhan%2Fvenus-protocol-bsc
  - (complete) https://okgraph.xyz/?q=dhruv-chauhan%2Fvenus-protocol-bsc-2
- Note:
  - grafted deployment fixed missing markets, lists new markets for UNI and FDUSD and records deposit events for them
  - for capturing mint (deposit) events older markets like BTC, we will have to wait for complete syncing

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2128,7 +2128,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.2.2",
+          "subgraph": "1.3.0",
           "methodology": "1.0.0"
         },
         "files": {
--- a/subgraphs/compound-forks/abis/venus/ComptrollerV2.json
+++ b/subgraphs/compound-forks/abis/venus/ComptrollerV2.json
@@ -1,0 +1,3092 @@
+[
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": true,
+          "internalType": "contract VToken",
+          "name": "vToken",
+          "type": "address"
+        },
+        {
+          "indexed": true,
+          "internalType": "enum ComptrollerTypes.Action",
+          "name": "action",
+          "type": "uint8"
+        },
+        {
+          "indexed": false,
+          "internalType": "bool",
+          "name": "pauseState",
+          "type": "bool"
+        }
+      ],
+      "name": "ActionPausedMarket",
+      "type": "event"
+    },
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": false,
+          "internalType": "bool",
+          "name": "state",
+          "type": "bool"
+        }
+      ],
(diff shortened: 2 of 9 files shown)


melotik (commented):
@dhruv-chauhan I think I am confused how this change is going to fix the issue..?

- You create the template for Ctoken and VtokenV2 for each and every market
- and the events `Mint()` and `Redeem()` point to the same function twice

1. Wouldn't the event signature be different and not even work for `handleMint` for the V2 contract?

If you look at what I did in compound V2 (https://github.com/messari/subgraphs/blob/48b6016c64a11d8d6d3f59cc2952c479af4d6678/subgraphs/compound-forks/protocols/compound-v2/src/mapping.ts#L228) we just point to the new template when it was first used.

What do you think of this? Maybe I am wrong

dhruv-chauhan:
> * You create the template for Ctoken and VtokenV2 for each and every market

I think that's necessary, and should not be much overhead. I'm assuming `MarketListed` is not emitted again on ABI change.
I'm not sure how [compound-v2's change](https://github.com/messari/subgraphs/blob/48b6016c64a11d8d6d3f59cc2952c479af4d6678/subgraphs/compound-forks/protocols/compound-v2/src/mapping.ts#L228) would handle for ABI change in existing markets.

> * and the events `Mint()` and `Redeem()` point to the same function twice. Wouldn't the event signature be different and not even work for `handleMint` for the V2 contract?

Yes, the signatures will be different, but it works in this case since the event.params we are using inside the handlers are common for both signatures.
If that's not a good practice, I can create separate handlers for both versions.

@melotik

melotik:
@dhruv-chauhan 

I see, yeah I think I was getting confused bc comp v2 does not change the abi after it has been deployed. It simply just as a new abi for newer markets. So this seems like an accurate fix.

As for the question about using the same function for both ABIs, to me that is kind of confusing. But, it is not a huge deal. In an effort to not add more indexing time, I think we can leave it as is.

Approving, but make sure to fix the prettier stuff before merging :)

melotik (approved):

dhruv-chauhan:
Not sure why prettier is still failing. Failing check is passing fine for me,
```
prettier --check "subgraphs/compound-forks/protocols/venus/src/mapping.ts" "subgraphs/compound-forks/src/mapping.ts"
Checking formatting...
All matched files use Prettier code style!
```

melotik:
> Not sure why prettier is still failing. Failing check is passing fine for me,

@dhruv-chauhan I get the same thing. We can merge this to unblock for now

PR #2459, opened 2024-01-20
fix(#patch); currentHolderCount in erc20 subgraph

zjesko:
@nemani @dhruv-chauhan @melotik 

Fix: `currentHolderCount` value turns negative for some tokens.

Eg. arbitrum (0x9e20461bc2c4c980f62f1B279D71734207a6A356)

Before fix:
Screenshot 2024-01-20 at 8 13 50 PM

After fix:
Screenshot 2024-01-20 at 9 00 18 PM

### DIFF
--- a/subgraphs/erc20/src/mappings/token.ts
+++ b/subgraphs/erc20/src/mappings/token.ts
@@ -320,20 +320,15 @@ function handleTransferEvent(
     if (isNewAccount(destination)) {
       // It means the receiver is a new holder
       toAddressIsNewHolderNum = BIGINT_ONE;
-    } else {
-      balance = getOrCreateAccountBalance(
-        getOrCreateAccount(destination),
-        token
-      );
-      if (balance.amount == BIGINT_ONE) {
-        // It means the receiver's token balance is 0 before transferal.
-        toBalanceIsZeroNum = BIGINT_ONE;
-      }
+    }
+    balance = getOrCreateAccountBalance(getOrCreateAccount(destination),token);
+    if (balance.amount == BIGINT_ZERO) {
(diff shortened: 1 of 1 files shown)


subgraphs/erc20/src/mappings/token.ts
```diff
@@ -320,20 +320,15 @@ function handleTransferEvent(
...
+    }
+    balance = getOrCreateAccountBalance(getOrCreateAccount(destination),token);
+    if (balance.amount == BIGINT_ZERO) {
+      // It means the receiver's token balance is 0 before transferal.
+      toBalanceIsZeroNum = BIGINT_ONE;
     }
 
     token.currentHolderCount = token.currentHolderCount
       .minus(FromBalanceToZeroNum)
-      .plus(toAddressIsNewHolderNum)
```
dhruv-chauhan, line 336: is this because `toBalanceIsZeroNum` will already count a new toAddress?

zjesko: yes, because the earlier if else was simplified. `toAddressIsNewHolderNum` is not used in this calculation anymore because it doesnt matter if the addresses is a new holder or not

zjesko:
Ran prettier for the `erc20` folder

melotik:
@zjesko I have to approve the workflow runs for the ci/cd to pass. But you can actually do this locally. Here are the commands for the failing checks:

- Lint checker: `npx eslint .` and it only checks files that were changed in the PR
- Build checker: you can see the issues it is running into using `messari build erc20 -l`

steegecs (approved):

PR #2462, opened 2024-01-23
fix(#patch); goldfinch; failing assert on burned PoolToken

dhruv-chauhan:
- Issue: `assert(PoolToken.load(poolTokenIds[i]))` fails since some PoolToken was removed from store [here](https://github.com/messari/subgraphs/blob/master/subgraphs/goldfinch/src/mappings/pool_tokens.ts#L25) on `TokenBurned`. `tranchedPool.tokens` stores reference to PoolToken. Iterating over this array in code still shows the removed token Id.
- Previously seen: https://github.com/messari/subgraphs/pull/2369
- Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fgoldfinch-ethereum

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2638,7 +2638,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.4.2",
+          "subgraph": "1.4.3",
           "methodology": "1.0.1"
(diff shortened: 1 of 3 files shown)


melotik (approved):
Nice work on the clear documentation of the issue and fix

PR #2467, opened 2024-01-27
fix decimal overflow issue

DyT9qC:
**Problem Scenario:**

When the indexer attempts to call `try_decimals()`, if the decimal value exceeds uint8, the following error message will be triggered:

`Mapping aborted at ~lib/@graphprotocol/graph-ts/common/collections.ts, line 150, column 9, with message: overflow converting 0x0000000000000000000000000000000000000000000000000000000067e53c31 to i32\twasm backtrace:\t    0: 0x3ddd - <unknown>!~lib/@graphprotocol/graph-ts/chain/ethereum/ethereum.Value#toI32\t    1: 0x3ed9 - <unknown>!src/types/Factory/ERC20/ERC20#try_decimals`

**Fix Solution:**

Check if the decimal value is of type int or uint, and verify whether it is less than the maximum value of uint8. If both conditions are met, return the correct value; otherwise, return 0.

### DIFF
--- a/subgraphs/dopex/src/entities/token.ts
+++ b/subgraphs/dopex/src/entities/token.ts
@@ -1,4 +1,10 @@
-import { Address, BigDecimal, Bytes, ethereum } from "@graphprotocol/graph-ts";
+import {
+  Address,
+  BigDecimal,
+  BigInt,
+  Bytes,
+  ethereum,
+} from "@graphprotocol/graph-ts";
 import { ERC20 } from "../../generated/BasicWeeklyCalls/ERC20";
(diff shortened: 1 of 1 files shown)


dhruv-chauhan:
Hey @DyT9qC , I'm closing this PR since these are known uni-v3 issues caused by unverified token contracts and these are very insignificant to be dealt with separately. PR #2470 resolves the issues.

For future PRs, I'd request you to follow our [contribution guidelines](https://github.com/messari/subgraphs/blob/master/docs/CONTRIBUTING.md), and most importantly add a test deployment of the changes in your own hosted service environment. Cheers!

PR #2474, opened 2024-01-31
fix(#patch); burrow-near; failed deployment due to null event.args.msg in liquidate event

dhruv-chauhan:
Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fburrow-near

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -1474,7 +1474,7 @@
         "status": "prod",
         "versions": {
           "schema": "2.0.1",
-          "subgraph": "1.0.1",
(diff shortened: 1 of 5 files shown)


melotik (approved):
Imo, I think the null check is appropriate. If the data isn't there, the data isn't there.

Could you just add a warning log saying it's null and dumping some of the important info.

If you want to look through it further, I was scrolling through burrow's txs here: https://nearblocks.io/address/contract.main.burrow.near (around pg 95)

Screenshot 2024-02-02 at 10 17 58 AM

There is no transaction on that block...

PR #2489, opened 2024-02-20
feat(#major); friend-tech; subgraph optimizations

dhruv-chauhan:
- Huge arrays in `friend-tech-base` deployment causing issues for other subgraphs in the shard.
- These arrays were primarily defined to link `Trade` directly to `Account` entity and vice-versa. However, this can also be accessed using through a mapping table - `Connection`. 
Other arrays are either defined as derivedFrom fields or their lengths are bound by input tokens length.
- Test deployment: https://thegraph.com/studio/subgraph/friend-tech-base/
(Getting some graph-node error when trying to deploy on hosted service, checking with graph team and update)

### DIFF
--- a/subgraphs/friend-tech/src/mappings/handlers.ts
+++ b/subgraphs/friend-tech/src/mappings/handlers.ts
@@ -6,7 +6,6 @@ import {
   getOrCreateToken,
 } from "../common/getters";
 import {
-  updateConnection,
   updateRevenue,
   updateShares,
   updateTVL,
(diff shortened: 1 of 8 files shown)


subgraphs/friend-tech/src/mappings/handlers.ts
```diff
@@ -41,7 +40,7 @@ export function handleTrade(event: Trade): void {
     event
   );
 
-  const tradeID = createTrade(
+  createTrade(
```
melotik, line 43: nit: we can remove the return from `createTrade()` as this is the only function using it

melotik (approved):

PR #2497, opened 2024-03-12
fix(#patch); belt-finance-bsc; add check for null transaction.to

dhruv-chauhan:
Deployment: https://okgraph.xyz/?q=dhruv-chauhan%2Fbelt-finance


melotik (approved):
@dhruv-chauhan can you deploy to the hosted service?

PR #2509, opened 2024-03-21
fix(#patch); compound-v3-base; indexing error due to empty log.topics

dhruv-chauhan:
- Deployment ID (studio): QmPSrf5TWCDYBo9tfM39765gAeeUMC47seCNng1aySvtCj
(https://subgraphs.messari.io/subgraph?endpoint=QmPSrf5TWCDYBo9tfM39765gAeeUMC47seCNng1aySvtCj&tab=protocol)

### DIFF
--- a/deployment/deployment.json
+++ b/deployment/deployment.json
@@ -2288,7 +2288,7 @@
         "status": "prod",
         "versions": {
           "schema": "3.1.0",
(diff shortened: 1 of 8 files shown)


melotik:
@melotik to resync all of comp v3 so this doesn't happen again

melotik (approved):

PR #2530, opened 2024-04-15
Add Qidao Base

melotik:
Add QiDao Base

https://okgraph.xyz/?q=QmeivVyBQ8DpY5zBnzgHviy8fS5UnuCaVBNXVPUvyEKc7G


subgraphs/compound-forks/src/mapping.ts
```diff
@@ -1642,7 +1642,7 @@ export function updateMarket(
     );
   }
 
-  const interestAccumulatedUSD = interestAccumulatedMantissa
+  const interestAccumulatedUSD = `interestAccumulatedMantissa`
```
dhruv-chauhan, line 1645: typo?

deployment/deployment.json
```diff
@@ -4753,6 +4753,32 @@
           }
         }
       },
+      "qidao-base": {
+        "network": "base",
+        "status": "prod",
+        "versions": {
+          "schema": "1.3.0",
+          "subgraph": "1.0.2",
```
dhruv-chauhan, line 4761: shouldn't the subgraph version be 1.0.0?

steegecs (approved):
Getting some linting errors in `constant.ts`.

deployment/deployment.json
melotik: Yeah, it's weird now with only being able to deploy to the studio. You can't redeploy the same version, so I had to increment each time i tested

PR #2535, opened 2024-04-29
[DAT-2562]; Update node version to address security vulnerability

steegecs:
**Context:**
We want to see if pushing through an update of the node version to node:20 will resolve the security vulnerability issues we are facing in the subgraph monitor. 

The Dashboard is still up and functional after this update when run locally with docker as shown in the images below.

Screenshot 2024-04-29 at 5 24 46 PM

Screenshot 2024-04-29 at 5 25 05 PM

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -1,7 +1,5 @@
 # pull the base image
-FROM node:lts-alpine
-
-RUN apk update && apk upgrade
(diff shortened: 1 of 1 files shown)


samortaylor (approved):

melotik (approved):
lgtm, thanks @steegecs

PR #2550, opened 2024-05-17
Simplify Generic Schema

steegecs:
**Context:**

The purpose here is to create a simplified generic schema that will be applied to new subgraphs. Analogues of this may be made for specific schema types. 

**What has changed:**
- Removed Reward Token Entities and Fields
- Removed protocolControlledValue Field from protocol level metrics
- Removed hourly snapshots for usage metrics and pool entities
- Removed Output/Liquidity Token Fields from the pool entity.
- Removed Circular Buffer Entity


schema-generic.graphql
```diff
@@ -292,24 +231,9 @@ type Pool @entity @regularPolling {
   " Symbol of liquidity pool (e.g. 3CRV) "
   symbol: String
 
-  # Generally protocols accept one or multiple tokens and mint tokens to the depositor to track ownership
-  # Some protocols don't mint any tokens to track ownership, in that case outputToken is null.
-
-  " Token that is minted to track ownership of position in protocol "
-  outputToken: Token
```
melotik, line 299: I am thinking we should keep outputToken. I think it's a field that a lot of pools have and it's generally good info

schema-generic.graphql
```diff
@@ -336,165 +260,11 @@ type Pool @entity @regularPolling {
   " The USD value of input tokens in this pool. Should be the same order as the pool's `inputTokens` field. "
   inputTokenBalancesUSD: [BigDecimal!]!
 
-  " Total supply of output token. Note that certain DEXes don't have an output token (e.g. Bancor) "
-  outputTokenSupply: BigInt
```
melotik, line 340: Also keep this ^^

steegecs: Added back!

schema-generic.graphql
steegecs: Added back!

melotik (approved):

PR #2586, opened 2024-07-15
feat(#major); Lybra Finance; Add subgraph

harsh9200:
# Summary
Adds subgraph for Lybra Finance on Ethereum network using generic schema (version: 3.0.0).
Deployment ID: `QmZqszUexR13HDZDeUKTCe9Spvzx7ESewBQhXDFHHs8Xz1`

### TVL 
image
image

### Supply Side Revenue
image
image

### Protocol Side Revenue
image
image

### Usage Metrics
image
image


steegecs:
The charts we have for supply and protocol side revenue look the same for the subgraphs and for defillama. Why is this the case? Also is there an explanation for why are revenues are higher than defillama?

steegecs (approved):

